PR #3, opened 2018-06-07
Add Typography, update theme

zephraph:
This PR aims to do two things

1. Lay the groundwork for bridging style values between our [design-system](https://www.notion.so/artsy/Design-92030f16ed7c4c72bb3eb832b4243d04), emission, and reaction
2. Start fleshing out some basic universal components


src/platform/README.md
```diff
@@ -0,0 +1,5 @@
+# Platform
+
+This directory exists to hold any platform specific implementations. 
+
+React-Native by default will require *.ios.js before *.js when building on an ios platform. Likewise for android.
```
damassi: 👍

zephraph:
I don't have an explicit export for the typography stuff yet. How should we expose those?

I think `import { Sans } from 'palette/Typography'` would be a nice semantic. I'm open to suggestions though.

damassi:
@zephraph - that seems good to me. Mind adding that export?

zephraph:
For that to work, Typography would need to end up in the root of the palette package after it was built, right?

damassi:
Similar to exports in `index.tsx`, can add 

```
export { Typography } from "./Typography"
```

src/elements/Typography.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const Sans = props => (
+  <Text family="unica" typeSize={`sans${props.size || 4}`} {...props} />
+)
+export const Serif = props => (
+  <Text family="garamond" typeSize={`serif${props.size || 4}`} {...props} />
+)
+export const Display = props => (
+  <Text family="avantgarde" typeSize={`display${props.size || 2}`} {...props} />
+)
```
damassi: Can you add some quick interim prop-types here just so its clear what the api is, with a tiny example? Something similar to 

https://github.com/artsy/reaction/blob/master/src/Styleguide/Elements/Typography.tsx#L37-L49

src/platform/primitives.js
```diff
@@ -0,0 +1,6 @@
+import styles from "styled-components"
+
+export const styled = {
+  Text: styles.span,
```
l2succes: @zephraph I just ran exogenesis locally, I see what you mean by a lot of classes. Looks like `react-native-web` is generating a separate class for every CSS property. But if that is the only issue, along with `TouchableOpacity` not working as intended. We should be able to fix those. For now tho, I think this is a fine solution if it works out.

src/elements/Typography.jsx
```diff
@@ -0,0 +1,43 @@
...
+const fontPath = props => [
+  `fontFamily.${props.family}.${selectFontType(props)}`,
+  "regular", // fallback
+]
+
+const fontSize = props => `typeSizes.${props.typeSize}.fontSize`
+
+const lineHeight = props => `typeSizes.${props.typeSize}.lineHeight`
+
+const Text = styled.Text`
```
l2succes: @zephraph as @damassi mentioned, let's add an interface to that one as well. Something like,

```
interface TextProps {
  family: "unica" | "garamond" | "avantgarde"
  ...
}
```

src/platform/primitives.js
damassi: One thing that makes this _kinda_ ok is if you click into the react devtools tab in chrome this appears, which piggybacks on the ReactNative styled-components hook:

screen shot 2018-06-07 at 11 44 49 am

src/elements/Typography.tsx
```diff
@@ -32,12 +37,16 @@ const Text = styled.Text`
...
+export const Sans = (props: TypographyProps) => (
+  <Text family="unica" typeSize={`sans${props.size || 3}`} {...props} />
 )
-export const Serif = props => (
-  <Text family="garamond" typeSize={`serif${props.size || 4}`} {...props} />
+export const Serif = (props: TypographyProps) => (
+  <Text family="garamond" typeSize={`serif${props.size || 3}`} {...props} />
 )
-export const Display = props => (
+export const Display = (props: TypographyProps) => (
```
damassi: I think this can be 

```
import React, { SFC } from 'react'
...
export const Display: SFC<TypographyProps> = (props) => ... 
```

zephraph: Okay, cool.

damassi:
This is an awesome start. Gonna merge and see how `semantic-release` goes 👍

yuki24:
Excellent stuff! I'm not sure if this is the right time to do this, but would it make sense to set up [the Storybook Info Addon plugin](https://github.com/storybooks/storybook/tree/master/addons/info) as part of this PR?

yuki24:
Please ignore my comment, it seems like I was late by a hair.

PR #4, opened 2018-06-07
Add ember sem-release config; tslint

damassi:
Updates package with `conventional-changelog-ember` which mirrors Reaction, as well as Reaction's `tslint.json` config.

### DIFF
--- a/src/elements/Typography.tsx
+++ b/src/elements/Typography.tsx
@@ -22,7 +22,7 @@ const fontSize = props => `typeSizes.${props.typeSize}.fontSize`
 
 const lineHeight = props => `typeSizes.${props.typeSize}.lineHeight`
 
-interface TextProps {
+export interface TextProps {
   family: "unica" | "garamond" | "avantgarde"
   typeSize: string // TODO: Make this more granular
 }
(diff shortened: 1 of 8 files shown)


src/elements/Typography.tsx
```diff
@@ -37,7 +37,7 @@ const Text = styled.Text.attrs<TextProps>({})`
   ${maxWidth};
 `
 
-interface TypographyProps {
+export interface TypographyProps {
```
damassi: @zephraph - this is needed for `tsc --emit`, which will automatically emit type definition files for consumers.

tslint.json
```diff
@@ -0,0 +1,20 @@
...
+    "member-access": [false],
+    "member-ordering": false,
+    "no-unused-expression": false,
+    "no-var-requires": false,
+    "object-literal-sort-keys": false,
+    "ordered-imports": false,
+    "no-namespace": [true, "allow-declarations"],
+    "trailing-comma": false,
+    "variable-name": [true, "ban-keywords", "allow-leading-underscore"]
+  }
```
damassi: Taken verbatim from reaction. We can can clean this up at some point.

zephraph: Yeah, there's a lot of config stuff shared across projects. Would be good to figure out a good way to share those.

damassi: Totally agree, this is also dup'd in force now too

PR #14, opened 2018-06-11
Support vertical-align on text components

zephraph:
This supports using the `verticalAlign` prop on `Sans`, `Serif`, and `Display` components. 

`verticalAlign` isn't supported by styled-system out of the box, so this is another one of those special cases where I had to create it... not sure about where the types go. We might want to create a package kinda dedicated to these special things... or make PRs for them to styled system.


damassi:
I think things should def go to styled-system when we get a minute 👍

PR #15, opened 2018-06-12
Update grid gutter to match designs

zephraph:
Currently the designs have the cutter at 20px. This updates the col padding to reflect that. 

image


damassi:
LGTM 👍 - I think i need to double triple check that the theme values from `https://github.com/artsy/palette/blob/master/src/Theme.tsx#L253` are properly being passed through; i'll follow up.

PR #16, opened 2018-06-13
Add mediaQueries to theme

zephraph:
Moves the media queries here and out of reaction. There's a bit of duplication between these definitions and the grid definitions. These values should be usable directly in styled-components.


alloy:
👍 

I’ll take these, as I need the media queries for my current work.

PR #20, opened 2018-06-15
Update space units to reflect design system

zephraph:
Design express their spacing units as multiples of 10. 

image

In an effort to unify our language, we're going to transition to match design's values exactly. Instead of a function that just converts values directly, we've added a white-list of accepted design values. This allows us both to use the unit system that design uses, gives us the flexibility to add in-between values in the future without introducing breaking changes to the spaces, and enforces the exact units that design wants to use.


damassi:
cc @briansw

PR #23, opened 2018-06-17
[FEATURE Types] Emit type declarations and add types to Typography.

alloy:
:warning: This code is tested only in this repo.


src/Theme.tsx
```diff
@@ -42,21 +42,18 @@ export const themeProps = {
   },
 
   fontFamily: {
-    // Sans
-    unica: {
+    sans: {
```
damassi: 👍

src/elements/Typography.tsx
```diff
@@ -52,18 +51,132 @@ const Text = styled.Text.attrs<TextProps>({})`
...
+   * to `regular`.
+   */
+  weight?: null | "regular" | "semibold"
+}
+
+export const Serif = createStyledText<SerifProps>("serif", (weight, italic) => {
+  if (italic && weight && weight !== "regular") {
+    throw new Error(
+      `The serif font does not have an italic font with weight \`${weight}\``
+    )
```
zephraph: I like having an error to catch this case, but would it be easy to miss if we added an italic state for that weight?

src/elements/Typography.tsx
```diff
@@ -52,18 +51,132 @@ const Text = styled.Text.attrs<TextProps>({})`
...
-// Wrap to yield control back to consuming components.
-// See: https://www.styled-components.com/docs/advanced#referring-to-other-components
-export const Sans = styledWrapper(_Sans)``
-export const Serif = styledWrapper(_Serif)``
-export const Display = styledWrapper(_Display)``
+export type FontFamily = typeof themeProps["fontFamily"]
+export type FontWeights =
+  | keyof FontFamily["sans"]
+  | keyof FontFamily["serif"]
+  | keyof FontFamily["display"]
```
zephraph: ❤️

alloy:
Not sure what’s up with Danger. If everything’s green locally then I think we can merge it.

 /cc @orta

zephraph:
👍 I'll merge. We'll have to get artsy/reaction#821 in before upgrading this in master though.

PR #25, opened 2018-06-19
Adds JSDocs to the exported palette

orta:
* adds JSDoc mostly everywhere, with the aim being that you're a client dev who doesn't have it all memorized

* **Removes** breakpoints that aren't directly in the notion doc ( I poked @zephraph about this, but @damassi you might wanna weigh in, thus you being the assignee )

### DIFF
--- a/src/Theme.tsx
+++ b/src/Theme.tsx
@@ -7,36 +7,57 @@ import { ThemeProvider } from "styled-components"
 // API: https://jxnblk.com/styled-system/api
 // Table: https://jxnblk.com/styled-system/table
 
+// https://www.notion.so/artsy/Grid-e489a52e26bd4319b6ee7898044a8a53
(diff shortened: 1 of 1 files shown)


src/Theme.tsx
```diff
@@ -7,36 +7,57 @@ import { ThemeProvider } from "styled-components"
...
-  desktop: 1024,
+  /** Between 1024 and  1191 */
   lg: 1024,
-  tablet: 900,
+  /** Between 900 and 1023 */
   md: 900,
-  phone: 768,
+  /** Between 768 and  899 */
   sm: 768,
-  smaller: 767,
```
damassi: This _should_ be fine; was initially set up by our grid library.

damassi:
Looks great 👍

PR #26, opened 2018-06-19
Add platform specific fonts

zephraph:
This adds support for platform specific font declarations. The font families implement the same interface across platforms. 

Also adds a helper to allow expressing more complex font-family configurations (like when we have to define weight, style). 

Adds fallbacks for web. Might want to double check  those and update any that are missing.


alloy:
Nice job 👏

PR #27, opened 2018-06-19
[PATCH toolchain] Fix typechecking

damassi:
This fixes our typechecking so that `yarn type-check` also emits declarations and thus catches associated interface needs.


alloy:
Aha, ta 👍

src/Theme.tsx
```diff
@@ -4,6 +4,9 @@ import { GridThemeProvider, injectLayoutBaseCSS } from "styled-bootstrap-grid"
 import { ThemeProvider } from "styled-components"
 import { fontFamily } from "./platform/fonts"
 
+// @ts-ignore
+import { FontFamily} from './platform/fonts'
```
zephraph: Why was the `@ts-ignore` needed here?

PR #28, opened 2018-06-20
Fix a few typography issues

zephraph:
Fixes a few things...

Types were always pulling sizes from sans
A few comments were wrong
Some sizes changed in notion


damassi:
thanks for the quick fix 👍 Dangerrrrr merge on green

PR #31, opened 2018-06-21
Add styled-system types from npm

zephraph:
Removes vendored styled-system types and pulls in new version from definitely typed


src/styled-system.d.ts
```diff
@@ -1,468 +0,0 @@
...
-    | "normal"
-    | "nowrap"
-    | "pre-line"
-    | "pre-wrap"
-    | "pre"
-
-  export interface WhiteSpaceProps {
-    whiteSpace?: WhiteSpaceValue
-  }
-}
```
damassi: 👋

PR #33, opened 2018-06-21
[FIX build] Remove script that was causing release process to fail

zephraph:
Removes a styled-systems cleanup artifact that was causing the deploy to fail.


zephraph:
merge on green

PR #36, opened 2018-06-25
[PATCH dev] move styled-components to devDeps

l2succes:
Fixes multiple instances of `styled-components`


l2succes:
merge on green

PR #38, opened 2018-06-29
Add some helpers to make it easy to access theme values

zephraph:
I've got some cases in my current work that I need to access values of the theme outside of the context of styled-components or styled-system. I.E. passing a color to an svg prop and using space for a width.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,0 +1,13 @@
+import { themeProps } from "./Theme"
+
+/**
+ * A helper to easily access colors when not in a styled-components or styled-systems context
+ */
+export const color = (colorKey: keyof typeof themeProps["colors"]) =>
+  themeProps.colors[colorKey]
+
+/**
+ * A helper to easily access space values when not in a styled-components or styled-systems context
+ */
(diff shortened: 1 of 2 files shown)


src/helpers.ts
```diff
@@ -0,0 +1,13 @@
+import { themeProps } from "./Theme"
+
+/**
+ * A helper to easily access colors when not in a styled-components or styled-systems context
+ */
+export const color = (colorKey: keyof typeof themeProps["colors"]) =>
```
damassi: So this is how you do that ... `keyof typeof themeProps["colors"]`

damassi:
LGTM 👍 

Danger, merge on green

zephraph:
I'll merge. Didn't do the proper commit for the release.

src/index.tsx
```diff
@@ -1,2 +1,3 @@
 export { Theme, themeProps } from "./Theme"
 export { Display, Sans, Serif } from "./elements/Typography"
+export { color, space } from "./helpers"
```
oxaudo: Very neat!

PR #39, opened 2018-06-29
Update space keys to be numerical

zephraph:
This updates the keys of our space theme settings to be numbers instead of strings. 

This change will correct the types for anything referencing space keys. 

For example, this will allow the `space` helper to be used like `space(0.3)` instead of `space("0.3")`


alloy:
Let’s do it 👍

PR #40, opened 2018-06-29
Update typescript to 2.9

zephraph:
The work that was done in #39 can't be used until our TS version is _at least_ 2.9.


zephraph:
merge on green

PR #42, opened 2018-07-10
[FEATURE types] Exposes type defs for colors and spacing values

zephraph:
The types for colors and spaces are useful in other places. This just defines and exports them explicitly so we don't have to keep redefining them.


zephraph:
merge on green

PR #45, opened 2018-07-13
Expose type sizes for easier external typing

zephraph:
Directly exports the sizes for the typography components so that they can be reused in reaction.


zephraph:
@orta The CI for this repo is kinda wonky. There's a circle ci status that's reported as required, but it's not running... not sure what's happen. Have you seen this before?

PR #50, opened 2018-07-14
Export typesizes from the index

zephraph:
Another update to expose some of the types so that they don't have to be manually imported from a sub-module.


zephraph:
merge on green

PR #54, opened 2018-07-19
Add media query helper function

zephraph:
This adds an easy way to build media query strings with palette.

```javascript
styled.div`
  ${media('xl')} {
    display: none;
  }
`
```


l2succes:
Looks good!

PR #57, opened 2018-07-26
[FEATURE typography] Add `style` prop to interface

damassi:
Adds ability to override typography props via `style`.

### DIFF
--- a/src/elements/Typography.tsx
+++ b/src/elements/Typography.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { CSSProperties } from "react"
 import styled from "styled-components"
 import { styled as primitives } from "../platform/primitives"
 import {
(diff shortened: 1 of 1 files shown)


src/elements/Typography.tsx
```diff
@@ -68,6 +68,7 @@ export interface TextProps
   fontFamily?: string
   fontSize: number
   lineHeight: number
+  style: CSSProperties
```
alloy: This is a required prop, is that intentional? The description of the PR makes it sound optional.

damassi: Ah, good catch -- fixed 👍

alloy:
Ace 👍 Merge on green.

damassi:
Danger seems to be hanging -- going to go ahead and merge since other status's are 🍏

PR #60, opened 2018-07-31
Upgrade to typescript 3

just-be-dev:
🕺


alloy:
Whoop whoop!

Did the emitted type declarations change in any way?

zephraph:
Nope. Diff was the same.

PR #62, opened 2018-08-09
Prototype a design system site

zephraph:
This is just a test using [docz](https://www.docz.site/) to generate a design system landing page. This isn't intended to be merged as is, more of an experiment to let us explore the possibility of having a design system site.  

image

It might not be worth keeping, but I thought it'd be cool for us to play around with. Let me know what you think. 

Here's a preview: https://dist-pylqxcidwa.now.sh/


orta:
This is really nice and small, feels good to me TBH

sweir27:
Yeah this is super nice. I didn't realize you could `Edit` right from the UI-- nice that it can be maintained by anyone.

alloy:
Looking great @zephraph 👌

PR #64, opened 2018-08-22
Update theme to supply breakpoints to styled-system

ds300:
https://artsy.slack.com/archives/C2TQ4PT8R/p1534966253000100

Tested against reaction:

image

breakpoints

### DIFF
--- a/src/Theme.tsx
+++ b/src/Theme.tsx
@@ -32,6 +32,11 @@ export const themeProps = {
   borders: [0, "2px solid"],
 
   /**
+   *  This allows styled-system to hook into our breakpoints
+   */
+  breakpoints: [breakpoints.sm, breakpoints.md, breakpoints.lg, breakpoints.xl],
+
+  /**
    * Artsy's color schemes:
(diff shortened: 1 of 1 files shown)


src/Theme.tsx
```diff
@@ -31,6 +31,11 @@ const breakpoints = {
 export const themeProps = {
   borders: [0, "2px solid"],
 
+  /**
+   *  This allows styled-system to hook into our breakpoints
+   */
+  breakpoints: [breakpoints.sm, breakpoints.md, breakpoints.lg, breakpoints.xl],
```
damassi: 👍

damassi:
LGTM 👍

damassi:
@ds300 - Once the release goes out via Circle can you update Reaction with the latest version of palette?

ds300:
Yep

src/Theme.tsx
zephraph: This is missing xs... was that intentional?

zephraph: Neeeever mind. I see what it's doing. `xs` is just whatever is smaller than `sm`. 👍

PR #66, opened 2018-08-31
[WIP] Update layout primitives

zephraph:
This is a continuation of #65 which I might've kinda accidentally merged. I'm just going to keep the work going in this PR. 

Original description below

---

Adds base layout primitives, which were first tested out in Emission to see what would work: 

- `Box`
- `Flex`
- `Join`
- `Separator`
- `Spacer`

In `Box` we have some Web-specific values that will need to be updated but need to discuss approach.

### DIFF
--- a/src/elements/Box.tsx
+++ b/src/elements/Box.tsx
@@ -1,7 +1,6 @@
 import { color, space } from "../helpers"
 // @ts-ignore
 import React from "react"
-// import { media } from "styled-bootstrap-grid"
 import { styled as primitives, styledWrapper } from "../platform/primitives"
 import {
   background,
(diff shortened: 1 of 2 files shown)


src/elements/Box.tsx
```diff
@@ -56,23 +55,6 @@ export const BorderBox = styledWrapper(Flex).attrs<BorderBoxProps>({})`
   ${width};
 `
 
-// export const StackableBorderBox = styled(BorderBox)`
```
zephraph: Given `StackableBorderBox` isn't built in a way that can work on RN atm, I think it's fine to just keep that in Reaction until we can figure out a good way to make it universal.

ds300:
This all seems sensible. The one thing I'm a little weirded out by is having the `styledWrapper` typing be the same for both platforms. e.g.  you could use `styledWrapper.div` in a native context. Maybe we should overload the type to prevent that. Might be possible to do `styledWrapper: Extract<typeof styled, Function>` or similar.

Maybe we should also add a lint rule preventing importing from 'styled-components' directly, so you have to add a `tslint-ignore` comment when you really need to do it?

damassi:
Going to merge this so that we can iterate

PR #67, opened 2018-09-03
Adds a tslint config requesting JSDoc for exporting components, and does some of the easier comments

orta:
This PR needs additional work, @zephraph - any chance you could throw 15m at adding some of the additional fails? It's things like `Box`, `Flex` etc, which I don't really understand enough to feel like I should be writing the core docs.

I didn't add interfaces in here, I don't think that matters too much. In the most ideal state, we turn on properties so and people know what all the options in props are - but I'm not sure if that's overkill.

### DIFF
--- a/src/Theme.tsx
+++ b/src/Theme.tsx
@@ -312,15 +312,25 @@ export const themeProps = {
   },
 }
 
+/**
+ * A wrapper component for passing down the Artsy theme context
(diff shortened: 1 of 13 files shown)


zephraph:
Yep, for sure.

orta:
👍

PR #68, opened 2018-09-05
[FEATURE Message] Add Message element; organize exports; stub Docs

damassi:
palette

### DIFF
--- a/docs/Message.mdx
+++ b/docs/Message.mdx
@@ -1,0 +1,53 @@
+---
+name: Message
+---
+
+import { Playground, PropsTable } from 'docz'
+import { Box, Message, Sans, Spacer } from '../src'
+
+# Message
+
+<Playground>
+
+  <Box width="50%">
+    <Message>
+      Thank you for your order. You’ll receive a confirmation email shortly.
+      If you have questions, please contact <a href="#">orders@artsy.net</a>.
+    </Message>
+
+    <Spacer my={4} />
+
+    <Message size="2">
+      Thank you for your order. You’ll receive a confirmation email
+      shortly. If you have questions, please contact{" "}
+      <a href="#">orders@artsy.net</a>.
+    </Message>
+
+    <Spacer my={4} />
+
+    <Message size="3">
+      Thank you for your order. You’ll receive a confirmation email
+      shortly. If you have questions, please contact{" "}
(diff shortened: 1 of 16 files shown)


src/platform/primitives.ios.ts
```diff
@@ -7,5 +7,3 @@ export const styled = {
   Text: styles.Text,
   View: styles.View,
 }
-
-export const styledWrapper = styles
```
damassi: @ds300 - Realized I didn't need this export because styled-components offers an `.extends` function!

src/elements/index.tsx
```diff
@@ -0,0 +1,7 @@
+export * from "./Box"
+export * from "./Flex"
+export * from "./Join"
+export * from "./Message"
+export * from "./Separator"
+export * from "./Spacer"
+export * from "./Typography"
```
damassi: @zephraph - I simplified these exports here and below. This will make things easier to manage.

zephraph: I forgot you could do that, heh.

src/elements/Message.tsx
```diff
@@ -0,0 +1,30 @@
...
+  ...others
+}) => {
+  return (
+    <StyledFlex p={2} {...others}>
+      <Sans size={size} color={color("black60")} weight="regular">
+        {children}
+      </Sans>
+    </StyledFlex>
+  )
+}
```
orta: I think these will need JSDocs when you rebase master now that #67 is in

orta:
very cool

src/elements/Message.tsx
damassi: 👍

damassi:
(Going to go ahead and merge this)

docs/Message.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+  <Box width="50%">
+    <Message>
+      Thank you for your order. You’ll receive a confirmation email shortly.
+      If you have questions, please contact <a href="#">orders@artsy.net</a>.
+    </Message>
+
+    <Spacer my={4} />
+
+    <Message size="2">
```
ds300: should these `size` props be `textSize`?

damassi: Ah good call! will update.

PR #69, opened 2018-09-06
[PATCH lib] Match typescript version to Emission 2.5.3

damassi:
Attempting to lock TS version to that which is in Emission until it can be updated to 3.0.


zephraph:
This is going to cause a breaking type change. Palette has a hard dependency on 2.9.

PR #71, opened 2018-09-06
[FEATURE elements] Add additional elements and set down divergent cross-platform Universal Component pattern

damassi:
Moves a few more elements over from Reaction: 

- BorderBox (hover state for web)
- StackableBorderBox 
- Avatar 

Additionally sets down a pattern for when universal components diverge across platforms: 

```
/BorderBox
  - index.tsx
  - BorderBoxBase.tsx
  - BorderBox.tsx
  - BorderBox.ios.tsx
```


src/elements/StackableBorderBox.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const StackableBorderBox = BorderBox.extend`
+  :not(:first-child) {
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+  }
+  :not(:last-child) {
+    border-bottom: 0;
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: 0;
+  }
```
damassi: @ds300 - note that we're missing the `media` helper here. I have a question out to @zephraph about how we're using the one in `helpers`.

ds300:
So is it the case that not every component in palette needs to be cross-platform?

zephraph:
We do want every component to be cross platform in the sense that you can use that component on both platforms. Doesn't mean that they need to be the same implementation or work in exactly the same way though. Ideally we'd keep the same API for the components where possible.

damassi:
@ds300 - things can be iterative in terms of platform (which got me thinking about designing a nice badge for the docs that indicates platform / cross-platform capability)

PR #75, opened 2018-09-07
[FEATURE elements] Add remaining elements

damassi:
**Follow-up**: https://github.com/artsy/reaction/pull/1216

Moves the remaining elements over from Reaction and into Palette, updated docs, added / enabled CI tests and updated Babel 7.beta to 7.

#### Components updated:  

- Button
- Checkbox
- Image
- Radio
- RadioGroup
- Spinner

**NOTE**: Things will change slightly as we start bringing more universal components online and that this migration is meant to be iterative.

**Next Steps:** 
- Upgrade to styled-components 4 to make use of new `<GlobalStyles>` component, which will simplify how we handle global styles in Reaction.
- Move GlobalStyles from Reaction (`Styleguide/Elements/GlobalStyles`) to Palette once ^ is done 
- Move `<Select />` over to Palette
- Finish updating Reaction, Volt and Rosiland with correct `@artsy/palette` import paths
- Delete `Styleguide/Elements` directory from Reaction 

palette


damassi:
@orta - the docblock linter setting is working well, I like how it fails the build.

.babelrc
```diff
@@ -2,11 +2,33 @@
...
+      {
+        "extensions": [
+          ".ts",
+          ".tsx"
+        ],
+        "root": [
+          "./src"
+        ],
+        "alias": {
+          "palette": "./src/index"
```
damassi: When working around the project `palette` refers to the lib itself, as if you were importing from `@artsy/palette`:

```jsx
import { Box, Theme, Foo, Bar } from 'palette'
```

src/elements/__tests__/Select.test.tsx
```diff
@@ -0,0 +1,47 @@
+// import { selectProps } from "Apps/__test__/Fixtures/Select"
+// import { Boot } from "Artsy/Router"
+// import { mount } from "enzyme"
+// import React from "react"
+// import { LargeSelect, Select, SmallSelect } from "../Select"
+
+describe("Select", () => {
+  it("TODO", () => {
```
damassi: Will cover this in a follow-up pr

zephraph:
Why was the netlify plugin needed? It was already deploying to netlify.

damassi:
@zephraph - sorry that was leftover from a test. Removed

PR #88, opened 2018-09-18
[PATCH avatar] Add xsmall size

oxaudo:
To be used in designs like that: https://app.zeplin.io/project/5ac6790d7f57100518f9ab44/screen/5b606192307c780395c2b26e

### DIFF
--- a/src/elements/Avatar.tsx
+++ b/src/elements/Avatar.tsx
@@ -9,6 +9,8 @@ import { space, SpaceProps } from "styled-system"
 
 const sizeValue = ({ size = "" }) => {
   switch (size) {
+    case "xsmall":
+      return "45px"
     case "small":
       return "70px"
     default:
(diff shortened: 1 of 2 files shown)


src/elements/Avatar.tsx
```diff
@@ -9,6 +9,8 @@ import { space, SpaceProps } from "styled-system"
 
 const sizeValue = ({ size = "" }) => {
   switch (size) {
+    case "xsmall":
```
zephraph: I talked to Brian about the wording on this. He said he'd prefer to keep it to two letter characters (xs, md, lg) but this is good for now. I'll create an issue to update the sizing in a separate ticket.

oxaudo: Updating with the next pr since I forgot to include the `xs` value as a valid value for interface attribute.

I believe reaction is the only consumer of this for now that will have to change with this update, right?

Issue #83, opened 2018-09-13
[RFC] Sidebar Organization


owendodd:
This is an initial proposal of how Palette could be organized. I moved consolidate existing components that have already been added to Palette along with some that live in Reaction and have tried to outline where new components would go.

I'd like to get feedback on this general structure. Along with thoughts on whether there should be any further additions / subtractions to this outline.

I'm currently using this framework for classifying tokens/elements/components.

**Token:** Should primarily be a value, functionally agnostic
**Element:** Composition of various tokens. Specific function, content agnostic.
**Utility:** Specific function, is either not visible or exists temporarily. 
**Component:** Content specific UI. 

Here I am suggesting that a limited number of components should be included in Palette as they are a core part of our site and are regularly used across projects. I would imagine that this list of components would stay fairly fixed. 

---

**Home**
- About
- Change-log

---

**Tokens**
- Color
- Typography
- Spacing
- Grid
- Icons

---

**Elements**
- Images
  - Avatar
  - Image
  - ImageCarousel
  - LightBox
- Layout
  - BorderBox
  - Box
  - Flex
  - StackableBorderBox
  - Separator
  - Join
  - Overflow
  - ReadMore
- Buttons
  - Buttons
  - Text links
- Forms
  - Input
  - QuickInput
  - Radio
  - GroupRadio
  - Checkbox
  - Dropdown
  - MiniDropdown
  - Slider
- Dialogs
  - Banner
  - Modal
  - Alert
  - Message
  - Confirmation
  - Tooltip
  - PopOver
- Navigation
  - Tabs
  - Pagination
- Animation
- Loaders
  - Spinner

---

**Components**
- Header
- Footer
- Bricks
  - ArtistBrick
  - GalleryBrick
  - ShowBrick
  - ArticleBrick
  - ArtworkBrick

damassi:
One edit I would suggest is putting `Flex` at the same level as `Box` (they're definitely close friends around the codebase), and renaming `Box` category header to `Layout` and moving all layout related things inside of it (including `Join`). 

Another suggestion is that `ReadMore` be moved out of utilities since it's a component in practice, and has very specific requirements. (Utilities is somewhat hard to define within a component context.)

owendodd:
@damassi Cool, that all makes sense. Incorporated those changes and collapsed "Utilities"

damassi:
[...]

PR #89, opened 2018-09-18
Add Tooltip Component

### LINKED ISSUE #85: [Addition] Elements / Dialogs / Tooltip
Proposed component: Tooltip component
Responsible Team:  To be built by @jonallured on Sell Team

[View zeplin spec](https://zpl.io/a8NjvL4)

**Large**
Used to provide more information on hover, stated in sentences. Should be used when user hovers on a word or icon.

**Small**
Used to provide more information that is brief (direct, short, not sentences). Alternatively, use if you need to conserve space, and more-info is too large. Should be used when user hovers on a word or icon.

(test — not really sure what other info we want to be adding here)

jonallured:
This is my first Palette component so lmk if I'm missing some stuff!! ❤️ 

I've classed this as WIP mostly because I can't seem to get the alignment of the tooltip to follow the anchor as seen here:

screen shot 2018-09-18 at 1 24 55 pm

I also need to address tapping on mobile being the trigger there and some animation stuff, but would love any direction on this one!!

Closes #85


oxaudo:
Will this replace https://github.com/artsy/reaction/blob/master/src/Components/Tooltip.tsx ?

Seems like we should try to consolidate some things between Reaction/Palette. Otherwise it'll be pretty confusing if both repositories have object with the same name and different functions.

jonallured:
@oxaudo yeah, I think it will. The Tooltip that you linked to doesn't match the new design system Tooltip, so the one I'm working on will ultimately replace it. I chatted with @damassi about this a little yesterday and it sounds like we just don't have the bandwidth to properly deprecate and migrate all components, so we will have a weird in-between period where there will be a Tooltip both in reaction and in palette. Not sure if there's more we can do here! 😕

jonallured:
Ok, after some discussion with @anandaroop I've moved to inline-block with this tooltip and things are looking better in terms of alignment:

screen shot 2018-09-19 at 11 12 54 am
screen shot 2018-09-19 at 11 13 01 am

My biggest problem now is that I had to set the width to 230px in order for it to not be shrunken. That's fine for tips with lots of content, but for that last one it has extra whitespace. I tried playing with different display types and width values but wasn't able to get what I wanted. Open to ideas!

nicoleyeo:
@jonallured is there anywhere I can perform design QA, watch the animation etc? (maybe you can merge once you get the green-light from [eng], and then we can do 1 round of QA on https://artsy-palette.netlify.com/?

The only thing I can see so far is there is a `border` on the tooltip box that doesn't need to be there. 😄

zephraph:
Ahh, good point @nicoleyeo. I'll turn on branch builds...

zephraph:
Deploy preview for *artsy-palette* processing.

Built with commit 97840432fc8216266f4008c43f84fd64e57b64cd

https://app.netlify.com/sites/artsy-palette/deploys/5baa9bca4ed62f73d722c004

jonallured:
@nicoleyeo good call on the border - totally missed that! Fixed with 7fd7abb.

@zephraph thanks for the preview link! I ended up rebasing on top of master so maybe I'll get another one now?

zephraph:
It should rebuild on every change.

zephraph:
Does this conflict w/ Grow's Tooltip? https://github.com/artsy/reaction/blob/master/src/Components/Publishing/ToolTip/ToolTip.tsx

jonallured:
@zephraph I'm not sure! I mean, what I'm working on here is adding the Official Artsy Design System Tooltip ™️ which doesn't match Grow's Tooltip. I think we'll ultimately want both and merging them seems goofy, but maybe I'm wrong. What do you think? Any thoughts @damassi?

src/elements/Tooltip.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    &:hover {
+      opacity: 0;
+    }
+  }
+`
+
+export interface TooltipProps {
+  content: string
+  small: boolean
```
anandaroop: #optional To address your point about the extra whitespace when the text content doesn't fill up the whole 230px — it might be nice if Tooltip took an optional `width` prop to occasionally override the hard-coded 230.

src/elements/Tooltip.tsx
```diff
@@ -0,0 +1,90 @@
...
+      </Wrapper>
+    )
+  }
+}
+
+const formattedTip = tip => {
+  let substring = tip.substring(0, 300)
+
+  if (substring !== tip) {
+    substring += "..."
```
anandaroop: #minor I wonder if design folks would prefer a typographical ellipsis here (i.e. option-semicolon on the Mac)

nicoleyeo:
Official Design QA™

1. I know we discussed previously that it will be difficult to not have tool-tip be cut off on the browser; but I am concerned that this will be cut off frequently. What other options do we have (@other eng?)? (like maybe...a hard rule that enforces the tooltip display right-aligned with edge of type? Example usecase: tool tip on Orders Detail Page, Artsy Collector [?] in right sidebar). Concerned about shipping this with the type cut off. 

2. Can we replace the title "Too much tip component" with "Large" to be more semantically clear ( 😛).

3. (100% feel free to say "no one has time for this!!") Had you seen my note on having y-axis animation? This may have slipped through the Jira WYSIWYG cracks. The desire is for tooltip to animate 5px up on hover (appear-move-up), and animate 4px down on exit (disappear-move-down). Unclear on what kind of available animations we have so far tho, maybe cc @zephraph.

4. Type color (inside tooltip) should be `black60`, looks like it's currently `black100`.

Great work!!!!! Thank you @jonallured.

Process wise — lmk about these QA things. After we're good to go, let me know after you have merged and I'll move this to the "done" column on the [Component Workflow](https://github.com/artsy/palette/projects/1) board. After I move it, we can should announce announce that it's available for use in #design-system (and #frontend). Sound ok?

nicoleyeo:
Ah also note on Grow's tool-tip — that is the "rich tool tip" and has much more functionality than this. From design's perspective, did not feel relevant to directly connect the two beyond simply placing both in `Dialogs` category. Open to other thoughts tho!

zephraph:
Right, right, I was just thinking from a visual perspective. If everyone thinks it's okay though, then I'm 👍

nicoleyeo:
@jonallured a good process thing that came up in design-system sync meeting today; as a follow up to shipping this new tool-tip, can we make a note in Reaction to that this old tooltip is [deprecated](https://artsy-reaction.netlify.com/?selectedKind=Components%2FTooltips&selectedStory=Left%20bottom%20aligned&full=0&addons=1&stories=1&panelRight=0&addonPanel=PeterPanen%2Fstorybook-addon-scissors%2Fpanel)

Just need a cheap, quick note — maybe even just putting (Deprecated) in the title, so other eng know to avoid using it in the future.

damassi:
> (100% feel free to say "no one has time for this!!") Had you seen my note on having y-axis animation? This may have slipped through the Jira WYSIWYG cracks. The desire is for tooltip to animate 5px up on hover (appear-move-up), and animate 4px down on exit (disappear-move-down). Unclear on what kind of available animations we have so far tho

Would like to note that the animation library question resurfaced yesterday, with `react-spring` looking like the winner. I think we should approach animations in a centralized way so that patterns / timings can be established and inherited from before applying them to individual components in a 1-off sort of way. In @owendodd's [Sidebar RFC](https://github.com/artsy/palette/issues/83) there's a section on Animation; might be good to go ahead and add an `Animation.mdx` doc file and start prototyping.

jonallured:
Ok, I _think_ I've addressed all feedback:

* with 16b3749 I was able to use Mathematics to adjust the tooltip and not clip on edge of browser
* switched to using two character size names
* added support for arbitrary width of tooltip
* using the much cooler ellipsis now

What this does not address is animation. I guess I'm sorta taking @damassi's comment above as a nudge that maybe it would be better to wait on this until things mature a little. Hope that's ok!!

Here's a screenshot of that math in action:

screen shot 2018-09-28 at 11 37 41 am

It works on the right side too. Note, however, that this is about the edge of the browser, not the edge of any other containers:

screen shot 2018-09-28 at 11 40 10 am

Does that seem ok? Sorta makes the docs look goofy, but it does work in real usage!

zephraph:
Awesome! I'm looking into the netlify build failures. Doesn't look to be anything related to this code change though. There's an RN related issue.

jonallured:
@zephraph were you able to find anything on this one? Can we try the build again, maybe it was a flicker?

PR #92, opened 2018-09-20
[FIX button] Do not invoke the onClick callback if loading is true

yuki24:
We have seen a case where the user clicks a `<Button>` with `loading: true` and if the user clicks the loader it'll invoke the `onClick` callback again. I was wondering if this is something that should be handled on the palette side rather than having a check in every `onClick` function.


zephraph:
Good point! LGTM.

PR #93, opened 2018-09-22
[FEATURE avatar] Add ability to display initials if no image provided

damassi:
Also added some tests. Note that there's no formal spec for this yet, but wanted to account for various sizes in the text.  

screen shot 2018-09-23 at 2 47 19 pm

### DIFF
--- a/src/elements/Avatar.tsx
+++ b/src/elements/Avatar.tsx
@@ -1,37 +1,104 @@
-// @ts-ignore
-import React from "react"
+import React, { ImgHTMLAttributes, SFC } from "react"
 import styled from "styled-components"
 import { space, SpaceProps } from "styled-system"
+import { color } from "../helpers/color"
+import { SerifSize } from "../Theme"
+import { Flex } from "./Flex"
+import { Serif } from "./Typography"
 
 /**
  * Spec: TODO
  */
 
-const SIZE = { xs: "45px", sm: "70px", md: "100px" }
+interface SizeProps {
+  [key: string]: {
+    diameter: string
+    typeSize: SerifSize
+  }
+}
+
+const Size: SizeProps = {
+  xs: {
+    diameter: "45px",
+    typeSize: "3",
+  },
+  sm: {
+    diameter: "70px",
+    typeSize: "6",
+  },
+  md: {
+    diameter: "100px",
+    typeSize: "8",
+  },
+}
 
 const sizeValue = ({ size = "" }) => {
   switch (size) {
     case "xs":
-      return SIZE.xs
+      return Size.xs
     case "sm":
-      return SIZE.sm
+      return Size.sm
     case "md":
     default:
-      return SIZE.md
+      return Size.md
   }
 }
 
-export interface AvatarProps extends SpaceProps {
+export interface AvatarProps extends ImgHTMLAttributes<any>, SpaceProps {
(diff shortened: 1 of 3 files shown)


src/elements/Avatar.tsx
```diff
@@ -1,37 +1,102 @@
...
 
 /**
  * Spec: TODO
  */
 
-const SIZE = { xs: "45px", sm: "70px", md: "100px" }
+interface SizeProps {
+  [key: string]: {
+    diameter: string
+    typeSize: SerifSize
```
oxaudo: #superMinor - but maybe this second prop would be more descriptive if it was named `serifSize`

It is much more clear already to put it into interface here!

src/elements/Avatar.tsx
```diff
@@ -1,37 +1,102 @@
...
+import { SerifSize } from "../Theme"
+import { Flex } from "./Flex"
+import { Serif } from "./Typography"
 
 /**
  * Spec: TODO
  */
 
-const SIZE = { xs: "45px", sm: "70px", md: "100px" }
+interface SizeProps {
```
oxaudo: This capitalizing const from Ruby is still alive in me:)

src/elements/Avatar.tsx
```diff
@@ -1,37 +1,102 @@
...
-export interface AvatarProps extends SpaceProps {
+export interface AvatarProps extends ImgHTMLAttributes<any>, SpaceProps {
+  /** If an image is missing, show initials instead */
+  initials?: string
   /** The size of the Avatar */
   size?: "xs" | "sm" | "md"
 }
 
+/** An circular Avatar component containing an image or initials */
+export const Avatar: SFC<AvatarProps> = ({ src, initials, size = "sm" }) => {
```
oxaudo: Should not it default to `md` since it's the default for the props? Or is it changed intentionally here?

src/elements/Avatar.tsx
```diff
@@ -1,37 +1,102 @@
...
+  if (src) {
+    return <AvatarImage size={size} src={src} />
+  } else if (initials) {
+    const { diameter, typeSize } = sizeValue({ size })
+    const justifyContent = initials.length > 4 ? "left" : "center"
+
+    return (
+      <InitialsHolder
+        width={diameter}
```
oxaudo: 👍 for making it dynamic!

src/elements/Avatar.tsx
```diff
@@ -1,37 +1,102 @@
...
+        height={diameter}
+        justifyContent={justifyContent}
+        alignItems="center"
+      >
+        <Serif
+          size={typeSize}
+          color="black60"
+          weight="semibold"
+          lineHeight={45}
```
oxaudo: Maybe this can also be `diameter` so it perfectly aligns with the holder?

src/elements/Avatar.tsx
damassi: I think by using `serifSize` it might tie the component to the implementation a little too tightly; with `typeSize` we'd be able to swap out the font-face at a later date without worrying about updating callsites.

src/elements/Avatar.tsx
damassi: Ohh, good catch, will update 👍

PR #95, opened 2018-09-25
[Docs] Update router to use hash for deeplinking

damassi:
Updates routes to support a `/#/` so that deeplinking works. Can flip this off once we figure out how to support dynamic urls.

### DIFF
--- a/doczrc.js
+++ b/doczrc.js
@@ -1,4 +1,5 @@
 export default {
+  hashRouter: true,
   indexHtml: "docs/index.html",
   protocol: "http",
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 9413f85e57064574c0c32e7d00291f3f5273a143

https://deploy-preview-95--artsy-palette.netlify.com

zephraph:
It doesn't _completely_ fix deep linking... we stil lhave issues with sub items. Still, it's better than what we have.

PR #96, opened 2018-09-26
feat: add disabled prop to select

javamonn:
Adds a disabled prop and state to `LargeSelect`. I'm not aware of an official spec for this, but I followed the design seen [here](https://app.zeplin.io/project/5ab94282c124db8d58365b47/screen/5b7c5afcd2b23b7c8a0c3a2e) and followed the disabled state of `Radio` for how to handle the hovered disabled state - `default` cursor.

### DIFF
--- a/src/elements/Select.tsx
+++ b/src/elements/Select.tsx
@@ -17,6 +17,7 @@ interface Option {
 export interface SelectProps extends PositionProps, SpaceProps {
   options: Option[]
   selected?: string
+  disabled?: boolean
   onSelect?: (value) => void
 }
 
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit c7b138b19a77a83d271787af8270dd28b28705d4

https://deploy-preview-96--artsy-palette.netlify.com

src/elements/Select.tsx
```diff
@@ -89,10 +91,11 @@ const hideDefaultSkin = css`
   }
 `
 
-const caretArrow = css`
+const caretArrow = css<SelectProps>`
```
zephraph: Hmm... this works now? Nice. Is that a typescript 3 thing? 🤔

zephraph:
This looks good to me.. @nicoleyeo or @briansw could either of you just do a quick sanity check? https://deploy-preview-96--artsy-palette.netlify.com/#/docs-select

src/elements/Select.tsx
javamonn: I'm not sure! I get a type error without it and followed existing patterns, though we haven't had to pass a type argument to `css` exactly like this before. Perhaps @damassi could chime in?

damassi: Yup this works with latest styled components!

PR #97, opened 2018-09-26
[PATCH Typography] added width inherit property to Text component

ashleyjelks:
Noticed that Text content was not filling parent component's full width, so added inherit prop

### DIFF
--- a/src/elements/Typography.tsx
+++ b/src/elements/Typography.tsx
@@ -91,6 +91,7 @@ export const Text = primitives.Text.attrs<TextProps>({})`
   ${fontFamilyHelper};
   font-size: ${({ fontSize }) => fontSize}px;
   line-height: ${({ lineHeight }) => lineHeight}px;
+  width: inherit;
   ${color};
   ${display};
   ${maxWidth};


damassi:
Nice first palette PR 👍  

merge on green

artsyit:

Built with commit 6dd85903ecaba7cba147da8efa8ccf2579817424

https://deploy-preview-97--artsy-palette.netlify.com

jonallured:
Hello friends! While working through this one:

https://artsyproduct.atlassian.net/browse/SELL-983

I noticed that `width: inherit` was added and tracked it down to this PR. Here's the structure of what we've got over on Volt on the orders list page:

```xml
<OrderRow>
  <CellLayout>
    <Sans>
      <Truncate>OMGHAHA</Truncate>
    </Sans>
  <CellLayout>
  ...more cells
</OrderRow>
```

The `OrderRow` is `display: flex` and the `CellLayout` is `width: 8%`, which then gets inherited by `Sans` with this PR. Because the `Truncate` is there, we're getting all the text smushed into `...` even though there is plenty of room:

screen shot 2018-10-03 at 8 54 55 am
 
So, my first question is what's a good way to fix this, but my bigger question is about the safety of upgrading palette, done on Volt recently here: https://github.com/artsy/volt/pull/3242. When something as small as adding a single line of CSS can break an entire page, I get a little twitchy! 😝 

/cc @anandaroop @zephraph

ashleyjelks:
Adding @damassi to the loop on this....

@jonallured For now I will revert the change in Palette and either make this change using inline CSS or speak with Chris to see if he has a better suggestion as I understand we are trying to move away from this and towards using Palette for styling.

jonallured:
Yeah, I don't know if a revert is the right way to go - maybe `Sans` should inherit width, I don't feel strongly one way or the other. After tinkering with this, I realized that I could wrap an empty div and be back in business:

```xml
<OrderRow>
  <CellLayout>
    <div>
      <Sans>
        <Truncate>OMGHAHA</Truncate>
      </Sans>
    </div>
  <CellLayout>
  ...more cells
</OrderRow>
```

But I'm def interested to hear people's thoughts about how customers of palette can be confident in upgrading! Maybe it's on Volt to implement visual regression testing? 🤷‍♂️

damassi:
@jonallured - sorry about that! I think the only way to catch this stuff is through visual snapshot tests :/ We're getting there! 

@ashleyjelks - we can revert this for now and add an inline style or branstorm a fix. After i'm off this meeting I'll back out of this change.

PR #99, opened 2018-09-27
add styling to border

erikdstock:
This PR adds border styling to our `Select` html elements.
Spec: https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5ace8399553bdae0668efe69
Tracked in https://artsyproduct.atlassian.net/browse/PURCHASE-504
cc @damassi 

The gif below doesn't capture it well but the `border-transition` property is working identically to the other components (whose borders are controlled by reaction).

_last

(from `yarn link`ed in reaction )storybook:
address

### DIFF
--- a/docs/Select.mdx
+++ b/docs/Select.mdx
@@ -89,3 +89,30 @@ import { Value } from 'react-powerplug'
     )}
   </Value>
 </Playground>
+
+## Error State
+
+<Playground>
+  <Value>
+    {({ value="bad", set }) => {
+      const error = value === "good" ? "" : "bad value"
+      return (
+        <LargeSelect
+          error={error}
+          options={[
+            {
+              text: "Good",
+              value: "good",
+            },
+            {
+              text: "Bad",
+              value: "bad",
+            },
+          ]}
+          selected={value}
+          onSelect={value => set(value)}
+        />
+      )
+    }}
+  </Value>
+</Playground>
--- a/docs/Select.mdx
+++ b/docs/Select.mdx
@@ -94,19 +94,19 @@ import { Value } from 'react-powerplug'
 
 <Playground>
   <Value>
-    {({ value="bad", set }) => {
-      const error = value === "good" ? "" : "bad value"
+    {({ value="badValue", set }) => {
+      const error = value === "goodValue" ? "" : "Error Message"
       return (
         <LargeSelect
           error={error}
           options={[
             {
               text: "Good",
-              value: "good",
+              value: "goodValue",
             },
             {
               text: "Bad",
-              value: "bad",
+              value: "badValue",
             },
           ]}
           selected={value}
--- a/src/elements/Select.tsx
+++ b/src/elements/Select.tsx
@@ -18,6 +18,7 @@ export interface SelectProps extends PositionProps, SpaceProps {
   options: Option[]
   selected?: string
   disabled?: boolean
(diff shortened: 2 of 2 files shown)


artsyit:

Built with commit d451fd9bd1683070d1c18395b71de5b0b320ef8a

https://deploy-preview-99--artsy-palette.netlify.com

sweir27:
👏 👏 👏

src/elements/Select.tsx
```diff
@@ -108,10 +109,16 @@ const LargeSelectContainer = styled.div.attrs<SelectProps>({})`
...
+    border: 1px solid
+      ${({ error }) => (error && color("red100")) || color("black10")};
     border-radius: 0;
+    transition: border-color 0.25s;
     padding-right: ${space(1)}px;
     ${styledSpace};
+    &:hover,
+    &:focus {
+      border-color: ${color("purple100")};
+    }
```
damassi: 👍

damassi:
(When ready to merge be sure to update the commit message so that a deploy goes out)

src/elements/Select.tsx
```diff
@@ -17,6 +17,7 @@ interface Option {
 export interface SelectProps extends PositionProps, SpaceProps {
   options: Option[]
   selected?: string
+  error?: string
```
damassi: Can you add an example to the Select docs showing the error state?

docs/Select.mdx
```diff
@@ -89,3 +89,30 @@ import { Value } from 'react-powerplug'
     )}
   </Value>
 </Playground>
+
+## Error State
+
+<Playground>
+  <Value>
+    {({ value="bad", set }) => {
+      const error = value === "good" ? "" : "bad value"
```
damassi: Should `"bad value" be "bad"? 

I think rather than "good" and "bad" semantics "Error" / "error" might be more appropriate since it maps to the prop thats being passed.

erikdstock: How about "This is an error message set by the onChange handler"?

damassi: 😄  either way, the code is short and concise, it probably doesn't really matter much. My thinking about cleaning up bad is that other users would then see  "bad", "bad value", "error" and good and get confused documentation-wise. I think "error" is sufficient since its 1:1

src/elements/Select.tsx
```diff
@@ -18,6 +18,7 @@ export interface SelectProps extends PositionProps, SpaceProps {
   options: Option[]
   selected?: string
   disabled?: boolean
+  error?: string
```
damassi: Should `error` be a boolean?

damassi: If error will eventually display message it might be clearer to label this prop `errorMsg` and then check against that, with a little `TODO: Implement ErrorMessage component` somewhere.

erikdstock: happy to adapt this however is good- but in reaction Input takes an ` error?: string`. Your thoughts?

damassi: Ahh cool cool, this seems fine for now 👍 It would be good to add a `TODO` though so its clear that we haven't yet hooked up the message part, which will eventually display the string that's passed into this prop.

zephraph:
Sounds like everything is good to go. Can always follow up w/ the minor items if needed. 

LGTM!

PR #103, opened 2018-10-02
[PATCH select] make arrow clickable

l2succes:
Simple fix to move the arrow into the target area and vertically aligning it with the label on the `Select` component. As of now it's a bit misleading as it's outside of the target area.

### DIFF
--- a/src/elements/Select.tsx
+++ b/src/elements/Select.tsx
@@ -92,10 +92,7 @@ const hideDefaultSkin = css`
   }
 `
 
-const caretArrow =
-  css <
-  SelectProps >
-  `
+const caretArrow = css<SelectProps>`
   border-left: 4px solid transparent;
   border-right: 4px solid transparent;
   border-top: 4px solid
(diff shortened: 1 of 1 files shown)


src/elements/Select.tsx
```diff
@@ -155,8 +152,8 @@ const SmallSelectContainer = styled.div.attrs<SelectProps>({})`
     cursor: pointer;
     pointer-events: none;
     position: absolute;
-    top: 12px;
-    margin-left: 8px;
+    top: 10px;
+    margin-left: -8px;
```
damassi: this damn thing 🤕 The whole select component could probably be rewritten... Not my best code.

artsyit:

Built with commit d370bd935c793c41c03501f6856a276ba69dca95

https://deploy-preview-103--artsy-palette.netlify.com

src/elements/Select.tsx
l2succes: Just realized, we also need to account for the width of the select element as well

l2succes: All good, we do what we can with the time we have ;)
I'll take a stab at cleaning things up a bit.

l2succes: It's actually not bad at all, we just gotta account for the arrow, that's the only issue with it.

damassi:
@l2succes @xtina-starr - going to merge this since some restructuring is coming down the pipe

src/elements/Select.tsx
damassi: @l2succes - can you elaborate? We can probably get away with not having to use JS measurements and rely only on CSS, yah? (This will also need to be SSR capable)

PR #107, opened 2018-10-03
Revert "[PATCH Typography] added width inherit property to Text component"

damassi:
Reverts artsy/palette#97


damassi:
@jonallured caught something downstream so we'll back out of this for now since its not imperative.

Issue #110, opened 2018-10-05
[NEW] Elements / Dialogs / Banner


nicoleyeo:
## New Component: Banner

**Product team** Sell Team
**Design**   @nicoleyeo
**Eng** @jonallured 

[View zeplin link
](https://zpl.io/adkjxqp)
**Additional info**
Banners are used to convey time bound messages, statuses and specfic actions to our users. Depending on the timespan and importance of the message, they can either be a permanent part of the page, or can be dismissable. 

This build of the banner only includes the "error" state. In the future (when work is appropriately prioritized), we want to build upon this banner to include other states, such as: success, notification, warning.

**Context that initiates build**
We are building this to display a few errors in CMS when a gallery is managing orders (confirming, etc.)

jonallured:
Closed with #114

PR #114, opened 2018-10-11
Add banner

jonallured:
This PR is for #110 and represents some messing around with how we might do SVGs. Open to thoughts and ideas!! ❤️ 

Here's some shots of the component:

screen shot 2018-10-16 at 8 32 48 am

screen shot 2018-10-16 at 8 33 04 am

While working on this I explored an idea I had about getting SVG images into Palette. I made `src/svgs` and then created `CloseIcon` in there for the purposes of completing the Banner component. Then, I took a stab at pulling out the other SVG in the Checkbox element.

From there, I added a page for Icons that looks like this:

screen shot 2018-10-15 at 4 23 22 pm

So this is one way we could do it. Would love feedback.

### DIFF
--- a/src/elements/Banner/Banner.tsx
+++ b/src/elements/Banner/Banner.tsx
@@ -5,20 +5,31 @@ import { space } from "../../helpers"
 import { Box } from "../Box"
 import { Sans } from "../Typography"
 
+import { CloseIcon } from "../../svgs"
+
 const Target = styled.div`
   height: 30px;
+  padding: 9px;
   position: absolute;
-  right: 11px;
-  top: 11px;
+  right: 15px;
+  top: 15px;
   width: 30px;
 
   &:hover {
     cursor: pointer;
   }
+
+  svg {
+    display: block;
+  }
 `
 
 const CloseButton = ({ onClick }) => {
-  return <Target onClick={onClick}>X</Target>
+  return (
+    <Target onClick={onClick}>
+      <CloseIcon fill="white100" />
+    </Target>
+  )
 }
 
 export interface BannerProps {
@@ -50,11 +61,13 @@ export class Banner extends React.Component<BannerProps> {
       <Box
         bg="red100"
         color="white100"
-        p={space(2)}
+        height={space(6)}
         position="relative"
         textAlign="center"
       >
-        <Sans size="2">{this.props.message}</Sans>
+        <Sans size="2" style={{ lineHeight: "60px" }}>
+          {this.props.message}
+        </Sans>
         {showCloseButton && <CloseButton onClick={this.handleCloseClick} />}
       </Box>
     )
--- a/src/elements/Banner/Banner.tsx
+++ b/src/elements/Banner/Banner.tsx
@@ -1,7 +1,6 @@
 import React from "react"
 import styled from "styled-components"
 
-import { space } from "../../helpers"
 import { Box } from "../Box"
 import { Sans } from "../Typography"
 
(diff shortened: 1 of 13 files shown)


artsyit:

Built with commit 485298ab94156cd558a7b124bd4bbb927c8859a8

https://deploy-preview-114--artsy-palette.netlify.com

src/elements/Banner/Banner.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <Target onClick={onClick}>
+      <CloseIcon fill="white100" />
+    </Target>
+  )
+}
+
+export interface BannerProps {
+  dismissable: boolean
+  message: string
```
damassi: If the design spec doesn't call for it it's arguable we should support it at this time, but making this 

```
message: string | JSX.Element
```

would allow us to pass in markup to the prop.

src/elements/Banner/Banner.tsx
```diff
@@ -0,0 +1,74 @@
...
+  render() {
+    if (this.state.dismissed) return null
+    const showCloseButton = this.props.dismissable
+
+    return (
+      <Box
+        bg="red100"
+        color="white100"
+        p={space(2)}
+        pr={space(6)}
```
damassi: No need for the helper, can pass in a number directly

src/elements/Banner/Banner.tsx
jonallured: Hmm, interesting. On the one hand the ability to pass markup makes this component more flexible but on the other, it creates an opportunity to complicate - for instance, the spec does not specify link styles. I think I'm going to leave it for now, but we'll see how this evolves.

src/elements/Banner/Banner.tsx
jonallured: Good call, fixed with 4386bf7.

nicoleyeo:
@jonallured looks great to me! Thank you for being flexible with this solution 💪🏼

jonallured:
Ok merging, thanks pals!!

src/elements/__tests__/Banner.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+  })
+
+  it("disappears when dismissed", () => {
+    const message = "There was an error."
+    const wrapper = mount(<Banner dismissable message={message} />)
+    expect(wrapper.find("CloseButton")).toHaveLength(1)
+    wrapper.find("CloseButton").simulate("click")
+    expect(wrapper.find("CloseButton")).toHaveLength(0)
+  })
+})
```
damassi: Thank you! 👍

jonallured: Yeah, I'm glad you said something - I didn't notice tests while poking around the first time because I'm so used to seeing them in the same folder as the implementation! When looking again, I noticed the `__tests_` folders and figuring out the precedent from looking through them. 👍

damassi: haha! @jonallured - just went and looked and actually your message helped me catch something -- I realized I didnt move the tests when i created folders for stuff, will update today.

oxaudo:
Just catching up here. Nice to see new components added here and progress on making svgs more findable/reusable.

One thing surprised me here is that the color for this banner is hardcoded inside the component. Seems like generic `Banner` should support not only error red color. But I guess nothing is wrong with adding it once we need it.

Nice stuff @jonallured!

jonallured:
@oxaudo yep, you're totally right and the spec in Zeplin actually mentions this:

https://zpl.io/adkjxqp

But I felt like I'd just wait and we can circle back to the colors topic the next time we use this sucker and need a different color. 💪

PR #116, opened 2018-10-16
[Toolchain] Co-locate tests with components 

damassi:
Move tests inline with sources and flatten out `/svg` directory.


jonallured:
Netlify failed, but that's a known issue so I'm merging, thanks for doing this!! 🏆

PR #126, opened 2018-11-08
Add Collapse component to Palette

sepans:
Moves `Collapse` component from Reaction to Palette to be used by **volt**.

This is exactly the copy paste of the Reaction [implementation](https://github.com/artsy/reaction/blob/master/src/Styleguide/Components/Collapse.tsx).

@yuki24 brought up a point that the palette repo is meant to be very light in terms of dependency but `react-spring` that is used for collapsing is a relatively large one so:

- If this kind of animations will be used in future palette components we can keep `react-spring` here
- Or use an alternative library or implementation
- Or keep `Collapse` in reaction and import it from there into volt, although reaction is becoming a very heavy dependency itself, to be imported as a component library.

screen shot 2018-11-08 at 4 29 53 pm


package.json
```diff
@@ -87,6 +87,7 @@
   "dependencies": {
     "rc-slider": "^8.6.2",
     "react": "^16.5.0",
+    "react-spring": "^5.7.2",
```
damassi: Palette will have an animation section fairly shortly so it was expected that this dependency would be added in time.

src/elements/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react"
+import { animated, Spring } from "react-spring"
+
+export interface CollapseProps {
+  /** is open? */
```
damassi: `/** Determines whether content is expanded or collapsed /*`

artsyit:

Built with commit 4ed412bc8641363431b98152bfbcc7d4fe1eea01

https://deploy-preview-126--artsy-palette.netlify.com

src/elements/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,35 @@
...
+  open: boolean
+}
+
+/** Collapse component */
```
damassi: This comes through in VSCode tooling and eventually docs; mind expanding this decription a bit?

docs/Collapse.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+# Collapse
+
+## Open
+
+<Playground>
+    <Collapse open>
+      <BorderBox>
+        A Collapse component ...
+        Not really cool yet!
```
damassi: It's cool!

damassi:
Ahh @sepans - message is case sensitive: 

`[FEATURE Collapse] Add Collapse component to Palette`

PR #130, opened 2018-11-21
Add Link element

oxaudo:
This is an attempt to follow definitions from here: https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5afc758c09a18d3d0f176edd

The code here is just making Link a `styled.a` that can potentially take noUnderline property or color property. The options generated are here:
screen shot 2018-11-21 at 4 19 12 pm


damassi:
If it helps at all note that the base implementation for this is located in our GlobalStyles in Reaction: https://github.com/artsy/reaction/blob/master/src/Styleguide/Elements/GlobalStyles.tsx 

(This eventually needs to move over to palette.)

oxaudo:
@damassi - Thanks for the link! I basically added very similar definitions here. Not sure if/when we could remove the other one, because I'm pretty sure we use `a` tag directly in Reaction a lot. The idea here is to replace current usage where `Link` is redefined in Reaction with this Link implementation.

@zephraph - Added a placeholder for ios implementation. Seems like in all the other cases it's just left blank for now. Doing that here as well.

I think I'm ready to test it in Reaction in my super bare implementation and we can definitely iterate here going forward.

damassi:
Going to merge this since i'll be using it in a sec!

PR #135, opened 2018-11-28
[FEATURE ios] Add Collapse implementation

javamonn:
As part of ongoing Local Discovery work, we needed an animated `Collapse` implementation and wanted to extend the existing implementation in Palette. We ran into a couple issues while building this out:

- `react-spring` has a `peerDependency` [on `react` at `>=16.4.0`](https://github.com/drcmda/react-spring/blob/v5.9.2/package.json#L111). It looks like it depends on [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html), a feature added in React 16.3. Emission is on `16.3.0-alpha.1` and appears to not have this feature, causing `react-spring` to throw an error. I'm opting to pin `react-spring` via `resolutions` in emission at 5.5 (the latest working version) rather than downgrade the version used in palette as we are in the process of upgrading React Native and could upgrade seamlessly in the future and there are some fixes to the `height: auto` animation used in the web implementation that we'd lose if we downgraded.
- `react-spring`'s `height: auto` animation target support doesn't work on RN, so we have to manually measure the view to derive a static height prior to animating. I cribbed much of this logic from [`react-collapsible`](https://github.com/oblador/react-native-collapsible/blob/master/Collapsible.js). It looks like we could upstream this to `react-spring` based on the same logic used for web support: https://github.com/drcmda/react-spring/blob/master/src/targets/web/fix-auto.js

### DIFF
--- a/src/elements/Collapse/Collapse.ios.tsx
+++ b/src/elements/Collapse/Collapse.ios.tsx
@@ -1,0 +1,145 @@
+import React from "react"
+import { View } from "react-native"
+import { animated, Spring } from "react-spring/dist/native.cjs.js"
+
+const AnimatedView = animated(View)
+
+export interface CollapseProps {
+  /** Determines whether content is expanded or collapsed */
+  open: boolean
+  /**
+   * If we're rendering within a statically-sized component (e.g. FlatList), we need
+   * to propagate a sentinel value in order to trigger re-render or re-measure.
+   */
+  onAnimationFrame: (animateValue: { height: number }) => void
+}
+
+interface State {
+  isMounted: boolean
+  hasMeasured: boolean
+  isMeasuring: boolean
+  isAnimating: boolean
+  measuredHeight?: number
+}
+
+/** Collapses content with animation when open is not true */
+export class Collapse extends React.Component<CollapseProps, State> {
+  measureRef: View
+
+  state: State = {
+    isMounted: false,
+    isMeasuring: false,
+    isAnimating: false,
+    hasMeasured: false,
+  }
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 5d9cf21d5c7eb7a811648b08d7d5e968e37278f7

https://deploy-preview-135--artsy-palette.netlify.com

src/elements/Collapse/Collapse.ios.tsx
```diff
@@ -0,0 +1,143 @@
...
+    if (isMeasuring) {
+      return (
+        <View
+          ref={this.handleMeasureRef}
+          style={{ opacity: 0, position: "absolute" }}
+        >
+          {children}
+        </View>
+      )
+    }
```
damassi: This might be a little clearer if it were moved into a method and then called like
```js
if (isMeasuring) {
  return this.measureView() 
} 
...
```

zephraph:
This might be one of those cases where we can explore the idea of how we can share an interface across components. This is a really minimal case where `open` is the only prop served between both of them. I don't really see it as necessary, just raising the idea. 

Thoughts?

javamonn:
This would be a good simple case to test out sharing an interface. My concern though is that we don't currently have a way to test it from the perspective of Emission, as we're still consuming only web types as noted [here](https://github.com/artsy/palette/pull/109#issuecomment-436417935). I might open an issue for this along with a note about figuring out shared and disjoint component props interfaces.

l2succes:
Right, even before we get type safety on the emission side with respect to not exposing web types on RN components. 
If we can figure out a way to share the Props interface between React native web and mobile we'd be in a good spot.

PR #138, opened 2018-11-30
[DO NOT MERGE | BLOCKED] Make typography element customizable

zephraph:
I was working on https://github.com/artsy/reaction/pull/1608 which requires certain typography components be `h1`s or `h2`s. In that PR I just wrapped said typography components in elements with the styles reset. That really doesn't feel like a great way to handle this though. 

Looking through the styled-components docs, styled-components 4 supports an [`as` prop](https://www.styled-components.com/docs/api#as-polymorphic-prop) which can be used to change the element. Turns out it's relatively straight forward to get this working with our typography components. 

Unfortunately this is blocked due to the same issue that @javamonn mentioned [here](https://github.com/artsy/palette/pull/135#issue-234386685). We'll need to upgrade react-native before we proceed... and styled-components across all the projects. Maybe a non-trivial amount of work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@types/node": "^10.3.0",
     "@types/react": "^16.4.13",
     "@types/react-test-renderer": "^16.0.1",
-    "@types/styled-components": "^3.0.0",
+    "@types/styled-components": "^4.1.2",
     "@types/styled-system": "^3.0.7",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^23.0.1",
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit a79a496bc762a93e297cdfff2170a41409a58480

https://deploy-preview-138--artsy-palette.netlify.com

artsyit:

Built with commit f0c9b80bae6822535c889adedb1f41fdf5804da0

https://deploy-preview-138--artsy-palette.netlify.com

zephraph:
Looks like there are some type issues after all 😅

alloy:
Easy peasy 👍

package.json
```diff
@@ -80,7 +80,7 @@
     "regenerator-runtime": "^0.11.1",
     "semantic-release": "^15.5.0",
     "static-extend": "^0.1.2",
-    "styled-components": "^3.4.5",
+    "styled-components": "^4.1.2",
```
damassi: My memory is a bit fuzzy, but for some reason I feel like we need to take a hard look at updating force to the latest styled-components before we can implement elsewhere.

PR #139, opened 2018-12-04
[PATCH banner] adds backgroundColor prop to Banner

kierangillen:
- Adds backgroundColor to banner component for transitions
- Possibly needs some animation added as well, need to check with design team

### DIFF
--- a/src/elements/Banner/Banner.tsx
+++ b/src/elements/Banner/Banner.tsx
@@ -34,6 +34,7 @@ const CloseButton = ({ onClick }) => {
 export interface BannerProps {
   dismissable: boolean
   message: string
+  backgroundColor?: string
 }
 
 /**
@@ -42,6 +43,7 @@ export interface BannerProps {
 export class Banner extends React.Component<BannerProps> {
   static defaultProps = {
     dismissable: false,
+    backgroundColor: null,
   }
 
   state = {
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 7a01a2438206914bc8cd6fdedf6acaafd2074f28

https://deploy-preview-139--artsy-palette.netlify.com

src/elements/Banner/Banner.tsx
```diff
@@ -34,6 +34,7 @@ const CloseButton = ({ onClick }) => {
 export interface BannerProps {
   dismissable: boolean
   message: string
+  backgroundColor?: string
```
damassi: Would you mind updating [the test](https://github.com/artsy/palette/blob/7a01a2438206914bc8cd6fdedf6acaafd2074f28/src/elements/Banner/Banner.test.tsx) with this new API option?

jonallured:
Nice to see this little component learning some new tricks! 👍 

Here's what I'm thinking:

* Add a case to this file: docs/Banner.mdx maybe called something like "Status with Custom Background" and then demonstrate overriding the default.
* Since the default color is that red, let's promote that up into the `defaultProps` bit and that'll make the property required not optional.
* Pass `this.props.backgroundColor` directly to `Box` rather than creating a local for it in the render.

How does that sound to you?

src/elements/Banner/Banner.tsx
```diff
@@ -23,6 +23,10 @@ const Target = styled.div`
   }
 `
 
+const Wrapper = styled(Box)`
+  transition: background-color 250ms linear;
```
starsirius: Do we want to transition the background color? I might miss the conversations, but this feels a new UI pattern worth checking in with design.

starsirius:
As @damassi mentioned, let's add a test in https://github.com/artsy/palette/blob/master/src/elements/Banner/Banner.test.tsx and test the option!

starsirius:
Nice one! 🥇

PR #141, opened 2018-12-07
[WIP] Take screenshots via Jest puppeteer

orta:
Uses [jest-puppeteer-react](https://github.com/Hapag-Lloyd/jest-puppeteer-react) to combine [jest-image-snapshot](https://github.com/americanexpress/jest-image-snapshot) with WebPack to take a screenshot of the image at runtime headless-ly.

It uses the examples from inside the docz files as the sources of truth, so no new process to add the images. 

This technique also leaves us open to automating the React Native screenshots using the same type of code. 

--- 

Blockers:

 - [ ] Need to get any React component rendering

Nice to haves:

 - [ ] Name the snapshots after a preceding header in the AST


jest-puppeteer-react.config.js
```diff
@@ -0,0 +1,76 @@
...
+        alias.family === "IPv4" &&
+        alias.address !== "127.0.0.1" &&
+        !alias.internal
+      )
+        return alias.address
+    }
+  }
+
+  return "0.0.0.0"
+}
```
damassi: This is necessary for running against localhost?

orta: No idea, came with the template ( it runs inside a docker container ) so probably is

damassi:
So much has changed since this has opened it would be good to revisit and see what we can do with Gatsby. Closing for now, and we can re-open if need be.

PR #142, opened 2018-12-07
#minor fix to docs

oxaudo:
Just wanted to see what we currently have as a tooltip and wanted it to load:)

But the real question I have is - wondering how to do this more permanent closable popup like the one that is defined here: https://app.zeplin.io/project/5b113f413ede4bb10dbfdc42/screen/5b354b3f925ef06d24e323f0

Want to have the wrapper style and positioning logic from tooltip essentially but add close button and support for more elaborate content.

I think for now I would just copy needed things but we should probably have a more generic style for positioning and bordering that can be included wherever.

### DIFF
--- a/docs/Tooltip.mdx
+++ b/docs/Tooltip.mdx
@@ -11,7 +11,7 @@ import { Box, Sans, Tooltip } from '../src'
 
 <Playground>
   <Box textAlign="left" mt={100}>
-    <Sans>
+    <Sans size="2">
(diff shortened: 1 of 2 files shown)


jonallured:
I agree - copy and paste what you need and if/when we feel like there's some higher abstraction worth making, then we can refactor. I'm a big fan of adhering to the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) so I think we're safe to copy paste here and not even make an abstraction. 😄

PR #144, opened 2018-12-11
Remove the purple border from <Link>

yuki24:
While the border is helpful when focusing on a link and using the tab key to move forward, it looks too disruptive for most of our default
links. Let's just remove it for now and re-visit when we think we really need it.

Also adding href to <Link> otherwise browsers do not apply default styles to link tags.

### DIFF
--- a/src/elements/Link/Link.tsx
+++ b/src/elements/Link/Link.tsx
@@ -13,9 +13,7 @@ export interface LinkProps extends SpaceProps {
  * Spec: https://zpl.io/2Gm6D3d
  */
 export const Link = styled.a<LinkProps>`
-  ${space};
-  ${styledColor};
-  cursor: pointer;
+  color: ${color("black100")};
   transition: color 0.25s;
   text-decoration: ${props =>
     props.noUnderline || props.color ? "none" : "underline"};
@@ -24,7 +22,8 @@ export const Link = styled.a<LinkProps>`
     color: ${color("black100")};
   }
   &:focus {
-    border: 1px solid ${color("purple100")};
     color: ${props => (props.color ? color(props.color) : color("black100"))};
   }
+  ${space};
+  ${styledColor};
 `
--- a/docs/Link.mdx
+++ b/docs/Link.mdx
@@ -9,13 +9,13 @@ import { Box, Link } from '../src'
 
 <Playground>
   <Box>
-    <Link>This is a default link</Link>
+    <Link href="#">This is a default link</Link>
   </Box>
   <Box>
-    <Link noUnderline>This is a non-underlined link</Link>
+    <Link href="#" noUnderline>This is a non-underlined link</Link>
   </Box>
   <Box>
-    <Link color="purple100">This is a non-standard color link</Link>
+    <Link href="#" color="purple100">This is a non-standard color link</Link>
   </Box>
 </Playground>
 


src/elements/Link/Link.tsx
```diff
@@ -24,7 +22,8 @@ export const Link = styled.a<LinkProps>`
     color: ${color("black100")};
   }
   &:focus {
-    border: 1px solid ${color("purple100")};
     color: ${props => (props.color ? color(props.color) : color("black100"))};
   }
+  ${space};
+  ${styledColor};
```
yuki24: We need those at the end otherwise they don't override CSS properties as expected.

artsyit:

Built with commit 2e96248561306157e7a46878d64ffcf8f71b91e1

https://deploy-preview-144--artsy-palette.netlify.com

src/elements/Link/Link.tsx
```diff
@@ -24,7 +22,8 @@ export const Link = styled.a<LinkProps>`
     color: ${color("black100")};
   }
   &:focus {
-    border: 1px solid ${color("purple100")};
```
anandaroop: Curious about the accessibility implications of this — will e.g. keyboard-only users get enough of a visual affordance when tabbing through links?

yuki24: It was very helpful to see the border as an indicator when only using the keyboard for navigating, but the border also looked more disruptive for common use cases than being accessibility-friendly. So even just a normal click was adding a border:

clicking-link

Perhaps @nicoleyeo @briansw can comment on this?

src/elements/Link/Link.tsx
oxaudo: 👍

src/elements/Link/Link.tsx
```diff
@@ -13,9 +13,7 @@ export interface LinkProps extends SpaceProps {
  * Spec: https://zpl.io/2Gm6D3d
  */
 export const Link = styled.a<LinkProps>`
-  ${space};
-  ${styledColor};
-  cursor: pointer;
```
oxaudo: I think we might need cursor: pointer just to be on a safe side. We are often use `Link` with just onClick assignment. I think the underline might be lost in this case.

src/elements/Link/Link.tsx
oxaudo: 👍

src/elements/Link/Link.tsx
oxaudo: One thing with this border is that in case we need to keep it - we might be forced to also add a transparent border for all other states to prevent a pixel jump on focus.

src/elements/Link/Link.tsx
yuki24: This component is based on the `a` tag so it defaults to `cursor: pointer;` and javascript functions don't affect CSS, so this should be good.

src/elements/Link/Link.tsx
yuki24: > we might be forced to also add a transparent border for all other states to prevent a pixel jump on focus.

I'm not sure if that's a reasonable change since I don't think we want the border in the first place in almost any case.

oxaudo:
K. Merging this.

src/elements/Link/Link.tsx
damassi: @yuki24 - is it possible to add the border without the visual shift?

PR #172, opened 2019-01-12
Add another bucket of icons

damassi:
Adds another bucket of icons from Reaction and refactors the Chevron icon so that spacing is properly distributed when rotated in multiple directions.


artsyit:

Built with commit abd6142d977df9934cbf6d88913490897d45336d

https://deploy-preview-172--artsy-palette.netlify.com

src/svgs/ChevronIcon.tsx
```diff
@@ -1,42 +1,74 @@
 import React from "react"
-import styled from "styled-components"
+import { Flex } from "../elements"
 import { color } from "../helpers"
 
-export enum Direction {
+type Direction = "left" | "right" | "up" | "down"
+
+// TODO: This is for backwards compat with Volt; need to update there
+export enum Rotation {
```
damassi: @jonallured - I updated this component to make it simpler (now just pass in up / down / left / right rather than having to remember degrees). Would you mind checking out Volt and seeing if things are cool? I tried to keep things backwards compatible.

jonallured: Whoa, nice one! I might need a little direction on how to check Volt - I guess I'd need to check out a branch, upgrade Palette to this point, start a server and then bring up a page with the Chevron and ensure that it's behaving as expected - is that right?

damassi: @jonallured - I have a PR open for a palette upgrade here: https://github.com/artsy/volt/pull/3397, and this seems backwards compatible once that's merged, just not sure where to find that little chevron page-wise.

jonallured: Threw my feedback over here:

Hope that's ok! ❤️

PR #181, opened 2019-01-13
Move grid provider from reaction

damassi:
Removes the grid theme provider from Reaction since the grid now exists in palette


zephraph:
I think this was technically a `Version: Minor` since it added a new feature (from the perspective of an arbitrary consumer of palette). Not that it's too important.

damassi:
Part of weekend work that should have been grouped with other updates

PR #191, opened 2019-01-16
[Elements] Move Tabs and Toggle from Reaction

damassi:
Moves the `<Tabs>` and `<Toggle>` components over from Reaction. 

Cool that now that most code is built using palette things can be copied over essentially verbatim without changes, including straight from storybooks and from tests. 

(Moving components as they're needed in the new docs designs.)


artsyit:

Built with commit 253ff8bf2493869af796b54f4f1e621ea5fdf584

https://deploy-preview-191--artsy-palette.netlify.com

PR #201, opened 2019-01-17
[blocked] Migrate TimeRemaining component from Reaction

dblandin:
Migrate the `<TimeRemaining>` component from Reaction.

**Note**: Since this PR introduces a new dependency (luxon / https://moment.github.io/luxon/), it's likely blocked until we resolve an RFC to discuss its adoption by the team, following [these guidelines](https://github.com/artsy/README/blob/master/playbooks/dependencies.md#a-rubric-on-whether-a-dependency-needs-an-rfc) first proposed in [this RFC](https://github.com/artsy/README/issues/117).

2019-01-17-161413_1354x1209_scrot


dblandin:
I'll roll the dependency change into a separate "Switch from Moment.js to Luxon" PR once this batch of Reaction components are migrated to Palette and the new dependency goes through the RFC process.

PR #205, opened 2019-01-17
Migrate StaticCountdownTimer from Reaction

dblandin:
Migrate `<StaticCountdownTimer>` component from Reaction.

**Note**: This ports over a dependency from Reaction as well: `moment-timezone`. We eventually want to switch to [Luxon](https://moment.github.io/luxon/) but that'll first have to go through the [new dependency RFC process](https://github.com/artsy/README/blob/master/playbooks/dependencies.md#a-rubric-on-whether-a-dependency-needs-an-rfc).

2019-01-17-175440_1386x1731_scrot


src/elements/StaticCountdownTimer/StaticCountdownTimer.tsx
```diff
@@ -0,0 +1,63 @@
+import moment from "moment-timezone"
```
damassi: This is such a heavy dependency :(

damassi: Would you mind adding a FIXME here?

artsyit:

Built with commit 326fa8c922335e893a29326147e8c4116488fb5e

https://deploy-preview-205--artsy-palette.netlify.com

damassi:
👍

PR #218, opened 2019-01-23
Update OpenEyeIcon

oxaudo:
cc @mdole, @jeffreytr 

Also added the display of it to docs:
screen shot 2019-01-23 at 3 37 57 pm


artsyit:

Built with commit a38f98f46371f9e7065fd2d7a8888d34d0f537a6

https://deploy-preview-218--artsy-palette.netlify.com

artsyit:

Built with commit 5aadd5b74106838de425718df0d743e2f8946b6c

https://deploy-preview-218--artsy-palette.netlify.com

src/svgs/OpenEyeIcon.tsx
```diff
@@ -1,28 +1,44 @@
...
+      <title>Action/View/Md</title>
+      <g
+        id="Action/View/Md"
+        stroke="none"
+        stroke-width="1"
+        fill="none"
+        fill-rule="evenodd"
+      >
+        <g
+          id="eye"
```
zephraph: Maybe remove these ids?

oxaudo: They are coming from the original svgs and we are having those in all of our Icons here. I agree - we should maybe clean it up, but should probably do it in all the files here.

damassi: Even if we don't clean everything up at once its best to think of this stuff as incremental, so removing now would require less effort in the future.

damassi:
👍

PR #219, opened 2019-01-23
Move content out of `src` and into `/content`

damassi:
Previously we had content mixed with our app source code; this moves it into a top-level `/content` folder.


artsyit:

Built with commit f73a137b86aeaea51012bb564d9447c4a9d9004f

https://deploy-preview-219--artsy-palette.netlify.com

PR #228, opened 2019-01-24
Add CHANGELOG page

damassi:
Reads the changelog in from `auto` and displays it in the page. 

Could probably do this in a more efficient way, but `gatsby-source-filesystem` reads directories, not files, and we don't want to bring the whole of Palette into memory just for a single file. 

This could also use its own layout, the auto-generated headers are a bit too big within our theme: 

screen shot 2019-01-23 at 10 45 06 pm

And what happens when this file becomes huge? 

cc @zephraph


artsyit:

Built with commit ebc77d30a39e53873b34106728dcdd4207cd486c

https://deploy-preview-228--artsy-palette.netlify.com

PR #232, opened 2019-01-25
Update filter styles

mdole:
Updates filter styles according to spec from [DISCO-666](https://artsyproduct.atlassian.net/browse/DISCO-666). These changes impact the Collect page, Collection page, and Artist pages—I think that's everything, and there's a sister PR in Reaction to handle some spacing weirdness on specific components. Link: [Reaction PR](https://github.com/artsy/reaction/pull/1948)


docs/LabeledRange.mdx
```diff
@@ -30,8 +30,8 @@ All `LabeledRange` props are passed to the underlying [Slider](#/docs-slider) co
     defaultValue={[1, 100]}
     disabled={false}
     label="Height"
-    formatter={(min, max, maxIndicator) => {
-      return "" + min + " - " + max + maxIndicator + " in" 
+    formatter={(min, max, unit, maxIndicator) => {
```
dblandin: I'm wondering if it might be clearer to either use the `unit` or the `formatter` prop, but not both.

The `unit` prop can be used as a convenient way of specifying a unit with standard formatting. If you're passing in a `formatter` prop, I think that function can handle the unit displayed as well.

src/elements/LabeledRange/LabeledRange.tsx
```diff
@@ -20,11 +22,13 @@ export class LabeledRange extends React.Component<
 > {
   static defaultProps = {
     disabled: false,
+    unit: "",
   }
 
   state = {
     min: this.props.defaultValue[0],
     max: this.props.defaultValue[1],
+    unit: this.props.unit,
```
dblandin: The properties for a component are defined when the component is created. Sometimes the properties pass down into the initial state for the component. A component's state tracks attributes that change over time.

For the `unit` properly, I don't think it's likely to change so shouldn't need to track it as state. You can refer to it within the component via `this.props.unit`.

docs/LabeledRange.mdx
```diff
@@ -30,8 +30,8 @@ All `LabeledRange` props are passed to the underlying [Slider](#/docs-slider) co
...
+      return "" + min + " - " + max + maxIndicator + " " + unit
```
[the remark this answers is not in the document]
mdole: Oh my bad! I thought I had. Updated! now has the unit: [screenshot](https://cl.ly/ad440a67be57)

src/elements/PriceRange/PriceRange.tsx
```diff
@@ -29,7 +29,7 @@ export class PriceRange extends React.Component<PriceRangeProps> {
     return (
       <LabeledRange
         formatter={this.formatter.bind(this)}
-        label="Price"
+        label={this.props.currency}
```
dblandin: I haven't seen this change in any design comps. Are we changing the label for this component from "Price" to "USD" or "GBP"?

mdole:
thanks for reviewing! will make changes changes and re-commit shortly

src/elements/PriceRange/PriceRange.tsx
mdole: Yeah! It's part of the change in which price becomes a drop-down (it wasn't before). Shows up in this comp: [Zeplin](https://zpl.io/aByJzZe)

src/elements/LabeledRange/LabeledRange.tsx
mdole: Makes sense! Still wrapping my head around props + state :)

docs/LabeledRange.mdx
mdole: [Talked IRL - definitely makes sense, adding a fresh commit to implement]

src/elements/Toggle/Toggle.tsx
```diff
@@ -49,13 +49,13 @@ export class Toggle extends React.Component<ToggleProps> {
...
         <Separator mb={2} />
-        <Header mt="-6px" onClick={this.toggleExpand} disabled={disabled}>
+        <Header mb={2} onClick={this.toggleExpand} disabled={disabled}>
           <Flex justifyContent="space-between" alignItems="center">
             <Sans
               size={this.props.textSize as SansSize}
               weight="medium"
               color="black100"
-              mt={0.3}
+              lineHeight="1.83"
```
damassi: Since line-heights are typically defined [according to the font](https://github.com/artsy/palette/blob/master/src/Theme.tsx#L160) it's better to use margins and padding and stick with the defaults from the theme.

mdole: ah that makes sense. will tweak!

mdole:
@damassi so I rolled up most of my commits, but I'm unsure what to do with [these](https://cl.ly/91bc9f6201d9) since I think they're already merged into master. Any pointers?

mdole:
Def see what you mean—`git reflog` is also very helpful to know about. So since it's not a good idea to mess with commits from other users, is the best way to resolve this just to merge and go from there? And happy to wait till tomorrow to resolve and/or pick @dblandin's brain

damassi:
Yup, totally cool 👍

PR #235, opened 2019-01-28
Add script to generate a tokens file

jonallured:
This PR adds a script to generate the basic tokens that make up our design system. This would allow any other context where one can parse JSON to rely on palette as the source of truth for these values without having to be able to read TypeScript. An extremely naive example:

```ruby
token_data = open("https://raw.githubusercontent.com/artsy/palette/master/tokens.json")
tokens = JSON.parse(token_data)
puts tokens["colors"]["purple100"] # => "#6E1EFF"
```

For good measure we also include a test that will fail if/when the values in the TypeScript file drift from the generated JSON file.

Note: in order to do this we added `ts-node` so that doing this wasn't too hard.


artsyit:

Built with commit 2553e0b562fda60e57b0c768c74ff76269c43df0

https://deploy-preview-235--artsy-palette.netlify.com

damassi:
Oooh, this is great -- thanks @jonallured 👍

zephraph:
We could potentially looking into a way to automatically run the generate command via something like `lint-staged` in the future.

artsyit:
:rocket: PR was released in v2.36.0 :rocket:

scripts/generate-tokens.test.ts
```diff
@@ -0,0 +1,9 @@
+import { themeProps } from "../src/Theme"
+
+import tokens from "../tokens.json"
+
+// if this test fails, get them back in sync like this:
+// $ yarn generate-tokens
+it("tokens file does not drift from the theme props", () => {
+  expect(tokens).toEqual(themeProps)
+})
```
orta: nice move

PR #246, opened 2019-02-01
Add netlifycms categories

damassi:
Updates our CMS collections: 

screen shot 2019-01-31 at 4 11 59 pm

Unfortunately we can't do folder nesting in the collection sidebar but this should be sufficient.


artsyit:
:rocket: PR was released in v2.37.2 :rocket:

PR #247, opened 2019-02-01
Add react-helmet

damassi:
Removes react-head and adds react-helmet (since the newer version seems to be ideal API-wise)

Palette should deploy after this.


artsyit:
:rocket: PR was released in v2.37.3 :rocket:

PR #254, opened 2019-02-02
Fixes broken references to neflity cms paths on github

just-be-dev:
The references from netlify-cms to githubs content was broken do to it referencing content files from the root of the repository instead of using the `www` base root configured in gatsby. This change will leave the content path as is for development, but update the path for live deployments.


artsyit:
:rocket: PR was released in v2.37.8 :rocket:

PR #284, opened 2019-02-12
updates README with correct setup info; deprecates www/README

mdole:
@damassi if you want to address the last couple things in the www/README so we can delete it entirely, that would be great! Moved most things into the master README, but feel free to tweak where appropriate.


artsyit:

Built with commit 5980ff512feea943d6c933d561b8479635121b11

https://deploy-preview-284--artsy-palette.netlify.com

README.md
```diff
@@ -11,6 +11,33 @@ Artsy's Design System
 Palette is a collection of primitive, product-agnostic elements that help encapsulate Artsy's look and feel at base level. This project is intended to be used
 across our digital product portfoilo.
 
+Palette is built on top of [Gatsby](https://www.gatsbyjs.org/).
```
damassi: Since Palette and its docs are separate apps with different requirements we should keep the readme in www and make sure that things work well (add the cd www && yarn install step inside of `yarn docs` script in package.json`)

www/README.md
```diff
@@ -1,52 +1,7 @@
...
-
-open http://localhost:8000
-open http://localhost:8000/___graphql
-open http://localhost:8000/admin
-```
-
-### Deployment
-
-Merges to master are automatically deployed to https://palette.artsy.net.
-
```
damassi: This info is all relevant and should stay

damassi:
@mdole - would you mind closing this and opening a second pr with a `yarn install` added to [this script](https://github.com/artsy/palette/blob/master/package.json#L18). We should keep the documentation as it is and improve on in rather than consolidate.

mdole:
sure, makes sense!

PR #289, opened 2019-02-14
Fix yarn linking with docs

damassi:
Fixes `yarn link` when working in docs by enforcing styled-components resolution to `www/node_modules/styled-components`. 

Also add babel-plugin-styled-components so we can see styled components' displayNames when looking at the dom.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,19 +6,21 @@
       "@babel/env",
       {
         "targets": {
-          "browsers": [
-            "last 2 versions"
-          ]
+          "browsers": ["last 2 versions"]
         }
       }
     ]
   ],
   "plugins": [
     "babel-plugin-add-react-displayname",
+    "babel-plugin-styled-components",
     "@babel/plugin-proposal-class-properties",
-    ["module-resolver", {
-      "extensions": [".ts", ".tsx"],
-      "root": ["./src"]
-    }]
+    [
+      "module-resolver",
+      {
+        "extensions": [".ts", ".tsx"],
+        "root": ["./src"]
+      }
+    ]
   ]
 }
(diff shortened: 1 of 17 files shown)


.babelrc
```diff
@@ -6,19 +6,21 @@
...
-            "last 2 versions"
-          ]
+          "browsers": ["last 2 versions"]
         }
       }
     ]
   ],
   "plugins": [
     "babel-plugin-add-react-displayname",
+    "babel-plugin-styled-components",
```
damassi: See https://www.styled-components.com/docs/tooling#babel-plugin

www/gatsby-node.js
```diff
@@ -107,6 +107,9 @@ exports.onCreateWebpackConfig = ({ actions }) => {
     ],
     resolve: {
       modules: [path.resolve(__dirname, "src"), "node_modules"],
+      alias: {
+        "styled-components": require.resolve("styled-components"),
```
damassi: @mdole - that warning you saw in the console regarding multiple instances of styled components was resolved here. What this says, basically, is when encountering 
```
import styled from 'styled-components'
```
force resolution to the closest node_modules folder, which in this case is `www/node_modules`.

This is weird webpack trickery...

artsyit:

Built with commit 436651d6e0f3ce300223acb73c362a1248fe8a54

https://deploy-preview-289--artsy-palette.netlify.com

www/gatsby-node.js
mdole: whoa that's dope. nice solution

mdole:
checked out your branch and gave it a whirl, LGTM!

artsyit:
:rocket: PR was released in v3.0.3 :rocket:

jonallured:
Can confirm, this is working great, thanks for the quick fix!! 🏆 ❤️ 💯

PR #290, opened 2019-02-14
Add `label` prop to Radio

ds300:
This is a backwards compatible change to improve the API of the `Radio` element

## Why?

Before this PR, anything you pass to the `children` of a `Radio` will be dumped inside a `<label>`. This is bad semantically if you start putting other form elements in there, because the first of those inputs becomes the *labelled input*.  Obviously the radio button should be the labelled input, not any text boxes or other controls you put inside it.

This also causes weird interactions in mobile safari where tapping anything inside the label focuses on the first input inside the label, meaning you can't focus on the second input if there is one (which there is in one of our flows).

## How?

I added another prop which can be used to explicitly set the `label` content. If you pass this prop *and* children, then the children will be rendered *outside* of the `label`. To preserve backwards-compatibility, if you supply only children they will be rendered inside the `label`.

I would suggest adding a warning for supplying `children` but no `label` to prevent people having these issues in the future and to force us to migrate existing usages or `Radio` to this new style. Not sure if we have a pattern for that kind of thing or whether we'd prefer to migrate first and then add an error later?

### DIFF
--- a/src/elements/Radio/Radio.tsx
+++ b/src/elements/Radio/Radio.tsx
@@ -78,17 +78,17 @@ export const Radio: React.SFC<RadioProps> = props => {
       >
         <InnerCircle />
       </RadioButton>
-      <HiddenInput
-        type="radio"
-        name={name}
-        checked={selected}
-        disabled={disabled}
-        onChange={() =>
-          !disabled && onSelect && onSelect({ selected: !selected, value })
-        }
-      />
       <Flex flexDirection="column">
-        <Label htmlFor={name} disabled={disabled}>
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 773f693586e58bd1abc8b13f32f63da907ccdb7d

https://deploy-preview-290--artsy-palette.netlify.com

orta:
You can skip the peril red

damassi:
👍

artsyit:
:rocket: PR was released in v3.0.4 :rocket:

PR #292, opened 2019-02-15
[wip] fix nested collapse edge case

ds300:
I encountered this weird behaviour with nested Collapses while working on some MO stuff. Haven't managed to find a workaround, so this is how it behaves in production right now 😢 

Tried various ways to use `position="relative"` around the collapses, but no luck. Seems like the outer collapse is ignoring the height of the inner collapse, but can't imagine why.

bad-collapse

### DIFF
--- a/www/content/docs/elements/layout/Collapse.mdx
+++ b/www/content/docs/elements/layout/Collapse.mdx
@@ -54,3 +54,39 @@ See https://react-spring.surge.sh/gotchas#animating-auto for more info.
     )}
   </Toggler>
 </Playground>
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 3b66ea78110bc2b06554474d8227c31d60ad2eab

https://deploy-preview-292--artsy-palette.netlify.com

damassi:
@ds300 - is this still relevant?

ds300:
Yes but we can close this. I have a ticket this sprint to fix it.

PR #295, opened 2019-02-18
Stop palette from automatically including all-webfonts.css

mdole:
@damassi noticed that Force was making two calls to get all-webfonts.css as a result of the file being referenced both there and in Palette. To keep Palette flexible and prevent duplicate calls, we removed the automatic import from Palette's global styles and instead made it a call in our Gatsby site.

### DIFF
--- a/www/content/docs/home/getting-started.mdx
+++ b/www/content/docs/home/getting-started.mdx
@@ -2,6 +2,7 @@
 name: Getting started
 hideInNav: false
 ---
+
 ### Adding Palette
 
 To bring Palette into your app, install it as a package:
@@ -22,7 +23,7 @@ const { GlobalStyles } = injectGlobalStyles(`
 
 export const App = props => {
   return (
-    <Theme>      
+    <Theme>
       <>
         <GlobalStyles />
         ...
@@ -55,4 +56,25 @@ const Artworks = props => {
     </Flex>
   )
 }
+```
+
+In order to use Artsy's fonts, you'll also want to add our webfont file to the
+CSS in the header of your app:
+
+```
+// Example using Styled Components:
+import { createGlobalStyle, css } from "styled-components"
+
+const GlobalStyles = createGlobalStyle<P>`
+  @import url("https://webfonts.artsy.net/all-webfonts.css");
+`
+
+// React Helmet example:
+<Helmet>
+  <link
+    href="https://webfonts.artsy.net/all-webfonts.css"
+    rel="stylesheet"
+    type="text/css"
+  />
+</Helmet>
 ```
--- a/www/content/docs/home/getting-started.mdx
+++ b/www/content/docs/home/getting-started.mdx
@@ -59,17 +59,29 @@ const Artworks = props => {
 ```
 
 In order to use Artsy's fonts, you'll also want to add our webfont file to the
-CSS in the header of your app:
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 10627175e9396d53dd85f40a76cf7976c00c8f77

https://deploy-preview-295--artsy-palette.netlify.com

www/content/docs/home/getting-started.mdx
```diff
@@ -56,3 +57,24 @@ const Artworks = props => {
...
+`
+
+// React Helmet example:
+<Helmet>
+  <link
+    href="https://webfonts.artsy.net/all-webfonts.css"
+    rel="stylesheet"
+    type="text/css"
+  />
+</Helmet>
```
damassi: I like the two different example approaches! both very modern. Note however that we use [react-head](https://github.com/tizmagik/react-head) in Reaction so might want to also include that, and put it above the React Head example.

www/content/docs/home/getting-started.mdx
```diff
@@ -56,3 +57,24 @@ const Artworks = props => {
   )
 }
 ```
+
+In order to use Artsy's fonts, you'll also want to add our webfont file to the
+CSS in the header of your app:
+
+```
+// Example using Styled Components:
+import { createGlobalStyle, css } from "styled-components"
```
damassi: Can remove `css` since that's not being used

www/content/docs/home/getting-started.mdx
```diff
@@ -56,3 +57,24 @@ const Artworks = props => {
...
+
+const GlobalStyles = createGlobalStyle<P>`
```
damassi: Can remove the `<P>` since that's TypeScript specific and unneeded here.

www/content/docs/home/getting-started.mdx
```diff
@@ -56,3 +57,24 @@ const Artworks = props => {
...
+  @import url("https://webfonts.artsy.net/all-webfonts.css");
+`
+
+// React Helmet example:
```
damassi: Each example section can be broken out across the markdown so that more details / links can be added, e.g., 

```
### [Styled Components](https://www.styled-components.com/docs/api#createglobalstyle) example:
... add a code block ... 

### [React Head](..link to react head docs) example: 
... add a code block ...
```
and so on.

mdole:
Good thoughts! Updated—I'm not 100% confident the react-head example is correct because I haven't worked with it extensively, but I _think_ it's right. Gotta run for now but can double check tomorrow

www/content/docs/home/getting-started.mdx
```diff
@@ -56,3 +57,46 @@ const Artworks = props => {
...
+
+### Using [Styled Components](https://www.styled-components.com/docs/api#createglobalstyle)
+
+```
+import { createGlobalStyle } from "styled-components"
+
+const GlobalStyles = createGlobalStyle`
+  @import url("https://webfonts.artsy.net/all-webfonts.css");
+`
+```
```
damassi: I would toggle the position of the `react-head` example with the styled components example since the react-head and react helmet approaches are similar.

artsyit:
:rocket: PR was released in v3.0.5 :rocket:

PR #305, opened 2019-03-01
[Flex] Make `display` prop responsive

damassi:
Noticed that the `<Flex>` component wasn't responsive. So now it defaults to flex, but then the user can override with responsive props, which is good for layouts: 

```jsx
<Flex display={['none', 'flex']}>...</Flex>
```

### DIFF
--- a/src/elements/Flex/Flex.tsx
+++ b/src/elements/Flex/Flex.tsx
@@ -7,6 +7,8 @@ import {
   AlignItemsProps,
   bottom,
   BottomProps,
+  display,
+  DisplayProps,
   flexBasis,
   FlexBasisProps,
   flexDirection,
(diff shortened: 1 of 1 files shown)


src/elements/Flex/Flex.tsx
```diff
@@ -7,6 +7,8 @@ import {
   AlignItemsProps,
   bottom,
   BottomProps,
+  display,
+  DisplayProps,
```
damassi: This is the only addition; everything else is just alpha.

artsyit:

Built with commit d6af81ff3e568f33bf77d0477bef0a003d11bd47

https://deploy-preview-305--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v3.0.12 :rocket:

PR #308, opened 2019-03-01
New icons

### LINKED ISSUE #283: [NEW] Tokens/ Icons
[icons 2.zip](https://github.com/artsy/palette/files/2862390/icons.2.zip)

## Component Name
Icons

### Zeplin link

zpl://screen?sid=https://zpl.io/V4Q43NM

### Product team

All

### Design lead

@jeffreytr 

### Additional info
 Adding a .zip folder of all the exported SVG icons here. Let me know if anything seems missing or incorrect.

You can refer to naming conventions in this table on Notion: https://www.notion.so/artsy/2520b95ca7ed4d918b5cce46e30be71b?v=083ea403b3034ff5a8f26bd74d4c8184

---

#### For Tokens

_*Color, type, spacing, icons, etc*_

- [x] Is it serving a purpose? (is it needed or just purely decorative?)

- [x] Can it be merged with any existing styles on the site?

- [x] Has it been pressure-tested in all instances?

- [x] Does it meet accessibility standards? (min type size, color contrast)

#### For Elements

_*Buttons, links, loaders, inputs, dropdowns, etc*_

- [ ] Is it using up-to-date tokens from above?

- [ ] Have all states been captured? (hover, selected, disabled, focused, normal, thinking, errors)

- [ ] Have all transitions/animations been defined?

- [ ] Is it useable? (Considering touch targets, screen sizes)

- [ ] Is it consistent with the rest of the visual system? (Corner radius, stroke weight, form, shadow, opacity, etc)

- [ ] Does it meet accessibility standards? (is it keyboard navigable, does it have required accessibility markup)

- [ ] Does it render and function properly in Artsy supported browsers?

#### For Components

_*Nav, modules, modals, cards, etc*_

- [ ] Is it using up-to-date elements **and** tokens from above?

- [ ] Does it communicate clearly? (Copy, writing style/tone)

- [ ] Is it flexible? (Internationalization, text wrapping)

- [ ] Is it logical from a UX perspective? Does it follow paradigms set on the rest of Artsy?

- [ ] Has responsive behavior at all breakpoints been defined?

- [ ] Have all transitions/animations been defined?

- [ ] Does it meet accessibility standards?

- [ ] Does it render and function properly in Artsy supported browsers?

#### Accessibility

- [ ] Do all images and multimedia have `alt` or `title` tags?

- [ ] Are semantic elements used appropriately (`nav`, `button`, etc)?

- [ ] Are new components keyboard-navigable?

- [ ] Are hover interactions available by other means?

- [ ] Are `aria-` attributes included where appropriate?

- [ ] Has a Chrome Devtools [accessibility audit](https://developers.google.com/web/tools/chrome-devtools/accessibility/reference#audits) been performed?

#### Compatibility

Has the new component been reviewed in [Browserstack](https://live.browserstack.com/dashboard):

- [ ] Desktop Chrome

- [ ] Desktop Edge

- [ ] Desktop Safari

- [ ] Desktop Firefox

- [ ] Android

- [ ] iOS

jonallured:
This PR adds our new Icons and it's been quite the journey - thanks for everyone's feedback!! ❤️ 

Here's some pictures:

Screen Shot 2019-03-11 at 3 16 01 PM
Screen Shot 2019-03-11 at 3 16 06 PM

That first one shows off the list of basic icons, but then I also broke out sections for the more meta ones - currently just `ChevronIcon` and `CreditCardIcon`.

## The Migration

So, this PR started with me writing a Ruby script that [uses `svgo` to minify](https://github.com/svg/svgo) the SVG that I got from @jeffreytr. I made a CSV of all the filenames and the title and component names and the scripted the conversion to Typescript. That Ruby script will live at `scripts/new_icons.rb` for now. It'll serve as a historical artifact if nothing else but could also possibly be used to automate adding new SVGs.

## Visual Tweaks

At a certain point I realized that one could not actually resize the icons because I didn't respect the `viewBox` property so I added that back in. I also ensured that the fill was working so that the icon colors could be properly customized.

## The Audit

Migrating all these files didn't really take all that long, but then I had to audit the list of icons that were in use by clients of Palette and figure out which ones mapped to existing component names and which were new. This lead to discovering that the `ChevronIcon` and `CreditCardIcon` had behavior worth supporting, so I broke out those things.

I've added aliases so that clients should be able to upgrade with a minimum of changes and then we can do follow up PRs that migrate the clients to the new names and remove the aliases.

Closes #283


new_icons/script.rb
```diff
@@ -0,0 +1,85 @@
+require 'csv'
+require 'nokogiri'
+require 'json'
+
+template = <<-TSX
```
damassi: What does the `<<-TSX` tag do?

jonallured: It's the Ruby syntax for a multiline string:

https://www.rubyguides.com/2018/11/ruby-heredoc/

src/new_svgs/AddCircleFillIcon.tsx
```diff
@@ -0,0 +1,14 @@
...
+export const AddCircleFillIcon: React.SFC<IconProps> = props => (
+  <Icon {...props}>
+    <title>unfollow</title>
+    <path
+      d="M9 1a8 8 0 1 0 0 16A8 8 0 0 0 9 1zm.551 7.449H13V9.55H9.551V13H8.45V9.551H5V8.45h3.449V5H9.55v3.449z"
+      fill="#000"
+      fillRule="nonzero"
+    />
+  </Icon>
+)
```
damassi: nice  :)

jonallured:
The build failures are because the Insta SVG is from their brand assets and @jeffreytr is going to redraw for me. Once I get that, this will be pretty much ready for review. I kept all the junk in here from the conversion pipeline in the `new_icons` folder - that's temporary. I'll remove it before merging.

damassi:
@jonallured - out of curiosity can you paste a screen shot of the new icon page into the description? I'm dying to see all of the icons at an even size..

jeffreytr:
Sent @jonallured an updated IG icon so hopefully this will be resolved early next week :)

jonallured:
Ok, I've got the new Insta Icon in here and it's all green except for the building of the docs site so I'll work with @damassi to get that cleared up and then we can talk about merging this one.

artsyit:

Built with commit 5e33ed0e46a4cbc6cb28051158bf2425eddbbb26

https://deploy-preview-308--artsy-palette.netlify.com

jonallured:
I'm torn about something here. Prior to this new set of icons we had an icon called `ChevronIcon`:

screen shot 2019-03-05 at 3 35 19 pm

And it isn't just a simple SVG - it's actually four because it has behavior. There's an enum that rotates to a given direction:

https://github.com/artsy/palette/blob/7070fd73b0374c5f6b0c8cf9fe088b74fff01a1e/src/svgs/ChevronIcon.tsx#L45-L48

But that's not what we've got here - the new icons include individual icons for each orientation:

screen shot 2019-03-05 at 3 40 35 pm

So I can leave this for clients to update, but I actually have to address this in Palette for components like Toggle:

https://github.com/artsy/palette/blob/4e8d46e21e360c65013a7af47e89ede6b763a9f7/src/elements/Toggle/Toggle.tsx#L63-L65

And one way I could do that would be something like:

```jsx
<Flex justifyContent="right">
  {expanded && <ArrowUpIcon right={-5} />
  {!expanded && <ArrowDownIcon right={-5} />
</Flex>
```

But maybe this is worth a meta icon? Maybe I should bring back the concept of a `ChevronIcon` that supports this rotation so that users of Palette don't keep having to build this themselves? Thoughts?

damassi:
I think a good middle-ground would be to update `<ChevronIcon>` with the new Icons 
```
const Left = <LeftIcon />
...
```
While keeping both the old icon and the new.

jeffreytr:
@jonallured I think I follow you here.. So what you're recommending is keeping the 4 unique directional icons and then also having a single chevron icon that could work for components?

I'll go with your recommendation here. We could probably easily duplicate `arrowright` and rename to `chevronicon` to use as the meta one

jonallured:
Ok, I moved this out of draft as I'm finally feeling like this is ready for real review and close to being able to be merged! 🎉 

Netlify is choking on something that I can't quite understand so if anyone has thoughts that would be great!

src/svgs/ArtsyLogoIcons.tsx
```diff
@@ -1,26 +0,0 @@
...
-  </Icon>
-)
-
-/** Icon */
-export const ArtsyLogoIcon = props => (
-  <Icon xmlns="http://www.w3.org/2000/svg" viewBox="0 0 129 44" {...props}>
-    <path d="M4 4h121v36H75v-6h-4v6H4V4zM0 44h129V0H0v44z" />
-    <path d="M26.26 14h-3.3L17.2 29h3.3l1.51-4.2h5.2L28.7 29H32l-5.74-15zm-3.35 8l1.7-4.55L26.28 22h-3.37zM55 14v3h5.28v12h3.42V17H69v-3zM108.02 14l-3.74 6.41-3.72-6.41h-3.62l5.8 9.47V29h3.1v-5.53l5.8-9.47zM90.63 21.1c-.96-.7-2.28-1.02-3.44-1.3l-.71-.18-.14-.04c-.95-.23-2.25-.56-2.25-1.7 0-1.2 1.41-1.52 2.24-1.52.95 0 2.13.28 2.42 1.63l.01.01h2.93v-.03c-.1-2.72-2.01-4.29-5.25-4.29-2.4 0-5.22 1.13-5.22 4.3 0 3.3 2.94 4 5.53 4.62l.7.17c1.32.35 1.99.96 1.99 1.82 0 1.86-2.44 1.96-2.93 1.96-1.71 0-2.64-.64-2.93-2.01v-.04H80.6v.05c.02 1.17.16 1.89.92 2.77 1.43 1.67 3.7 1.92 4.92 1.92 2.83 0 5.87-1.47 5.87-4.69a4.3 4.3 0 0 0-1.68-3.44M47.8 22.6a4.75 4.75 0 0 0 1.82-3.8c0-1.65-.84-3.18-2.2-3.99a6.36 6.36 0 0 0-3.47-.81H37.3v15h3.13v-7.41L46.23 29h3.8l-4.43-5.53c.8-.14 1.45-.32 2.2-.86M40.41 21v-4.02h3.52c.7 0 1.32.03 1.82.4.47.34.75.94.75 1.61 0 1.33-.95 2-2.83 2h-3.26v.01z" />
-  </Icon>
-)
```
damassi: Removing this file is a breaking change

src/svgs/BookIcon.tsx
```diff
@@ -1,16 +0,0 @@
...
-    <Icon width="25" height="25" xmlns="http://www.w3.org/2000/svg" {...props}>
-      <g fill="none" fillRule="evenodd">
-        <path d="M0 0h25v25H0z" />
-        <g stroke="#000">
-          <path d="M17.5 8.616v7.73L12.808 17.5H12.5V9.667l5-1.05zM7.5 8.616v7.73l4.692 1.154h.308V9.667l-5-1.05z" />
-        </g>
-      </g>
-    </Icon>
-  )
-}
```
damassi: This (could) be a breaking change.  Similar to comment above, to keep the fallout to a minium this PR should be purely additive and  a follow-up pr can cleanup.

src/svgs/CheckIcon.tsx
```diff
@@ -1,39 +1,16 @@
 import React from "react"
-import { color } from "../helpers"
-import { Color } from "../Theme"
 import { Icon, IconProps } from "./Icon"
 
-interface CheckIconProps extends IconProps {
-  fill?: Color
```
damassi: I'm not sure of the best way to go about this, but interfaces containing things like `style`, `fill` sand so on need to be extended globally to `Icon.tsx` or individual icon interfaces need to be preserved. If `fill` is present  than that  means that this icon has collor applied  to it somewhere in one of our apps.

src/svgs/CheckIcon.tsx
```diff
@@ -1,39 +1,16 @@
...
-  style?: object
-  onClick?: (e: any) => void
+/** CheckIcon */
+export const CheckIcon: React.SFC<IconProps> = props => {
+  return (
+    <Icon {...props} viewBox="0 0 18 18">
+      <title>check</title>
+      <path
+        d="M6.936 12.206l7.64-7.63.848.849-8.492 8.48-4.248-4.282.852-.846z"
+        fill="#000"
```
damassi: E.g., the green checks that we see around will now all be black because `fill` is no longer  present. 

if there's a way to hoist fill properties up to `Icon` would be ideal (every icon would then support color) but I'm not sure of the best path forward for this.

src/svgs/ChevronIcon.tsx
```diff
@@ -1,84 +1,44 @@
...
-          case "down":
-            return Down
-          default:
-            return Right
-        }
-      })()}
-    </Icon>
-  )
+}: ChevronProps) => {
+  const Arrow = rotationMap[direction] || directionMap[direction]
```
damassi: Nice cleanup 👍 

Can you confirm that the arrows perfectly rotate around a center axis point?

src/svgs/CloseIcon.tsx
```diff
@@ -1,26 +1,16 @@
 import React from "react"
-import { color } from "../helpers"
-import { Color } from "../Theme"
 import { Icon, IconProps } from "./Icon"
 
-interface CloseIconProps extends IconProps {
-  fill: Color
```
damassi: ^

src/svgs/CreditCardIcon.tsx
```diff
@@ -1,150 +1,16 @@
...
-      return <VisaIcon {...others} />
-    case "Discover":
-      return <DiscoverIcon {...others} />
-    case "American Express":
-      return <AmexIcon {...others} />
-    case "Unknown":
-    default:
-      return <FallbackIcon {...others} />
-  }
-}
```
damassi: Why remove this code?

src/svgs/CreditCardIcon.tsx
```diff
@@ -1,150 +1,16 @@
...
+        fill="#000"
+        fillRule="nonzero"
       />
-      <path
-        fill="#4287CB"
-        d="M1.992 0C.892 0 0 .898 0 1.998V6h30V1.998A1.999 1.999 0 0 0 28.008 0H1.992z"
-      />
-    </g>
-  </Icon>
-)
```
damassi: Why remove all of this code? This has specific relevance to the Purchase team order flow.

src/svgs/EyeOpenedIcon.tsx
```diff
@@ -1,35 +1,15 @@
 import React from "react"
 import { Icon, IconProps } from "./Icon"
 
-interface OpenEyeIconProps extends IconProps {
```
damassi: This file name should be preserved, and if a rename is necessary it can exist as a duplicate alongside the deprecated old file until a follow-up cleanup PR, at which point it can be fully deprecated / removed.

src/svgs/GroupIcon.tsx
```diff
@@ -1,18 +0,0 @@
-import React from "react"
```
damassi: Breaking change

damassi:
This PR is looking good, but there are a whole lot of breaking changes which we'll need to address. This is non-exhaustive, but some that stand out:

1. Icon-specific properties (such as `fill` and other behavior) are no longer preserved. We apply color and other behaviors to icons throughout Reaction and force (and possibly elsewhere).
1. Files are deleted
1. Files are renamed
1. Exports from files are removed, which will lead to import undefined errors.

For something this sensitive we should be preserving old behavior to the greatest extent that we can, ensuring that the transition is mostly smooth, and then move over a check-list that works through all of the *necessary* and justifiable breaking changes that have been made. Once that list is in place we can a) update client libs; and b) notify specific teams to be on the lookout and / or assist in the refactor.

damassi:
@jonallured - would you mind updating the description with an overview of what's included + a screenshot of the docs?

src/svgs/Icon.tsx
```diff
@@ -35,3 +38,7 @@ export const Icon = styled.svg.attrs<IconProps>({})`
   ${left};
   ${position};
 `
+
+Icon.defaultProps = {
+  fill: "black100"
```
damassi: 👍

jonallured:
Ok so with ebf640e I added a `fill` prop and then went through the icons and updated them to support this at whichever level of the svg hierarchy was required. I tested locally that I was able to set the color of all icons to purple100 and it worked great! Except for the white logo and mark, but I think that's ok - those are sorta special.

Then, with a2189cb I took a stab at backwards compatibility - what if I did something like this for all the ones being replaced @damassi? Would that work without clients having to do anything?

src/svgs/CreditCardIcon.tsx
jonallured: I did not notice that there was behavior in here - what I ended up doing was something like this:

```
$ rm -rf src/svgs
$ mv src/new_svgs src/svgs
```

And then just committed all the churn. So what I'm trying to say is that I didn't go file by file and delete things, I did it all in a batch and assumed we'd be good - clearly I was wrong! I'll follow up on some of these and figure out a strategy to not break things.

damassi: Ahh i see -- I had assumed that once the folder was outputted to `new_icons` things were dragged over manually and then cleaned up by hand. This makes sense now.

damassi:
> Would that work without clients having to do anything?

As long as filenames and export names are preserved I think we'll be good. (Make sure to add either a `FIXME` around each alias or create a running list as they're are added though.)

jonallured:
Update: with a8548ee I've added all the aliases so that existing icons will continue to be supported as clients upgrade and then we can remove them once migration to the new names has happened.

So I think I just have two things left to deal with:

* The `CreditCardIcon` actually has individual icons for MC, Visa, Discover and AMEX
* The `WeChatIcon` has some behavior around showing a QR code on hover

damassi:
Nice :) I think `CreditCardIcon` and `WeChatIcon` were just restored as they were we'd be good to go. Is that what yr thinking? 

(Let me know when I should do another review pass.)

src/svgs/CheckIcon.tsx
jonallured: This works now:

```
<CheckCircleIcon fill="green100" />
```

Gives you this:

Screen Shot 2019-03-11 at 3 27 47 PM

src/svgs/ChevronIcon.tsx
jonallured: I'm not 100% sure how to determine this, @jeffreytr do you have a good way of answering this?

jonallured:
@damassi ok I have addressed all feedback and feel like this is ready for another review! Note that I did updated the PR description with images and more details.

src/svgs/index.tsx
```diff
@@ -1,43 +1,69 @@
...
 export * from "./FairIcon"
 export * from "./FilterIcon"
 export * from "./GenomeIcon"
-export * from "./GroupIcon"
+export * from "./GraphIcon"
+export * from "./HeartFillIcon"
 export * from "./HeartIcon"
-export * from "./HelpIcon"
-export * from "./Icon"
+export * from "./HomeIcon"
```
damassi: Can we add `// TODO's` around the icons that have been aliased so that there's a quick reference list?

www/src/components/Sidebar/index.tsx
```diff
@@ -17,7 +17,7 @@ export const Sidebar = _props => {
       </Box>
 
       <Box>
-        <ArtsyMarkIcon width="30px" height="30px" mr={2} />
+        <ArtsyMarkBlackIcon width="30px" height="30px" mr={2} />
```
damassi: Ah, this is what I was looking for a second ago 👍

damassi:
Alright here we go, let us 🙏

artsyit:
:rocket: PR was released in v4.0.0 :rocket:

jonallured:
Thanks for mentoring me through the process @damassi!! ❤️ 💯 🏆

Issue #117, opened 2018-10-23
Implement Sidebar Structure


owendodd:
As confirmed in #83 

---

**Home**
- About
- Change-log

---

**Tokens**
- Color
- Typography
- Spacing
- Grid
- Icons

---

**Elements**
- Images
  - Avatar
  - Image
  - ImageCarousel
  - LightBox
- Layout
  - BorderBox
  - Box
  - Flex
  - StackableBorderBox
  - Separator
  - Join
  - Overflow
  - ReadMore
- Buttons
  - Buttons
  - Text links
- Inputs
  - Input
  - QuickInput
  - Radio
  - GroupRadio
  - Checkbox
  - Dropdown
  - MiniDropdown
  - Slider
- Dialogs
  - Banner
  - Modal
  - Alert
  - Message
  - Confirmation
  - Tooltip
  - PopOver
- Navigation
  - Tabs
  - Pagination
- Animation
- Loaders
  - Spinner

---

**Components**
- Header
- Footer
- Bricks
  - ArtistBrick
  - GalleryBrick
  - ShowBrick
  - ArticleBrick
  - ArtworkBrick

zephraph:
This is more or less finished? I'm going to close it for house keeping purposes. @owendodd let me know if we're missing anything.

PR #309, opened 2019-03-04
Bar chart

ds300:
It's BarChart time 🎉 

This was done in collaboration with @lilyfromseattle 

Zeplin over here: https://zpl.io/2GyGr5E
Jira over here: https://artsyproduct.atlassian.net/browse/PURCHASE-893

histogram


artsyit:

Built with commit 3ba9fda69db5f956440f5cbad1803fab202b78ac

https://deploy-preview-309--artsy-palette.netlify.com

src/elements/BarChart/Bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+const BarHoverLabel = ({ children }: { children: React.ReactNode }) => {
+  const ref = useRef(null)
+  const { x, y } = useContext(MousePositionContext)
+  if (ref.current) {
+    // position outside of the render loop to avoid GC churn
+    ref.current.style.top = y - LABEL_OFFSET - ref.current.offsetHeight + "px"
+    ref.current.style.left = x + "px"
+  }
+  return <HoverLabelPositioner ref={ref}>{children}</HoverLabelPositioner>
+}
```
damassi: This is a nice approach 👍

src/elements/BarChart/Bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+  })
+  return (
+    <HighlightLabelPositioner ref={ref as any} style={{ opacity }}>
+      <HighlightLabelBox>{children}</HighlightLabelBox>
+      <LabelLine />
+    </HighlightLabelPositioner>
+  )
+}
+
+// tslint:disable-next-line completed-docs
```
damassi: Since this is a good amount of code it might be good to add a description of how it fits in. Don't need like `@props` dance and all that, but maybe a couple words for overall context.

src/elements/BarChart/Bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+  onMeasureHeight,
+}: {
+  heightPercent: number
+  label: React.ReactNode
+  highlightLabel?: React.ReactNode
+  hasEnteredViewport: boolean
+  onMeasureHeight?: (height: number) => void
+}) => {
+  const [hover, setHover] = useState(false)
+  const finalBarHeight =
```
damassi: Can we add a few comments to the following logic that describe what's going on?

src/elements/BarChart/Bar.tsx
```diff
@@ -0,0 +1,163 @@
...
+  return <HoverLabelPositioner ref={ref}>{children}</HoverLabelPositioner>
+}
+
+const HighlightLabel = ({
+  children,
+  onMeasureHighlightLabel,
+  opacity,
+}: {
+  children: React.ReactNode
+  onMeasureHighlightLabel: (height: number) => void
```
damassi: Comment pls

src/elements/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+      }
+    }
+    window.addEventListener("scroll", handleScroll)
+    window.dispatchEvent(new Event("scroll"))
+    return () => {
+      window.removeEventListener("scroll", handleScroll)
+    }
+  })
+  return hasEntered
+}
```
damassi: Not quite ready to start moving these into a more general place (since it's not being used anywhere else) but def something to think about in the future.

src/elements/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+          height="80px"
+          alignItems="flex-end"
+          mb={0.5}
+          style={{ minHeight }}
+        >
+          {bars.map(({ value, label, highlightLabel }, index) => {
+            return (
+              <Bar
+                key={index}
+                heightPercent={(100 / maxValue) * value}
```
damassi: maybe `const heightPercent = (100 / maxValue) * value`?

src/elements/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,91 @@
...
+                hasEnteredViewport={hasEnteredViewport}
+                onMeasureHeight={
+                  highlightLabel ? height => setMinHeight(height) : null
+                }
+              />
+            )
+          })}
+        </ChartContainer>
+        <Flex justifyContent="space-between">
+          <Sans color={"black60"} size="2">
```
damassi: #minor - no need for `{}`

www/content/docs/components/BarChart.mdx
```diff
@@ -0,0 +1,134 @@
...
+        Price
+      </Serif>
+      <Sans size="3">Price ranges of small prints by Alex Katz</Sans>
+    </Flex>
+    <Flex>
+      <BarChart
+        bars={[
+          {
+            value: 2000,
+            label: <BarLabel title="Sept 30" description="423 views" />,
```
damassi: As a general API convenience, is it possible to supply a string to `label` and / or an object (`{ title: "Foo", description: "bar" }`) that maps to `<BarLabel>` under the hood? Seems like in most cases one would just need a title and a description passed into a default component, and in others to replace it, in which case it's nice that it accepts _also_ accepts a JSX node.

ds300: You can pass a string already, but it won't be styled. There were no designs featuring a single string, so I figured I'd leave it up to consumers to decide what their labels should look like.

I considered allowing an object instead of requiring JSX, but decided against it. It would be more complicated to consume since it increases the API surface area of the `label` and `highlightLabel` properties, and makes this JSX behaviour less discoverable in case people need to stray beyond the simple title/description format later on. It would also be more complicated to implement.

Happy to reconsider if you feel strongly about this?

www/content/docs/components/BarChart.mdx
```diff
@@ -0,0 +1,134 @@
+---
+name: BarChart
+---
+
+<Playground showEditor={true}>
```
sweir27: I'm getting a `ReferenceError: BarChart is not defined` when I open the deploy preview for this PR and navigate to the `Bar Chart` docs. Not sure if it's just the PR environment or an actual issue?

nicoleyeo:
@ds300 @lilyfromseattle This is looking great! One minor comment — the desired location path in Palette for this component is Elements/DataVisualization/BarChart (we want to make a new folder called DataVisualization), looks like it's currently placed at Components/BarChart. It would be great to have this organized properly in Elements to maintain organization/future find-ability (this was defined in the spec, in the title). Please let me know if theres a better way to note this in the future?  Thank you! 

cc @damassi please advise if there's a better way for me to communicate about component architecture.

www/content/docs/components/BarChart.mdx
ds300: This is because of the way the docs are set up. They currently load the latest version of palette published on npm, rather than the local version. I'm working on a fix for this (moving palette repo to yarn workspaces) but it'll happen after this PR.

sweir27: Ah! Okay cool. 👍

www/content/docs/components/BarChart.mdx
damassi: I'm down to iterate but do feel somewhat strongly and would like to bat this back and forth a bit... 

As far as discoverability goes having the option clearly in the docs is more than sufficient, and unless the _default_ is to provide ones own JSX to `label` (that isn't the default `BarLabel` component), I don't understand why it would be problematic to reference the `BarLabel` component internally? As a user of the chart, I don't want to have to import two separate components when the same outcome could be achieved by just importing  `BarChart`, and passing in a standard object (which could also be easily created dynamically, outside of a JSX context). Since there aren't other chart variations, passing JSX is less important.

damassi:
@ds300 - re folders, the NetlifyCMS [config collection](https://github.com/artsy/palette/blob/master/www/static/admin/config.js#L60) should  also be updated.

damassi:
@nicoleyeo - As long as the [Sidebar Structure](https://github.com/artsy/palette/issues/117) issue is updated as  things change I think we should be good!

www/content/docs/components/BarChart.mdx
ds300: >  I don't understand why it would be problematic to reference the BarLabel component internally?

It would mean we have to inspect the `label` object, check that it has the structure `{title, description}` and then turn it into a `<BarLabel>` if so. Not a huge deal, but definitely more complicated than just blindly passing on whatever the user supplies.

> As a user of the chart, I don't want to have to import two separate components when the same outcome could be achieved by just importing BarChart, and passing in a standard object (which could also be easily created dynamically, outside of a JSX context).

Personally speaking, importing components is not something that adds complexity or labour to my workflow. Similarly, managing JSX contexts is no trouble. I'm pretty sure that consumers of the `BarChart` component will be working in JSX contexts anyway.

I get why you want this from an aesthetic viewpoint. It's tidier not to have an extra component and not to mix up JSX in JS objects. I don't really mind either way, except it would take me 10-20 mins or so to do 😅

Give me a :+1: if I should go ahead.

damassi: If you could give it a quick shot that would good 👍We can use an interface to guard against unexpected data and for now assume that the user is either  a) passing in  a `BarLabel` component, or b) passing in an object with a `label` and `description` prop, and nothing else. (Probably shouldn't need to guard anywhere except at the interface level).

ty :)

ds300:
Ok @damassi this should be 🍏

src/elements/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const hasEnteredViewport = useHasEnteredViewport(ref)
+  const [minHeight, setMinHeight] = useState(0)
+  const maxValue = bars.reduce((max, { value }) => {
+    return value > max ? value : max
+  }, -Infinity)
+  return (
+    <ProvideMousePosition>
+      <Flex flexDirection="column" ref={ref} flexGrow="1">
+        <ChartContainer
+          height="80px"
```
sepans: Is the chart height tied to this value? I see the `MAX_BAR_HEIGHT` is also 80.

www/content/docs/elements/data-visualization/BarChart.mdx
```diff
@@ -0,0 +1,130 @@
...
+        Price
+      </Serif>
+      <Sans size="3">Price ranges of small prints by Alex Katz</Sans>
+    </Flex>
+    <Flex>
+      <BarChart
+        bars={[
+          {
+            value: 2000,
+            label: { title: "Sept 30", description: "423 clicks" },
```
damassi: 👍

damassi:
One q from @sepans we can loop back. Thanks  for making those changes, looks great 👍

artsyit:
:rocket: PR was released in v3.2.1 :rocket:

PR #310, opened 2019-03-04
Improve Link API

jonallured:
After some thought, I took a stab at what I think the API should be for a Link. The problems I'm trying to solve here are how one controls the underline and color properties together and separately. These concepts were intermingled in a way that @pepopowitz and I found difficult to work with - hopefully this is better!! ❤️ 

## Underline behavior

The first thing I did was extract an enum for the three types of underline behaviors I saw:

* underline the link all the time (default)
* never underline the link (none)
* underline the link on hover (hover)

And I added a prop called `underlineBehavior` to replace the `noUnderline` prop. If you were using that prop, it maps to the `hover` value of `underlineBehavior`.

## Color and hover color

Next I looked at how I could improve dealing with color. We want our links to be black by default and when a color is provided, we want a hover effect that transitions to black. That works great unless you want to run the link over something that is also black. Or my use case which was running the link over an image with an overlay. In those cases, we just always want the link to be white.

So, my solution here was to add a new prop called `hoverColor` that will override the default behavior of changing the color of the link on hover to black.

Then I updated the docs to reflect this and showed off the various ways these things can be used.

Would love any feedback on the approach!! 😍

### DIFF
--- a/src/elements/Link/Link.tsx
+++ b/src/elements/Link/Link.tsx
@@ -18,6 +18,7 @@ const computeUnderline = (state: string, behavior: UnderlineBehavior): string =>
 export interface LinkProps extends SpaceProps {
   underlineBehavior: UnderlineBehavior
   color?: Color
+  hoverColor?: Color
 }
 
 /**
@@ -30,10 +31,7 @@ export const Link = styled.a<LinkProps>`
   text-decoration: ${props => (computeUnderline("normal", props.underlineBehavior))};
   &:hover {
     text-decoration: ${props => (computeUnderline("hover", props.underlineBehavior))};
-    color: ${color("black100")};
-  }
-  &:focus {
-    color: ${props => (props.color ? color(props.color) : color("black100"))};
+    color: ${props => props.hoverColor ? color(props.hoverColor) : color("black100")};
   }
   ${space};
   ${styledColor};
--- a/src/elements/Link/Link.tsx
+++ b/src/elements/Link/Link.tsx
@@ -15,10 +15,16 @@ const computeUnderline = (state: string, behavior: UnderlineBehavior): string =>
   return none ? "none" : "underline"
 }
 
+const backwardsCompatCompute = (state: string, props: LinkProps) => {
+  const behavior = props.noUnderline ? UnderlineBehavior.Hover : props.underlineBehavior
+  return computeUnderline(state, behavior)
+}
+
 export interface LinkProps extends SpaceProps {
-  underlineBehavior: UnderlineBehavior
   color?: Color
   hoverColor?: Color
+  noUnderline?: boolean
+  underlineBehavior: UnderlineBehavior
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mzikherman:
I would make sure to keep this backwards compatible (ie- allow `noUnderline` and treat it as is currently), in order to minimize clients not being able to update w/o breaking changes. Then the work item to update clients and possibly drop backwards compatibility can be its own thing.

In semver you'd just make this a major version upgrade, but the way we've been using Palette (and Reaction), is more 'continuous deployment' and so when possible, I would avoid any breaking updates.

src/elements/Link/Link.tsx
```diff
@@ -3,9 +3,22 @@ import { color as styledColor, space, SpaceProps } from "styled-system"
...
+}
+
+const computeUnderline = (state: string, behavior: UnderlineBehavior): string => {
+  const blocklist: UnderlineBehavior[] = state === "hover" ? [UnderlineBehavior.None] : [UnderlineBehavior.Hover, UnderlineBehavior.None]
+  const none = blocklist.includes(behavior)
+  return none ? "none" : "underline"
+}
+
 export interface LinkProps extends SpaceProps {
-  noUnderline?: boolean
```
damassi: Like @mzikherman mentioned, it would be best to maintain this prop for now while mapping old behavior to the new API. Then once that's out can grep over Reaction to see what currently uses `noUnderline` and replace with the new API. Then once that's done, can come back to Palette and open a new PR with this deprecated prop removed and mark as a MAJOR release, then manually update Reaction with the new Palette version (since Major updates don't automerge).

src/elements/Link/Link.tsx
```diff
@@ -3,9 +3,22 @@ import { color as styledColor, space, SpaceProps } from "styled-system"
...
 
+enum UnderlineBehavior {
+  Default = "default",
+  Hover = "hover",
+  None = "none",
+}
+
+const computeUnderline = (state: string, behavior: UnderlineBehavior): string => {
+  const blocklist: UnderlineBehavior[] = state === "hover" ? [UnderlineBehavior.None] : [UnderlineBehavior.Hover, UnderlineBehavior.None]
+  const none = blocklist.includes(behavior)
```
oxaudo: Wondering if Case statement here would be more descriptive of what are we trying to express. It took me a second to realize what  this `blocklist ` is blocking.

oxaudo:
I think those changes are great for making the Link more flexible element.

There are couple of minor comments about implementation that I had. 

Also would be really cool for @design team to add comments here if that is how they think about links in our app. Wondering if they agree that we need ability to control underline/underline on hover/color/color on hover independently or they should come as a bundle and be regulated by the one link style.

Maybe this comp: https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5afc758c09a18d3d0f176edd can be updated with proper options/styles @nicoleyeo ?

jonallured:
Ok with 9f747ff I've got this thing backwards compatible now. My hope is that when we return to this we can just `git revert` that commit and be back in business.

For posterity, I did check in with @nicoleyeo and got the 👍 go ahead from her, so this behavior is approved.

Netlify is failing but I don't trust that thing so I think this is actually green and good to go? LMK if I'm wrong about that!! ❤️ ❤️ ❤️

oxaudo:
Looks like netlify might be failing because of 'Serif' use without specifying font size in example? Don't think we have any defaults there. No?

artsyit:

Built with commit fdbe95c431e45dac6b1a6c2b1586a466ccc32269

https://deploy-preview-310--artsy-palette.netlify.com

jonallured:
OMG, you were totally right, thanks @oxaudo!! 🏆

oxaudo:
I'm game to get this merged! Please do when feel ready :)

jonallured:
Cool, going to merge this and follow the rollout plan as outlined by @damassi above. Thanks for the feedback all!! ❤️

artsyit:
:rocket: PR was released in v3.2.0 :rocket:

PR #311, opened 2019-03-05
[Menu] Add <Menu> component

damassi:
Adds a new Menu component. Extracted from [Zeplin](https://zpl.io/aNPZPPZ). 

screen shot 2019-03-04 at 7 49 14 pm


artsyit:

Built with commit 54fd1b9c6c43d5651db41175e95288ef44d3c25b

https://deploy-preview-311--artsy-palette.netlify.com

src/elements/Menu/Menu.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { Sans } from "../Typography"
+
+interface MenuProps {
+  title?: string
+  children?: React.ReactNode
+}
+
+export const Menu: React.SFC<MenuProps> = ({ title, children, ...props }) => {
+  return (
+    <MenuContainer width={230} m="2px" {...props}>
```
damassi: `2px` is for the drop shadow

src/elements/Menu/Menu.tsx
```diff
@@ -0,0 +1,91 @@
...
+  color?: string // TODO:  Look into type conflict with styled-system
+}
+
+export const MenuItem: React.FC<MenuItemProps> = ({
+  children,
+  href,
+  ...props
+}) => {
+  return (
+    <MenuItemContainer onClick={() => window.location.assign(href)} {...props}>
```
damassi: Capture the outer container click while leaving an `href` link below for seo, screen-readers, etc.

www/static/admin/config.js
```diff
@@ -31,7 +31,7 @@ const defaultFields = {
     {
       label: "Status",
       name: "status",
-      widget: "boolean",
+      widget: "string",
```
damassi: Noticed that our `WIP` badge in palette docs was incorrect (had updated, but didn't switch from a boolean to a string, which allows us to pass different things).

src/elements/Menu/Menu.tsx
jonallured: I don't follow - why do we need a click handler here?

src/elements/Menu/Menu.tsx
```diff
@@ -0,0 +1,91 @@
...
+  ${Link} {
+    display: flex;
+    align-items: center;
+
+    &:hover {
+      text-decoration: none;
+    }
+  }
+
+  ${Sans} {
```
jonallured: 😎 cool!

damassi:
> netlify version doesn't properly render because Menu isn't found

Ah yes, @ds300 is working on this tomorrow.

src/elements/Menu/Menu.tsx
damassi: This keeps things a little visually flexible and easily clickable since there are some constraints around nesting block level content within an `a` tag. With a click handler on the outer container I can capture the entire area while still being able to style using common palette idioms.

pepopowitz: Have we tested this will work with SEO/screen-readers? I've always operated under the rule "if the user clicks on it and it takes them somewhere, use a link", and capturing the click on a div makes me slightly uncomfortable. I don't have any data or resources to point to that say "don't do it", but I'd want to be very sure that it was safe.

jonallured: Ohhh, this is about click target, right? Without that `onClick`, there would be space inside the `MenuItemContainer` where the user wouldn't be able to click. I guess that happens because of the padding on the `Box`, right? Any reason to not promote `Link` up and give it the padding? Are these the idioms you were referring to?

damassi: I don't have anything offhand, but I can update the container to be an `a` tag. This does seem a bit funny.

artsyit:
:rocket: PR was released in v3.1.0 :rocket:

PR #315, opened 2019-03-08
Make underlineBehavior simpler for consumers

damassi:
Noticed another issue when updating API in reaction - by using an `enum` it forces consumers to pass enums to define the behavior, e.g., `underlineBehavior={UnderlineBehavior.NONE}`, which is a bit verbose. 

```sh
ERROR in /Users/cn/Sites/artsy/reaction/src/Components/NavBar/NavItem.tsx
29:44 Type '"none"' is not assignable to type 'UnderlineBehavior'.
    27 |       onMouseLeave={() => toggleHover(false)}
    28 |     >
  > 29 |       <Link href={href} color={hoverColor} underlineBehavior="none">
```

While strings and enums can be compared within code, passing a string as an prop !== enum.string, unless it's coerced. Also fixes a possible regression where underline behavior wasn't being applied by default in our global styles. It's not apparent in Force because we already have a global set of styles, but in Storybooks I noticed body copy links weren't being underlined.

### DIFF
--- a/src/elements/Link/Link.tsx
+++ b/src/elements/Link/Link.tsx
@@ -3,35 +3,27 @@ import { color as styledColor, space, SpaceProps } from "styled-system"
 import { color } from "../../helpers"
 import { Color } from "../../Theme"
 
-enum UnderlineBehavior {
-  Default = "default",
-  Hover = "hover",
-  None = "none",
-}
+type UnderlineBehaviors = "default" | "hover" | "none"
 
 export interface LinkProps extends SpaceProps {
   color?: Color
   hoverColor?: Color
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 72b0fe2e2b144582d0a7f2a52cb84ecf69c511a9

https://deploy-preview-315--artsy-palette.netlify.com

jonallured:
Oh hrm. I thought this was a good use-case for enums, but looks like string is better here. Thanks!!

artsyit:
:rocket: PR was released in v3.2.4 :rocket:

PR #316, opened 2019-03-08
Use yarn workspaces

ds300:
## What are yarn workspaces?

Yarn workspaces is a feature that makes it easier to work on projects with more than one npm package. It `link`s the packages together automatically and 'hoists' dependencies that are shared between packages to a parent `node_modules` folder. This is useful for us because it simplifies local development of palette (by automatic linking) and prevents duplicate-dependency errors for packages like styled-components and react which have global state.

This also solves another issue which has been causing confusion lately: it makes the netlify PR previews represent the version of palette in the PR (rather than the version of palette published to npm).

## What does that mean for developers?

- When you run `yarn` or `yarn install` in the root folder it installs all the dependencies.

- When you run `yarn start` it runs gatsby in development mode while watching for changes in palette.

- There is only one `yarn.lock` file, in the project root.

- The .vscode config applies to the whole repo, so you need to open the root dir in vscode rather than opening a package dir if you want the vscode settings to apply. We could symlink the .vscode folder into the package dirs if people find this frustrating.

## How does this affect CI?

The CI setup remains almost exactly the same, the only difference is we use 'yarn release' in the root dir to publish palette to npm now (rather than using the [auto-release task](https://github.com/artsy/orbs/blob/ffdd758bf8a8ba19f45dcd7d089522222fe67df0/src/yarn/yarn.yml#L132-L140)). It still uses `auto` under the hood, it's just a different way to invoke it.

For the other checks we just delegate to the same scripts as before.

## What about the netlify config?

For netlify it looks like the sanest way to do this would be to wait until we're ready to merge, then change the build settings here: 

image

These changes should be applied:

- Base directory: [empty]
- Build command: `yarn build-docs`
- Publish directory: `packages/palette-docs/public`

## Thanks :heart:


alloy:
Nice one 👌 

I think the Palette README should stay in the root of the repo, though, so it gets shown immediately when visiting the repo.

ds300:
Thanks @alloy !

For netlify it looks like the sanest way to do this would be to wait until we're ready to merge, then change the build settings here: 

image

These changes should be applied:

- Base directory: [empty]
- Build command: `yarn build-docs`
- Publish directory: `packages/palette-docs/public`

orta:
nice stuff - yah

jonallured:
Wow, awesome!! Super excited to see this land and for netlify to reflect the PR state. ✨ ❤️

package.json
```diff
@@ -1,153 +1,20 @@
...
+    "test": "yarn workspaces run test",
+    "lint": "yarn workspaces run lint",
+    "compile-palette": "yarn workspace @artsy/palette compile",
+    "type-check": "yarn compile-palette && yarn workspaces run type-check",
+    "build-docs": "yarn compile-palette && yarn workspace artsy-palette-docs build",
+    "release": "yarn workspace @artsy/palette release"
+  },
+  "workspaces": {
+    "packages": [
+      "packages/*"
```
damassi: Our first workspace at Artsy 🎉

packages/palette-docs/package.json
```diff
@@ -13,6 +14,9 @@
     "test": "echo \"Write tests! -> https://gatsby.app/unit-testing\"",
     "type-check": "tsc --pretty --noEmit"
   },
+  "workspaces": {
+    "nohoist": ["gatsby", "gatsby-*"]
```
damassi: Was this setup by design in order to keep things cleanly separated or was there a conflict somewhere?

packages/palette-docs/src/components/GlobalComponents.jsx
```diff
@@ -73,6 +73,23 @@ export const { GlobalStyles } = injectGlobalStyles(`
...
+  code: ({ children, className: language }) => {
+    return (
+      <CodeEditor
+        code={children}
+        language={language}
+        editable={false}
+        scope={{}}
+      />
+    )
+  },
```
damassi: thanks for the alpha 👍

packages/palette-docs/package.json
ds300: The problem was that gatsby was resolving plugins relative to itself, and since gatsby doesn't know what plugins you're going to use, it can't specify them in it's package.json and yarn can't know that they must be hoisted too. so if you have a structure like this:

```
./
./node_modules/gatsby
./packages/palette-docs/gatsby-plugin-mdx
```

and then gatsby tries to resolve `gatsby-plugin-mdx`, it will fail.

See this for more deets: https://github.com/gatsbyjs/gatsby/issues/2135

damassi: 👍

artsyit:

Built with commit 013728e8ca765e1f8a0cd1667d37f5d45dea3ffd

https://deploy-preview-316--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.0.6 :rocket:

PR #317, opened 2019-03-08
Add prettier check

jonallured:
This PR aims to get prettier added to our CI setup so that formatting mistakes can be corrected when they are being introduced rather than after the fact. I've updated prettier to the latest which includes a new `check` flag (h/t @orta for finding that). This flag will list the files that have formatting mistakes and then a helpful message that maybe prettier was not run. Docs: https://prettier.io/docs/en/cli.html#check.

This is WIP until I'm able to land https://github.com/artsy/orbs/pull/25 and then I'll come back here and update the `.circleci/config.yml` file with the newer orb version and that new job.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "lint": "tslint \"src/*\" -c tslint.json tsconfig.json",
     "prepublishOnly": "yarn clean && yarn compile && yarn type-declarations",
     "prettier": "prettier",
+    "prettier-check": "prettier --check 'src/**/*.{ts,tsx}'",
     "prettier-project": "yarn prettier-write 'src/**/*.{ts,tsx}'",
     "prettier-write": "yarn prettier --write",
     "release": "auto shipit",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "docs": "cd www && yarn install && yarn start",
     "docs:build": "cd www && yarn build",
     "generate-tokens": "ts-node scripts/generate-tokens.ts",
-    "lint": "tslint \"src/*\" -c tslint.json tsconfig.json",
+    "lint": "tslint \"src/*\" -c tslint.json tsconfig.json && yarn prettier-check",
     "prepublishOnly": "yarn clean && yarn compile && yarn type-declarations",
     "prettier": "prettier",
     "prettier-check": "prettier --check 'src/**/*.{ts,tsx}'",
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 96547f6d7a36b0fe9f77a1ed560d614e63563c1d

https://deploy-preview-317--artsy-palette.netlify.com

damassi:
Echoed over in [this comment](https://github.com/artsy/orbs/pull/25#issuecomment-471005095), a formatting inconsistency is largely aesthetic, introduces no risk to a codebase, and is instantly fixed by the next person who touches the file, assuming they've correctly configured their development environment according to repo expectations. Adding a failure to CI just serves to increase friction for little gain.

ds300:
@damassi I think the good thing about enforcing formatting on CI is that it prevents PR noise for the next person who touches the file. Why should my carelessness annoy someone else and make them have to add comments about it for the people who are reviewing their PR?

damassi:
> have to add comments about it 

No comments necessary. The problem is that the project has been configured incorrectly, so helping the dev get  setup and a rebase would fix the problem. 

By failing, we're a) encouraging people to work outside of repo conventions; and b) creating unnecessary friction, as well as time / CI cycles for something that has no real bearing on the correctness of the code.

package.json
```diff
@@ -21,6 +21,7 @@
     "lint": "tslint \"src/*\" -c tslint.json tsconfig.json",
```
zephraph: I'd just append `&& yarn prettier-check` to the end of this like @ds300 was suggesting.

alloy:
@damassi We’re not encouraging people to work outside of repo conventions (as long as we’re committing any prettier changes from CI), quite the opposite. People that wouldn’t be running the hooks _and_ forget to otherwise run prettier are the ones that will get a failure and will be reminded to do so immediately next time.

I don’t think it should add unrelated noise to another person’s PR in the same file, as I will question the unrelated changes there.

I’m +1 on this.

alloy:
> No comments necessary. The problem is that the project has been configured incorrectly, so helping the dev get setup and a rebase would fix the problem.

I think @ds300 is referring to the next person working in the same file needing to add a comment to their PR saying “I made no actual changes here 🤷‍♂️“

damassi:
Ahh, that makes sense -- meaning, the comment is about the auto-formatting, not the lack thereof.

jonallured:
Ok, I updated this one with feedback so that `lint` will include prettier and I think it's ready for final thoughts. Thanks @ds300 for the nudge in the right direction! ❤️

artsyit:
:rocket: PR was released in v3.2.7 :rocket:

alloy:
Nice one @jonallured, thanks! 🙌

PR #321, opened 2019-03-11
[Tooltip] Add option to have component as tooltip content

sepans:
Mainly to be able to have lists and paragraphs in the tooltip.

Screen Shot 2019-03-11 at 4 08 40 PM

### DIFF
--- a/src/elements/Tooltip/Tooltip.tsx
+++ b/src/elements/Tooltip/Tooltip.tsx
@@ -58,7 +58,7 @@ enum TipSize {
 }
 
 export interface TooltipProps {
-  content: string
+  content: string | React.ReactElement<any>
   size: TipSize
   width: number
 }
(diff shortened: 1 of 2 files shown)


sepans:
hmmm, don't know what netlify fails 🤔

src/elements/Tooltip/Tooltip.tsx
```diff
@@ -58,7 +58,7 @@ enum TipSize {
 }
 
 export interface TooltipProps {
-  content: string
+  content: string | React.ReactElement<any>
```
ds300: This type should be `React.ReactNode` which includes `string` and other things not covered by `ReactElement`

ds300:
So the netlify failure is not your fault. It's because you updated the docs to have an example that passes something other than a string, but the docs are still being built with an old verison of palette which can only handle strings. We're fixing this really soon in #316 but for now I'm going to merge this with the netlify failure and then update palette for the docs in a separate PR to let this change be deployed, then I'll merge #316 and we will be free of these inconsistency issues 💆

artsyit:
:rocket: PR was released in v4.0.4 :rocket:

PR #327, opened 2019-03-13
Add 'element' property for specifying element type on Typography

pepopowitz:
This PR adds a prop named `element` to the Typography components, allowing the user to specify what type of element gets rendered. The `element` prop flows through to styled-components as [an `as` prop](https://www.styled-components.com/docs/api#as-polymorphic-prop). 

image

## Why not name the prop `as`, like styled-components wants?

I attempted to do this, but I ran into a confusing problem. Whenever I specified `as` on a JSX element, the rendered element would lose all other styles that were specified. In fact, the function [`createStyledText`](https://github.com/pepopowitz/palette/blob/62d94bb314a58be2bd93eaece1a87ac3438638d1/packages/palette/src/elements/Typography/Typography.tsx#L171) wouldn't even get called. I dug as far as I could, but I don't really know why it hates the prop being named `as`.

If you've got the courage to investigate why I can't name it `as`, I encourage you. I feel like this accomplishes the goal of being able to specify the element type, even though we'd probably all prefer it be named `as`. I'm also happy to pair if someone feels like they have ideas on the issue.

---

One other thing to note: we'll need to make sure we are properly resetting styles wherever this is consumed. Chrome applies default margins to headings, which we won't want. It looks to me like Force is applying a reset that will eliminate those default margins (see below), but I don't know if I need to worry about anything on the mobile side. 

image

### DIFF
--- a/packages/palette/src/elements/Typography/Typography.tsx
+++ b/packages/palette/src/elements/Typography/Typography.tsx
@@ -185,9 +185,13 @@ function createStyledText<P extends StyledTextProps>(
             fontFamilyType && themeProps.fontFamily[fontType][fontFamilyType]
           }
           {...determineFontSizes(fontType, size)}
-          // styled-components wants it called `as`, but this function somehow
-          //  gets skipped when there is an `as` prop passed in. So our clients
-          //  will pass in `element`, and we'll pass it through as `as`.
+          // styled-components supports calling the prop `as`, but there are
+          //  issues when passing it into this component named `as`. See
+          //  https://github.com/styled-components/styled-components/issues/2448
+          //  & https://github.com/artsy/palette/pull/327#issuecomment-473434537
+          //  for context.
+          // So we are naming it `element` on the way into this component, and
+          //  renaming it to `as` when we pass it to through.
           {...(element ? { as: element } : {})}
           {...textProps}
         />
--- a/packages/palette/src/elements/Typography/Typography.tsx
+++ b/packages/palette/src/elements/Typography/Typography.tsx
@@ -91,6 +91,7 @@ export interface TextProps
    * this prop.
    */
   ellipsizeMode?: string
+  element?: keyof JSX.IntrinsicElements | React.ComponentType<any>
 }
 
 /** Base Text component for typography */
@@ -172,7 +173,7 @@ function createStyledText<P extends StyledTextProps>(
   selectFontFamilyType: typeof _selectFontFamilyType = _selectFontFamilyType
 ) {
   return styled<P>(
-    ({ size, weight, italic, ...textProps }: StyledTextProps) => {
+    ({ size, weight, italic, element, ...textProps }: StyledTextProps) => {
       const fontFamilyType = selectFontFamilyType(_fontWeight(weight), italic)
       // This is mostly to narrow the type of `fontFamilyType` to remove `null`.
       if (fontFamilyType === null) {
@@ -184,6 +185,10 @@ function createStyledText<P extends StyledTextProps>(
             fontFamilyType && themeProps.fontFamily[fontType][fontFamilyType]
           }
           {...determineFontSizes(fontType, size)}
+          // styled-components wants it called `as`, but this function somehow
+          //  gets skipped when there is an `as` prop passed in. So our clients
+          //  will pass in `element`, and we'll pass it through as `as`.
+          {...(element ? { as: element } : {})}
           {...textProps}
         />
       )
--- a/packages/palette-docs/content/docs/tokens/Typography.mdx
+++ b/packages/palette-docs/content/docs/tokens/Typography.mdx
@@ -9,37 +9,39 @@ should be contextually applied. The goal is for this system to increase overall
 design consistency for our product as well as to increase efficiency for both
 the design and engineering teams.
 
-<Flex mt={4} justifyContent="space-between">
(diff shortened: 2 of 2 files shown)


packages/palette/src/elements/Typography/Typography.tsx
```diff
@@ -184,6 +185,10 @@ function createStyledText<P extends StyledTextProps>(
             fontFamilyType && themeProps.fontFamily[fontType][fontFamilyType]
           }
           {...determineFontSizes(fontType, size)}
+          // styled-components wants it called `as`, but this function somehow
+          //  gets skipped when there is an `as` prop passed in. So our clients
+          //  will pass in `element`, and we'll pass it through as `as`.
+          {...(element ? { as: element } : {})}
```
damassi: Do you tink we should stick to styled-components conventions? Is it a bug that `as` isn't getting picked up?

alloy:
I would love to pair debug this, but today I’m focussing on LD work. Perhaps @zephraph or @damassi have some time to help today, otherwise let’s do so tomorrow?

In the meantime, did you search the styled-components repo issues for similar reports?

> It looks to me like Force is applying a reset that will eliminate those default margins (see below), but I don't know if I need to worry about anything on the mobile side.

By “the mobile side” are you referring to usage in React Native? If so, I don’t _think_ we need to worry about that.

zephraph:
@alloy I did pair with @pepopowitz on this. It’s not clear why `as` fails. I might poke at it again, but it’s a pretty weird/obscure/hard to debug issue. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

damassi:
I can givre this a look today, i'm curious!

pepopowitz:
@alloy Yes, by "mobile" I meant usage in the Eigen/Emission stack. Sorry for the lack of clarity.

I haven't found any existing issues in styled-system or styled-components, but I don't _think_ it is an issue there. I'm able to use the `as` prop with both of those packages in standalone POC's via codesandbox. And in fact, things are working fine in this PR when I pass the prop through to styled-system named `as`. It's when the prop is passed into our Typography components named `as` that it has problems.

@damassi Let me know if you want a guided tour of what we know so far. Happy to pair on it!

pepopowitz:
For any interested parties, I pushed [another branch with the desired API](https://github.com/pepopowitz/palette/tree/typography-element-type-with-as) (using `as` instead of `element`). [Here is the diff from this PR](https://github.com/pepopowitz/palette/compare/typography-element-type...pepopowitz:typography-element-type-with-as). Below, you can see what ends up happening to the JSX elements that specify `as`:

image


At (1), I'm rendering a `Serif` element with a `size`, `color`, and `as="h1"` prop. 
At (2) you can see in the rendered output that it is no longer using the correct typeface and it is the wrong color. 
At (3) I'm rendering a `Serif` with a `size` and `color`, but no `as` prop. 
At (4) you see that the `color="red"` prop works, so it _should_ be taking effect in the first heading.
At (5) you can see that the element is rendered as an h1, and it _looks_ like it should have some styled-components classes applied.
At (6) you can see that there aren't any styled-components classes applied.

pepopowitz:
And here is a codesandbox that shows styled-system working correctly, in isolation. (There's a whole lot of other noise in there, but you can see the two `<Text>` elements being rendered with an `as` prop, and retaining the other applied styles.)

https://codesandbox.io/s/n6lrk57km

image

damassi:
@pepopowitz - I just pulled down your new branch to take it for a spin and it _seems_ correct over here? 

Screen Shot 2019-03-14 at 2 10 57 PM

And the devtools: 

Screen Shot 2019-03-14 at 2 12 28 PM

Am I missing something?

damassi:
Here's it updating in real time: 


updates

damassi:
(Ah, scratch that -- no proper classes being applied after doing some SH with @pepopowitz)

pepopowitz:
I'm going to let this PR sit until Monday, in case anyone else wants to take a look at it. Let me know if anyone needs a demo, wants to pair, or has questions!

artsyit:

Built with commit 02a0ff6c9527ecff7e28d4ac6c0ec22cb75c3580

https://deploy-preview-327--artsy-palette.netlify.com

zephraph:
For those following along I was able to replicate the issue. https://codesandbox.io/s/182ovy2ylq

Turns out it _is_ a styled-components bug... or at least if not a bug, a nuance that makes it where we can't use `as` with how we currently have our typography components architected. 

The specific issue is the `styled` wrapper that can be seen [here](https://github.com/artsy/palette/blob/f63805d54ebd77ff13f408b3cbdc860468308dee/packages/palette/src/elements/Typography/Typography.tsx#L174). That wrapper is constructing a new component that gets the `as` prop and fundamentally changes how the component renders. From what we're seeing, the child components are never rendered and the props are just passed to this newly created component.

zephraph:
Given the upstream challenges, I think we should merge this as is. The api drift from `Typography` components and other components is unfortunate, but it is what it is. 

Hopefully we'll get some upstream solutions.

pepopowitz:
Cool - thank you all for taking time to look into this. I'm going to push a comment pointing at the styled-components issue, and then I'll merge.

artsyit:
:rocket: PR was released in v4.1.3 :rocket:

PR #333, opened 2019-03-18
Add CSS reset for headings (h1-h6) to global styles

pepopowitz:
Force contains a reset that wipes out agent/user-agent styles for headings, but reaction/palette do not. Headings can look different in the palette docs & reaction storybooks than they do when they are deployed via Force.

For example, here's what an h2 looked like in the artwork sidebar in Reaction's storybooks:

headings-before

And here's what it looked like when I linked it into Force:

headings-after

Adding this h1-h6 reset should make it easier for engineers to build things in storybooks, and have them look identical in Force.

### DIFF
--- a/packages/palette/src/helpers/injectGlobalStyles.tsx
+++ b/packages/palette/src/helpers/injectGlobalStyles.tsx
@@ -129,6 +129,14 @@ export function injectGlobalStyles<P>(
       }
     }
 
+    h1, h2, h3, h4, h5, h6 {
+      font-style: inherit;
+      font-family: inherit;
+      font-weight: inherit;
+      font-size: inherit;
+      margin: 0;
+    }
+    
     ${additionalStyles};
   `
 


packages/palette/src/helpers/injectGlobalStyles.tsx
```diff
@@ -129,6 +129,14 @@ export function injectGlobalStyles<P>(
...
     }
 
+    h1, h2, h3, h4, h5, h6 {
+      font-style: inherit;
+      font-family: inherit;
+      font-weight: inherit;
+      font-size: inherit;
+      margin: 0;
+    }
+
```
damassi: @pepopowitz - would you mind taking care of this [bit of tech-debt](https://github.com/artsy/reaction/blob/master/src/Artsy/Router/Components/Boot.tsx#L36-L42)?

artsyit:

Built with commit 4b261d958f8b064856dc7dc31ace66102e128481

https://deploy-preview-333--artsy-palette.netlify.com

packages/palette/src/helpers/injectGlobalStyles.tsx
pepopowitz: I meant to mention that I'd clean that up, but forgot 😁

damassi: ty 👍 cc @eessex

artsyit:
:rocket: PR was released in v4.2.1 :rocket:

PR #334, opened 2019-03-18
Create LazyImage component, allow avatar to be lazy

zephraph:
This PR abstracts out the [lazy-load behavior](https://github.com/artsy/reaction/pull/2064) added to the artwork grid in reaction and creates a new `LazyImage` component. 

It's a bit rough around the edges right now, but it'll get better as we iterate. 

The original purpose of this work was to make the `avatar` lazy loadable. I've started that work (well, the lazy loading part is done), but I need to untangle everything and pull the lazy-loading logic into a web only version.


artsyit:

Built with commit bfaf6ad4bc59652a0ff13ff13d8b79458611643a

https://deploy-preview-334--artsy-palette.netlify.com

packages/palette/src/elements/LazyImage/LazyImage.tsx
```diff
@@ -0,0 +1,94 @@
...
+const pulse = keyframes`
+  0% { background-color: ${color("black10")}; }
+  50% { background-color: ${color("black5")}; }
+  100% { background-color: ${color("black10")}; }
+`
+
+const pulseAnimation = () =>
+  css`
+    ${pulse} 2s ease-in-out infinite;
+  `
```
damassi: At some point there's an intention to move these animations out into their own folder; it might be worth adding a `TODO: Move animations` (or something) here so that we'll know to swing by this component to collect.

packages/palette/src/elements/LazyImage/LazyImage.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+interface LazyImageProps
+  extends ImageProps,
+    WidthProps,
+    HeightProps,
+    BorderRadiusProps {
+  preload?: boolean
+  placeholderProps?: any
+  style?: CSSProperties
+}
```
damassi: Can we add a comment around `preload` and `placeholder` props? Also, I can see this being extended so it might be good to export the interface (and add a comment so the linter doesn't complain).

packages/palette/src/elements/Avatar/Avatar.ios.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FunctionComponent } from "react"
+import { styled } from "../../platform/primitives"
+import { AvatarProps, BaseAvatar, sizeValue } from "./Avatar.shared"
```
damassi: 👍

zephraph:
Okay @damassi I refactored it to allow the lazy loading of the normal image components. Both `Image` and `ResponsiveImage` have the prop `lazyLoad` which allows them to be lazily loaded. Like with avatar, I exposed mobile versions that don't use the lazy load functionality. 

I had some general typescript faffing that I eventually just gave up on. See [here](https://github.com/artsy/palette/blob/lazy-load-avatar/packages/palette/src/elements/Image/LazyImage.tsx#L49-L50)

packages/palette-docs/content/docs/elements/images/Avatar.mdx
```diff
@@ -3,7 +3,15 @@ name: Avatar
 ---
 
 <Playground title="Medium">
-  <Avatar size="md" src="https://picsum.photos/110/110/?random" />
+  <Flex>
+    <Avatar size="md" src="https://picsum.photos/110/110/?random" />
+    <Spacer ml={1} />
+    <Avatar
+      size="md"
+      lazyLoad
```
damassi: 👍

packages/palette-docs/content/docs/elements/images/Image.mdx
```diff
@@ -19,3 +19,15 @@ Used to fill an area provided regardless of proportions.
...
+
+Useful for images that render below the fold
+
+<Playground>
+  <LazyImage
+    width="300px"
+    height="200px"
+    src="http://deelay.me/3000/https://picsum.photos/300/200/?random"
+  />
+</Playground>
```
damassi: @zephraph - can you remove the lazy component in favor of the regular `<Image>` component with a `lazyLoad` prop? That or add and then move `LazyImage` down down with the disclaimer that its a lower-level primitive and that the regular `<Image>` component should be used. We should keep the number of components in circulation to a minimum if possible.

packages/palette/src/elements/Image/Image.tsx
```diff
@@ -1,70 +1,25 @@
...
-  width: "100%",
-  ratio: 1,
-}
+export const ResponsiveImage = ({
+  lazyLoad = false,
+  ...props
+}: ResponsiveImageProps) => (
+  <LazyImage
+    preload={!lazyLoad}
+    imageComponent={BaseResponsiveImage}
```
damassi: 👍

damassi:
One comment about adding example docs around 
```
<Image lazyLoad />
```
and moving `<LazyImage>` down to discourage direct use. 

The refactor looks 👌. 

Feel free to merge when the docs update is all done!

artsyit:
:rocket: PR was released in v4.3.0 :rocket:

PR #337, opened 2019-03-20
Go back to auto-release step in orb

zephraph:
Auto isn't strictly needed as a dependency of the project and the `auto-release` orb job _should_ properly deploy. I'm restoring this to use the old approach to be consistent with our other deployments and to remove the extra dep. 

This would normally be a `Trivial`, but I'm setting it to `Patch` in order to ensure the release process still works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ workflows:
       - yarn/test:
           requires:
             - yarn/workflow-queue
-      - yarn/release:
+      - yarn/auto-release:
           # The deploy job is the _only_ job that should have access to our npm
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 148d09f0aa302d4c6a4850efed1a2a2b993c313b

https://deploy-preview-337--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.3.1 :rocket:

ds300:
I considered changing it back but I'm not a fan of using `npx` on circle, since it bypasses a) the yarn lockfile (if auto gets a major update it could break our deploys), and b) the npm package cache.

PR #345, opened 2019-03-25
Fix web Collapse component

ds300:
https://artsyproduct.atlassian.net/browse/PURCHASE-942

bad-collapse-2

On the purchase team we have noticed a couple of issues with the Collapse component animating itself to incorrect heights, or not animating at all when nested. Unfortunately we were not able to resolve them with the current implementation by following the advice outlined [in the react-spring docs](https://www.react-spring.io/docs/props/gotchas#animating-auto)

I therefore propose we use a custom Collapse component with transparent internals. This amounts to ~100 lines of well-commented code.

This approach has a couple of other benefits:

- Lifting the burden on users of Collapse to know about the `height: auto` gotcha and how to mitigate it.
- Reducing web bundle size by 10k gzipped. (Collapse is the only web component so far that uses react-spring)

This is not to say that we should not use `react-spring` in the future. Just that it is not currently appropriate for this Collapse use case.

### DIFF
--- a/packages/palette/src/elements/Collapse/Collapse.tsx
+++ b/packages/palette/src/elements/Collapse/Collapse.tsx
@@ -1,35 +1,89 @@
 import React from "react"
-import { animated, Spring } from "react-spring"
 
-export interface CollapseProps {
-  /** Determines whether content is expanded or collapsed */
-  open: boolean
-}
+/**
+ * Collapse component for the web
+ */
+export class Collapse extends React.Component<{ open: boolean }> {
+  wrapperModifyTimeout: ReturnType<typeof setTimeout>
+  wrapperRef: HTMLDivElement | null = null
 
-/** Collapses content with animation when open is not true */
-export class Collapse extends React.Component<CollapseProps> {
-  state = {
-    mounted: false,
+  onTransitionEnd = (ev: TransitionEvent) => {
+    if (!this.wrapperRef) {
+      return
+    }
+    if (ev.propertyName === "height") {
+      this.wrapperRef.style.height = this.props.open ? "auto" : "0px"
+    }
+  }
+
+  firstRender: { open: boolean } | null = {
+    open: this.props.open,
   }
 
   componentDidMount() {
-    this.setState({ mounted: true })
+    if (!this.wrapperRef) {
+      return
+    }
+
+    this.wrapperRef.addEventListener("transitionend", this.onTransitionEnd)
+  }
+
+  componentDidUpdate() {
+    if (!this.wrapperRef) {
+      return
+    }
+    if (this.props.open && this.wrapperRef.style.height !== "auto") {
+      // open
+      const prevHeight = this.wrapperRef.style.height || "0px"
+      this.wrapperRef.style.height = "auto"
+      const actualHeight = this.wrapperRef.offsetHeight
+      this.wrapperRef.style.height = prevHeight
+      this.wrapperModifyTimeout = setTimeout(() => {
+        this.wrapperRef.style.height = actualHeight + "px"
+      }, 10)
+    } else if (!this.props.open && this.wrapperRef.style.height !== "0px") {
+      // close
+      const actualHeight = this.wrapperRef.offsetHeight
+      this.wrapperRef.style.height = actualHeight + "px"
+      this.wrapperModifyTimeout = setTimeout(() => {
+        this.wrapperRef.style.height = 0 + "px"
+      }, 10)
+    }
+  }
+
+  componentWillUnmount() {
+    this.wrapperRef.removeEventListener("transitionend", this.onTransitionEnd)
+    clearTimeout(this.wrapperModifyTimeout)
   }
 
   render() {
+    const { children, open } = this.props
+    // render explicit height before first change, to let us render closed
+    // elements on the server
+    let heightProps = {}
+    if (this.firstRender) {
+      // render() might be called multiple times before the first time `open` changes
+      if (this.firstRender.open !== open) {
+        // `open` prop has changed for the first time
+        // ditch explicit height and let `componentDidUpdate` take the wheel
+        this.firstRender = null
+      } else {
+        heightProps = {
+          height: open ? "auto" : "0px",
+        }
+      }
+    }
     return (
-      <Spring
-        native
-        immediate={!this.state.mounted}
-        from={{ height: 0 }}
-        to={{ height: this.props.open ? "auto" : 0 }}
+      <div
+        ref={ref => (this.wrapperRef = ref)}
+        style={{
+          transition: "height 0.3s ease",
+          overflow: "hidden",
+          ...heightProps,
+        }}
       >
-        {props => (
-          <animated.div style={{ ...props, overflow: "hidden" }}>
-            {this.props.children}
-          </animated.div>
-        )}
-      </Spring>
+        {children}
+      </div>
     )
   }
 }
--- a/packages/palette/src/elements/Collapse/Collapse.tsx
+++ b/packages/palette/src/elements/Collapse/Collapse.tsx
@@ -11,13 +11,12 @@ export class Collapse extends React.Component<{ open: boolean }> {
     if (!this.wrapperRef) {
       return
     }
+    // when animating open, we set the wrapper's height to an explicit pixel
+    // value. When the transition ends we want to set the height back to 'auto'
+    // so that if the content of the wrapper changes it's height will change too
     if (ev.propertyName === "height") {
       this.wrapperRef.style.height = this.props.open ? "auto" : "0px"
     }
-  }
-
-  firstRender: { open: boolean } | null = {
-    open: this.props.open,
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit 3c14568c7b79159a4899f9331cb227484e2f5285

https://deploy-preview-345--artsy-palette.netlify.com

artsyit:

Built with commit 64401ede0a2aff3b27ffe431c4a42c4acc213264

https://deploy-preview-345--artsy-palette.netlify.com

damassi:
One follow-up req: if `react-spring` isn't being used in the codebase can that be removed as a dep?

artsyit:
:rocket: PR was released in v4.4.2 :rocket:

ds300:
We need to keep `react-spring` for now. It is still used in the RN version of Collapse.

Thanks for your thoughts on prioritising OSS usage 🙏Will definitely keep that in mind for the future.

alloy:
Wether or not a OSS library has [partially] covered our needs does not mean we don’t pay a price.

Without researching nor speaking to react-spring specifically, taking on dependencies (and _their_ dependencies) that do more than what we actually need has maintenance costs too.

A few, but definitely not an exhaustive list:

* Code in the lib that we _don’t_ use or dependencies of a lib for features we don’t use may rely on APIs of other dependencies such as React, which makes it harder to upgrade those more important dependencies.
* Addressing our needs (be it fixing bugs or building out features) may require extra implementation time due to having to consider complexity in the library related to other parts of the library (that we may not use).
* More need for build chain faffing (dependency resolution, compiler configuration, and so many more things you don’t really want to spend time on when doing product work).
* More surface area for security issues (supply chain attacks are a hot topic in infosec, [here’s](https://blog.npmjs.org/post/180565383195/details-about-the-event-stream-incident) a recent example that I could totally see us running into).

Additionally, OSS libraries and their maintainers/communities cannot be considered a homogenous bunch. Some are stricter than others in who they will accept changes from, some simply have smarter people working on them than others, and so many more reasons that to make this not necessary a selling point of OSS.

In the end, to me, the rubric roughly boils down to:

* If the total amount of code added to a project is much greater than the custom code needed to exactly address our needs, then we should go with custom code. _(I.e. the library, its transitive dependencies, _and_ the custom code that builds on the lib’s abstractions.)_
* If we know concretely that we will be using _significant_ amounts of other features of the library _in a known short timeframe_ and we have considered the library well enough over other offerings, then we can choose to pay the price that comes with 3rd party dependencies. _(I.e. for planned work, not “probably” somewhere down the line, because by that time our overal requirements might have changed in a way that would require us to ditch the dependency [e.g. in favour of another dep] all the while paying the price of having taken on the dependency.)_

alloy:
@damassi As for this specific PR’s needs, if you know that we are simply using the library incorrectly, then we should prioritize time for you (or someone else) to address that and leave us with a good example for others to follow, going forward.

packages/palette/src/elements/Collapse/Collapse.tsx
```diff
@@ -1,35 +1,102 @@
...
+      // animate opening
+      // measure goal height
+      const prevHeight = this.wrapperRef.style.height || "0px"
+      this.wrapperRef.style.height = "auto"
+      const goalheight = this.wrapperRef.offsetHeight
+      this.wrapperRef.style.height = prevHeight
+      // wait for a tick before setting goal height to allow transition
+      this.wrapperModifyTimeout = setTimeout(() => {
+        this.wrapperRef.style.height = goalheight + "px"
+      }, 10)
```
alloy: Why `setTimeout(…, 10)` over `setImmediate(…)` if the need is to perform in on the next tick of the runloop?

ds300: `setImmediate` is not on the standards track any more. https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate. `requestAnimationFrame` might be the most conceptually 'correct' to use here, but `setTimeout` works and was chosen through muscle memory more than anything else.

alloy: Good to know!

l2succes:
Aside from the points @damassi mentioned above for using `react-spring`. One important thing to consider as we add and modify components in palette is our usage in emission. We make use of the `Collapse` in emission in the following component.
https://github.com/artsy/emission/blob/master/src/lib/Components/HoursCollapsible.tsx#L1

Although emission is currently on an older version of palette. We've finally been able to upgrade to react native v0.57. This unblocks us to upgrade palette in turn which I'm currently working on, I still believe that are benefits in not having to relearn

`react-spring` having a declarative API that supports for both react and react-native was a driving factor in adopting it in the first place. As of now we have a series of universal components that work for both platforms without needing any platform specific code aside from our universal primitives. This  is a great abstraction I wouldn't like to see us get away from if we can avoid it.

l2succes:
All that being said, @ds300 I understand you made the decision because of the issue you described,  animating "height: auto" is a hard problem with any animation library by the way. The following ticket describes the problem at length https://github.com/react-spring/react-spring/issues/30. 

And here's the full solution in CodeSandbox: https://codesandbox.io/s/pmjomxn60

```javascript
import React from 'react'
import ReactDOM from 'react-dom'
import { Spring, animated } from 'react-spring'
import './styles.css'

// Why <Spring force ... ?
// Because we are rendering the spring with the same props on add/remove text
// (height: 'auto'), so it has no reason to update its animations. This is an
// edge case unfortunately. More on that here: https://github.com/drcmda/react-spring/blob/master/API-OVERVIEW.md#animating-auto

const LOREM = `Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy
              text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has
              survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was
              popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop
              publishing software like Aldus PageMaker including versions of Lorem Ipsum.`

class App extends React.Component {
  state = { toggle: true, text: [LOREM] }
  onToggle = () => this.setState(state => ({ toggle: !state.toggle }))
  onAddText = () => this.setState(state => ({ text: [...state.text, LOREM] }))
  onRemoveText = () => this.setState(state => ({ text: state.text.slice(1) }))
  render() {
    const { toggle, text } = this.state
    return (
      <React.Fragment>
        <button onClick={this.onToggle}>Toggle</button>
        <button onClick={this.onAddText}>Add text</button>
        <button onClick={this.onRemoveText}>Remove text</button>
        <div className="content">
          <Spring native force from={{ height: 0 }} to={{ height: toggle ? 'auto' : 0 }}>
            {props => (
              <animated.div className="item" style={props}>
                {text.map(t => <p>{t}</p>)}
              </animated.div>
            )}
          </Spring>
        </div>
      </React.Fragment>
    )
  }
}

ReactDOM.render(<App />, document.getElementById('root'))
```


Definitely don't want to block you guys on feature work and this has already been merged but I'd like us to revisit this decision, happy to pair on a fix.

ds300:
@l2succes There's a separate Collapse component for RN, added by Daniel R 4 months ago: https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Collapse/Collapse.ios.tsx

It uses `react-spring` under the hood. Although again it is doing custom state management and content measuring on top of `react-spring`.

l2succes:
@ds300 Yes that's true. We're actually on an older version of `react-spring` which is what the RN version of Collapse uses. The solution I posted above would work for both web and iOS.

damassi:
> As for this specific PR’s needs, if you know that we are simply using the library incorrectly, then we should prioritize time for you (or someone else) to address that and leave us with a good example for others to follow, going forward.

I took time to resolve the issue during the second collapse code PR; there were docs on the react-spring that were overlooked regarding `height: auto` and I created a PR and opened up a PR to the fork in order to address the change.

Again, as mentioned above, this is the third time this custom code has been introduced, which @ds300 feels strongly about and I'm cool to merge. We *aren't* using `react-spring` elsewhere. I could repeat all of my points from the previous two PRs + surface other long discussions that we've had before then, but I don't want to labor the point. I hear what you're saying though, @alloy.

ds300:
@l2succes I tried the version in your snippet. It didn't fix the issue. In fact it made it worse. Maybe it needed something extra, but at this point I don't want to spend any more time debugging `react-spring`'s `height: auto` handling.

> this is the third time this custom code has been introduced, which @ds300 feels strongly about

It's not so much this custom code vs `react-spring` that I feel strongly about. For the record I'll sum up my thoughts on the wider discussion in here without directly addressing any more individual points people have contributed:

- I feel that providing good UX is more important than having good DX, or just about any other 'technical' concern we can think of.
- I feel like we should not be discouraged (or discourage others) from using the platform out of a fear of complexity. Rather we should empower ourselves and others to know when the platform is the best choice by learning its fundamentals and its shortcomings.
- I feel like we might be overestimating the quantity and complexity of animations that palette will provide in the future. Declarative css transitions go a really long way (see, e.g. BarChart)
- I feel that the benefit of being able to share the exact definition of primitive (i.e. elemental) components between web and native is quite insignificant. The rewards of cross-platform components are mainly reaped at higher levels of abstraction when building more complex features, so implementing+maintaining the bottom layer twice is not a big deal.
- I feel that leaking implementation details, like the `height: auto` gotcha, out of a palette component is undesirable. Especially when it can be easily overlooked during development and only surface as an issue when it's actually in front of users.

damassi:
> I feel that providing good UX is more important than having good DX, or just about any other 'technical' concern we can think of

This, however, comes at a cost. Developer experience is as much about maintainability as it is about anything else. For something low-level like animation, abstracting that away behind a library that is common, easy to use, and considered best in class for react users is reasonable. It's very often the case in programming that devs create ivory towers out of their creations, which are reachable only by themselves. Devs come and go. By sticking to OSS for things like this longevity can be achieved. Not saying we should _never_ write our own implementations, but for any sufficiently complex UI there should be a discussion around whether it's needed.

The `BarChart` is a good example of what I mean. It's solid, and the code is pretty straight forward to those who are familar with JS internals. However, did we need to lay down the beginnings of a charting library? Could it have lived behind a library such as Victory (as an arbitrary example). Did writing this from scratch solve an existing problem? Now, if other charts and graphs are needed, we'll be looking to our custom implementation and using the momentum that's been developed behind in-house code to create more in-house code, since moving to something 3rd party would create divergence. We shouldn't be reinventing BarCharts -- or any charts -- even though what was created is really nice, and is great UX. Charting is a solved problem that we shouldn't have to worry about maintaining.

PR #346, opened 2019-03-25
Begin stripping attributes from palette elements

zephraph:
This loosely depends on https://github.com/styled-system/extras/pull/3 before it can be completed. 

We've had issues for a while where props from styled-system was getting passed down to the DOM. Recently our configuration has changed in reaction to make warnings fail tests. React is warning about these props being passed to DOM elements which cause local tests to fail. 

This PR introduces `clean-tag`, a mechanism to be used on top of `styled-systems` to remove props that are filtering down to the DOM. 

**NOTE**: There are some test failures here because we were relying on those DOM properties in some tests. Those tests will need to be updated.

### DIFF
--- a/packages/palette/src/elements/Flex/Flex.tsx
+++ b/packages/palette/src/elements/Flex/Flex.tsx
@@ -1,4 +1,5 @@
-import { styled as primitives } from "../../platform/primitives"
+import tag from "clean-tag"
+import styled from "styled-components"
 
 import {
   alignContent,
@@ -63,7 +64,7 @@ export interface FlexProps
 /**
  * A utility component that encapsulates flexbox behavior
  */
-export const Flex = primitives.View<FlexProps>`
+export const Flex = styled(tag)<FlexProps>`
   display: flex;
   ${alignContent};
   ${alignItems};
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit 38c3ee36687509aa3e670d3267a291bc3f13bbeb

https://deploy-preview-346--artsy-palette.netlify.com

damassi:
Thoughts on `styled-system/extras` being considered legacy/experimental? 

Could we open a PR that fixes the current issue manually (to unblock tests in reaction) and then follow up with this? I've noticed in the past that jxnblk sometimes takes a while with PRs. 

I think there are only two test issues -- `ariaLabel` to `aria-label` and maybe one other thing.

packages/palette/src/elements/Flex/Flex.tsx
```diff
@@ -63,7 +64,7 @@ export interface FlexProps
 /**
  * A utility component that encapsulates flexbox behavior
  */
-export const Flex = primitives.View<FlexProps>`
+export const Flex = styled(tag)<FlexProps>`
```
ds300: isn't this going to break usages of Flex in RN?

zephraph:
It's unfortunate that styled-systems doesn't handle this better out of the box. Because of how it hooks into styled-components though, I guess it can't. 

I'm going to approach this from another angle.

PR #350, opened 2019-03-27
[RadioGroup] Add deselectable option

damassi:
Fixes https://github.com/artsy/palette/issues/349

Adds a new `deselectable` prop to `<RadioGroup>` which, if `true`, allows an item within a group to be toggled off and the selected item to be set to `null`. 

desel


mzikherman:
Looks great!

artsyit:
:rocket: PR was released in v4.5.0 :rocket:

PR #351, opened 2019-03-27
Fix debounce

damassi:
Tests for a fix in Reaction tests.


artsyit:
:rocket: PR was released in v4.5.1 :rocket:

PR #348, opened 2019-03-25
First step reorg

jonallured:
Let's see what CircleCI thinks of it.

PR #347, opened 2019-03-25
Re org

jonallured:
Let's see what CircleCI thinks of it.

PR #353, opened 2019-03-28
Deal with warnings

jonallured:
I'm coming back to palette after a few days away to work on reorganizing things (see #347 and #348) and with fresh eyes I saw an opportunity to clean up some warnings. I updated some weirdo prop names on the LazyImage so that the console.log messages would go away and then added a bunch of missing JSDoc comments (many of which I added! 😝 ). Check it:

### before

Screen Shot 2019-03-28 at 2 51 17 PM
Screen Shot 2019-03-28 at 2 51 03 PM

### after

Screen Shot 2019-03-28 at 2 50 17 PM

So now you can fit all the output on a single screen. 😎

### DIFF
--- a/packages/palette/src/elements/Image/LazyImage.tsx
+++ b/packages/palette/src/elements/Image/LazyImage.tsx
@@ -53,6 +53,8 @@ interface LazyImageProps
   /** The image component to render when preload is true */
   imageComponent?: any // FunctionComponent<ImageProps>
 }
+
+/** LazyImage */
 export const LazyImage: React.FC<LazyImageProps> = ({
   preload = false,
   imageComponent: ImageComponent = Image,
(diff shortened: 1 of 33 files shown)


packages/palette/src/elements/Image/LazyImage.tsx
```diff
@@ -83,8 +85,8 @@ export const LazyImage: React.FC<LazyImageProps> = ({
         src={src}
         title={title}
         alt={alt}
-        ariaLabel={ariaLabel}
-        borderRadius={borderRadius}
+        aria-label={ariaLabel}
+        borderradius={borderRadius}
```
zephraph: Hmmm... I don't think this is how to solve the problem. 

Ultimately, the error is that the `Image` element that styled-components is rendering is getting `ariaLabel` passed to it and react is complaining about that. Your change might stop that error from happening, but it'll break the image component because the props are different. And the `aria-label` is invalid JSX I believe. 

I've had it on my todo list to fix this problem and I can do so in a separate PR.

jonallured: Cool, I'll rebase that commit out and defer to #355 for this fix. 👍

jonallured:
Ok this is ready, thanks!

damassi:
👍

artsyit:
:rocket: PR was released in v4.5.3 :rocket:

PR #354, opened 2019-03-28
Reorg step 1: components and tokens

jonallured:
This PR supersedes #348 and #347 and sees me taking just a couple parts of the overall reorg project. What I've done here is create the following folders:

* packages/palette/src/components
* packages/palette/src/tokens

Based on what I found inside the organization of the docs folders I then moved the appropriate files over and altered import statements to get things linked back together.

Next steps would be to dig into the packages/palette/src/elements folder and reorganize that into subfolders. This was a lot of work, so I wanted to get this on master before too long in order to avoid merge conflicts, etc.


artsyit:

Built with commit eb7f3c5e982b9a413e9fd9849ec4a44810c27bb9

https://deploy-preview-354--artsy-palette.netlify.com

zephraph:
I don't think I have the full context on this. 

@jonallured what's the goal here? Is it just to mirror the docs structure? I have a few concerns from an organization/terminology perspective. 

For example, when we talk about design tokens, that's really the values that are in the `theme.tsx` file. I'm not sure it makes sense to group `Spacer` or `Typography` components in a tokens directory. 

I guess more so than that, what's the problem we're trying to solve here?

damassi:
I’ve been thinking about this on my trip and wanted to discuss when I got back. Originally the thought was to organize palette around the structure that the design team setup and is represented in the nav, but the more I think about it I feel like the flat structure in elements (as things currently are) is the least ambiguous as a future dev can see exactly what is in palette at a glance wishing having to dig.

jonallured:
We talked this over and decided it wasn't worth it - closing!!

PR #355, opened 2019-03-28
Remove extraneous attributes from Image component

zephraph:
This is a _rough_ but simple implementation that covers _some_ cases of removing styled props from elements. 

Unfortunately, styled-components and styled-systems works in a way that props from styled-system always get forwarded to styled-components... and end up being rendered in the DOM. Our only recourse here is to strip them before they get passed to the final component in the chain.

This doesn't work for native styled components like `styled.div` and the like. [clean-tag](https://github.com/styled-system/extras/tree/master/packages/clean-tag) accomplishes this, but it's deprecated. We can fork it and run our own or do something different.


jonallured:
Nice one! My only feedback is that maybe `src/elements` isn't the best place for `WithoutStyleProps` - what about something like `src/utilities` instead?

packages/palette/src/elements/Image/LazyImage.tsx
```diff
@@ -9,9 +9,10 @@ import {
 } from "styled-system"
 import { color } from "../../helpers/color"
 import { Box, BoxProps } from "../Box"
+import { withoutStyleProps } from "../WithoutStyleProps"
```
damassi: Can the casing of this be consistent 1:1, either all upper-case or (preferrably) lower-case (since it's technically a HOC)?

packages/palette/src/elements/WithoutStyleProps.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react"
```
damassi: Can this file be moved to helpers?

packages/palette/src/elements/Image/LazyImage.tsx
```diff
@@ -8,10 +8,11 @@ import {
   WidthProps,
 } from "styled-system"
 import { color } from "../../helpers/color"
+import { withoutStyleProps } from "../../helpers/withoutStyleProps"
```
damassi: 👍

artsyit:
:rocket: PR was released in v4.5.4 :rocket:

PR #356, opened 2019-03-29
Don't try to debounce if onSelect isn't defined

zephraph:
There was a docs build failure that happened because `onSelect` isn't always passed into radio but debounce requires a function to be passed to it. Gating the check fixes the build.


artsyit:

Built with commit 710363da01b9f4299cc6778061d105e032034991

https://deploy-preview-356--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.5.4 :rocket:

PR #360, opened 2019-03-29
Don't strip width and height props from images

zephraph:
@mzikherman found a bug caused by the width and height props being removed from an image. This readds those props.


artsyit:
:rocket: PR was released in v4.5.6 :rocket:

artsyit:

Built with commit ce767c96e10cb75090a27222a4a5001f15ae39d9

https://deploy-preview-360--artsy-palette.netlify.com

PR #370, opened 2019-04-04
Add a safer way to remove extraneous props

zephraph:
This is a bit of a big-ish PR and it's fairly impactful so please poke holes 🙏 . 

## Context

Last week I implemented #355 which ended up [causing an issue](https://github.com/artsy/palette/pull/360) in staging. Turns out, it also broke avatars in prod. It's just not worked out very well. 

I've created a fork/mild rewrite of @jxnblk's [clean-tag](https://github.com/styled-system/extras/tree/master/packages/clean-tag) library which he's no longer maintaining. I ported it over to ts, removed its sole dependency, and tried to make it cross platform.  

Essentially the library is used to strip props added by styled-system _after_ styled-components has finished with them. 

## Other stuff

I found a [quirk](https://reactjs.org/docs/accessibility.html#wai-aria) with react that requires `aria-*` attributes to work. 

Also, when I added [lazy loading to our images](https://github.com/artsy/palette/pull/334) I think I broke them in emission. As is, I'm thinking the `ResponseImage` should very likely be a web only thing.


artsyit:

Built with commit 6af778bc3e15efc08a30ff27b09657a7ceb6e3b1

https://deploy-preview-370--artsy-palette.netlify.com

packages/palette/src/elements/Tag/__tests__/Tag.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+
+  it("allows the type of element to be overloaded by props", () => {
+    const wrap = shallow(
+      React.createElement(Tag, {
+        is: "section",
+      })
+    )
+    expect(wrap.type()).toBe("section")
+  })
```
ds300: Would also be nice to see a test for `omitFromProps `

packages/palette/src/elements/Image/Image.shared.tsx
```diff
@@ -59,7 +61,7 @@ export interface ResponsiveImageProps
 /**
  * An Image component that responsively resizes within its environment
  */
-export const BaseResponsiveImage = styled.div<ResponsiveImageProps>`
+export const BaseResponsiveImage = styled(Tag.as(View))<ResponsiveImageProps>`
```
ds300: `View` is the default, no?

zephraph: You have a point, heh.

packages/palette/src/elements/Tag/Tag.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Derived from @jxnblk's clean-tag
```
alloy: Add a URL and copy copyright/license

alloy:
I wonder if a whitelist like Emotion’s would be useful and could simplify things? https://github.com/emotion-js/emotion/tree/master/packages/is-prop-valid

zephraph:
@alloy potentially. We could just it to replace the logic of the omit method. Most of the plumbing here would still be required to work with styled-components though.

alloy:
Cool, let’s do it 👍 Definitely happy to see those non-existant attributes go, thanks! 🙏

artsyit:
:rocket: PR was released in v4.7.0 :rocket:

PR #372, opened 2019-04-08
Resolve issue where some images didn't load when lazy loading

zephraph:
We had a weird issue with image lazy loading where some images failed to load on some browsers (chrome mostly). 

It seemed to be related to intersection observers. After some in-depth investigation I discovered that sometimes when an element is observed by an intersection observer but the width and height are 0 that the intersection event may never be fired. This isn't consistent behavior, but definitely happens on chrome. 

In this PR I allowlisted `width` and `height` props to the `LazyLoadImage` component and set the `InnerLazyImage` component to have `100%` width and height. This resolves the intermittent lazy-loading issues.


artsyit:

Built with commit 5a16c8da2bab2fc0e6e70ce1ad149ce783a58401

https://deploy-preview-372--artsy-palette.netlify.com

packages/palette/src/elements/Image/LazyImage.tsx
```diff
@@ -9,9 +9,13 @@ import {
 } from "styled-system"
 import { color } from "../../helpers/color"
 import { Box, BoxProps } from "../Box"
-import { Tag } from "../Tag"
+import { omitProps, Tag } from "../Tag"
 import { BaseImage as Image, ImageProps } from "./Image.shared"
 
+const imagePropsToOmit = omitProps.filter(
+  prop => prop !== "width" && prop !== "height"
```
damassi: Why is `width` and `height` blacklisted?

artsyit:
:rocket: PR was released in v4.7.4 :rocket:

packages/palette/src/elements/Image/LazyImage.tsx
zephraph: Basically it takes all styled-systems props and removes them. So if width and height were specified to be added as a styled-systems thing you probably down want them to flow down as top level props to the child component. In this case I did.

PR #377, opened 2019-04-16
creates a docs-only <ColorComponent> to allow display of separate colors

mdole:
Discussed with @owendodd and Design would like to have the option to not display every color in Palette in the same section, instead picking out a color or group of colors to display.

This is due to a change in the way we're dealing with colors - in [Notion](https://www.notion.so/artsy/Color-a0c24896daf8433d9409aee2146ac267), you can see Primary, Secondary, and Deprecated colors, and these lists will likely grow and change over time.

The docs update in this PR allows greater flexibility by creating a mini-component that takes a color (e.g. `black100`) and outputting a color bar, hex code, and title in the same manner as the old `Colors` page.

After this is merged, @owendodd can create `<ColorComponents />` using the docs admin to match our current colors.

Two things that could be improved about this implementation:
1. hexes are not automatically generated and have to be input separately from the color code. I feel like this could be fixed, but not sure how.
2. This will have to be manually updated every time we change or add to our colors. Would be great to figure out a way to automate that process.

Old page, which always displayed every color:
Screen Shot 2019-04-16 at 1 42 00 PM

New page, which allows display of select colors:
Screen Shot 2019-04-16 at 1 42 03 PM

### DIFF
--- a/packages/palette-docs/content/docs/tokens/Color.mdx
+++ b/packages/palette-docs/content/docs/tokens/Color.mdx
@@ -6,15 +6,6 @@ See [Notion](https://www.notion.so/artsy/Color-a0c24896daf8433d9409aee2146ac267)
 for more detailed spec.
 
 <Box>
-  {Object.entries(themeProps.colors).map(([colorCode, hex], index) => {
-    return (
-      <Box mb={3} key={index}>
-        <Sans size="3" weight="medium">
-          {colorCode}
-        </Sans>
-        <Sans size="3">{hex}</Sans>
-        <Box width="100%" height={5} bg={colorCode} />
-      </Box>
-    )
-  })}
+  <ColorComponent color="black100" hex="#000" />
+  <ColorComponent color="black80" hex="#333" />
 </Box>
--- a/packages/palette-docs/content/docs/tokens/Color.mdx
+++ b/packages/palette-docs/content/docs/tokens/Color.mdx
@@ -6,6 +6,6 @@ See [Notion](https://www.notion.so/artsy/Color-a0c24896daf8433d9409aee2146ac267)
 for more detailed spec.
 
 <Box>
-  <ColorComponent color="black100" hex="#000" />
-  <ColorComponent color="black80" hex="#333" />
+  <ColorComponent color="black100" />
+  <ColorComponent color="purple100" />
 </Box>
--- a/packages/palette-docs/gatsby-config.js
+++ b/packages/palette-docs/gatsby-config.js
@@ -33,10 +33,12 @@ module.exports = {
         // gatsby-mdx. See https://github.com/ChristopherBiscardi/gatsby-mdx/issues/243
         globalScope: `
           import * as Elements from "@artsy/palette"
+          import { ColorComponent } from "components/ColorComponent"
           import { CodeEditor, Playground  } from "components/Playground"
           import { Toggle as Toggler } from 'react-powerplug'
           export default {
             CodeEditor,
+            ColorComponent,
             Playground,
             Toggler,
             ...Elements
--- a/packages/palette-docs/src/components/ColorComponent.jsx
+++ b/packages/palette-docs/src/components/ColorComponent.jsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { Box, Sans } from "@artsy/palette"
+import { Box, Sans, themeProps } from "@artsy/palette"
 
 export const ColorComponent = props => {
   return (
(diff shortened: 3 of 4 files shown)


packages/palette-docs/content/docs/tokens/Color.mdx
```diff
@@ -6,15 +6,6 @@ See [Notion](https://www.notion.so/artsy/Color-a0c24896daf8433d9409aee2146ac267)
...
-        <Sans size="3" weight="medium">
-          {colorCode}
-        </Sans>
-        <Sans size="3">{hex}</Sans>
-        <Box width="100%" height={5} bg={colorCode} />
-      </Box>
-    )
-  })}
+  <ColorComponent color="black100" hex="#000" />
+  <ColorComponent color="black80" hex="#333" />
```
damassi: Why would we list only two colors? 

Shouldn't we match the Zeplin spec so that the docs are useful? (Two black colors also isn't very pretty, compared to before.)

@owendodd  - we can restructure the docs around the the new Primary, Secondary, and Deprecated sections; we don't need to throw everything out.

mdole: thanks @damassi - thought you were out sick today or would've requested your review! yes, this will show all of the colors, I just figured I'd leave it to Owen to update using admin after merging

damassi: We should probably think about this in a more dynamic way -- e.g., grouping things behind the three categories and then iterating over a JSON object (vs a manual list)

mdole: And the intent here definitely isn't to scrap everything; I didn't think we could easily separate out the colors into sections with the previous `map` implementation. But open to ideas!

packages/palette-docs/gatsby-config.js
```diff
@@ -33,10 +33,12 @@ module.exports = {
         // gatsby-mdx. See https://github.com/ChristopherBiscardi/gatsby-mdx/issues/243
         globalScope: `
           import * as Elements from "@artsy/palette"
+          import { ColorComponent } from "components/ColorComponent"
```
damassi: We should keep the global scope clear unless changes are necessary; what is the purpose of this component compared to what was there before?

artsyit:

Built with commit 5a864f439bc8696462c1afe93297436453312682

https://deploy-preview-377--artsy-palette.netlify.com

packages/palette-docs/content/docs/tokens/Color.mdx
damassi:
```
const colors = [
  {
    hex: "FF0000",
    category: "primary"
  },
  ... 
]
```
@owendodd  - would you mind creating a design in Zeplin that we could work from? In the meantime we can probably put this PR on hold.

packages/palette-docs/gatsby-config.js
zephraph: It just encapsulates the structure that was already there. Cleaner if we were going to list out each individually. 

Doesn't strictly need to be a global, but it wasn't immediately clear to us the best way to pull things in.

packages/palette-docs/content/docs/tokens/Color.mdx
owendodd: I'm not too concerned about having group the `primary`, `secondary`, `depreciated` colors more manually here as we will want to add guidelines for the usage of each color alongside the rendered blocks anyways, which seems a little tricky to do while iterating through them all. 

In my mind the idea solution would be be something where I can quickly input a color name (e.g. black30), which would return a block that includes the color name, hex code, and a preview of the color. Then I could easily add text below that outlines the useage guidelines for that color and group accordingly under a 'Primary', 'Secondary' or 'Depreciated' header.

packages/palette-docs/gatsby-config.js
damassi: If it only exists singularly on the `colors.mdx` page (and will only ever exist there) then why break it out into its own component, etc?

packages/palette-docs/content/docs/tokens/Color.mdx
damassi: 👍 that makes sense. 

Y'all will need to update the [Global Components](https://github.com/artsy/palette/blob/master/packages/palette-docs/src/components/GlobalComponents.jsx#L75) file as well.

mdole: ah cool, thanks @damassi! learning my way around the new docs site

mdole:
ok, updated! I'm a little unclear on how the `GlobalStyles` bit works/how to test it - I couldn't figure out where we're currently using those. I think this should be good to go @zephraph if you want to give it a quick look

zephraph:
I think this is good for a first pass. We can always follow up with how we can improve the pattern going forward. 

Great work @mdole and thanks @damassi and @owendodd for the valuable input 🙏

artsyit:
:rocket: PR was released in v4.9.1 :rocket:

mdole:
Thanks @zephraph - it does look like the MDX portion is bugged, and I'm not sure how to fix this. @ds300 or @damassi would one of you have a few minutes to pair on this today? 
Screen Shot 2019-04-17 at 9 33 34 AM

PR #392, opened 2019-04-22
X Axis labels for bar chart

sepans:
Fixes [GALL-1645](https://artsyproduct.atlassian.net/browse/GALL-1645)

Screen Shot 2019-04-24 at 2 50 12 PM

### DIFF
--- a/packages/palette/src/elements/BarChart/Bar.tsx
+++ b/packages/palette/src/elements/BarChart/Bar.tsx
@@ -4,7 +4,6 @@ import { color, media, space } from "../../helpers"
 import { breakpoints } from "../../Theme"
 import { Box } from "../Box"
 import { Flex } from "../Flex"
-import { Sans } from "../Typography"
 import { MousePositionContext } from "./MousePositionContext"
 
 const MAX_BAR_HEIGHT = 80
@@ -164,7 +163,6 @@ const TriangleHighlight = styled.div`
 export const Bar = ({
   heightPercent,
   label,
-  axisLabelX,
   highlightLabel,
   hasEnteredViewport,
   onMeasureHeight,
@@ -215,19 +213,6 @@ export const Bar = ({
         </HighlightLabel>
       )}
       {hover && label && <BarHoverLabel>{label}</BarHoverLabel>}
-      <AxisLabelX color="black60" size="2">
-        {axisLabelX}
-      </AxisLabelX>
     </BarBox>
   )
 }
-
-const AxisLabelX = styled(Sans)`
-  position: absolute;
-  top: calc(100% + ${space(0.5)}px);
-  width: 100%;
-  text-align: center;
-  white-space: nowrap;
-  width: 200%;
-  margin: 0 -50%;
-`
--- a/packages/palette/src/elements/BarChart/Bar.tsx
+++ b/packages/palette/src/elements/BarChart/Bar.tsx
@@ -227,4 +227,7 @@ const AxisLabelX = styled(Sans)`
   top: calc(100% + ${space(0.5)}px);
   width: 100%;
   text-align: center;
+  white-space: nowrap;
+  width: 200%;
+  margin: 0 -50%;
 `
--- a/packages/palette/src/elements/BarChart/Bar.tsx
+++ b/packages/palette/src/elements/BarChart/Bar.tsx
@@ -164,7 +164,7 @@ const TriangleHighlight = styled.div`
 export const Bar = ({
   heightPercent,
   label,
-  xAxisLabel,
+  axisLabelX,
   highlightLabel,
   hasEnteredViewport,
   onMeasureHeight,
@@ -174,7 +174,7 @@ export const Bar = ({
 }: {
   heightPercent: number
   label: React.ReactNode
-  xAxisLabel: React.ReactNode
+  axisLabelX: React.ReactNode
   highlightLabel?: React.ReactNode
   hasEnteredViewport: boolean
   onMeasureHeight?: (height: number) => void
@@ -215,16 +215,16 @@ export const Bar = ({
         </HighlightLabel>
       )}
       {hover && label && <BarHoverLabel>{label}</BarHoverLabel>}
-      <XAxisLabel color="black60" size="2">
-        {xAxisLabel}
-      </XAxisLabel>
+      <AxisLabelX color="black60" size="2">
+        {axisLabelX}
+      </AxisLabelX>
     </BarBox>
   )
 }
 
-const XAxisLabel = styled(Sans)`
+const AxisLabelX = styled(Sans)`
   position: absolute;
-  bottom: -${space(2)}px;
+  top: calc(100% + ${space(0.5)}px);
   width: 100%;
   text-align: center;
 `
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/BarChart/Bar.tsx
```diff
@@ -137,6 +138,7 @@ const HighlightLabel = ({
 export const Bar = ({
   heightPercent,
   label,
+  xAxisLabel,
```
damassi: Maybe `axisLabelX` / `AxisLabelX`?

artsyit:

Built with commit 8aad17a23986ad2e49c28c3fc9f7b8f3f6645e1b

https://deploy-preview-392--artsy-palette.netlify.com

packages/palette/src/elements/BarChart/Bar.tsx
```diff
@@ -185,6 +188,16 @@ export const Bar = ({
...
+      <XAxisLabel color="black60" size="2">
+        {xAxisLabel}
+      </XAxisLabel>
     </BarBox>
   )
 }
+
+const XAxisLabel = styled(Sans)`
+  position: absolute;
+  bottom: -${space(2)}px;
```
sepans: This part is a bit too much a hack and bothers me. It is because the label is inside the `<BarBox>` to have the horizontal alignment right but then had to do a negative margin bottom here (which is dependent on the height of the label).

packages/palette/src/elements/BarChart/BarChart.tsx
```diff
@@ -115,13 +116,17 @@ export const BarChart = ({ bars, minLabel, maxLabel }: BarChartProps) => {
...
+            (
+              { value, label, xAxisLabel, highlightLabel, onClick, onHover },
+              index
+            ) => {
               const heightPercent = (100 / maxValue) * value
               return (
                 <Bar
                   key={index}
                   heightPercent={heightPercent}
                   label={coerceLabel(label)}
```
ds300: in hindsight this probably should have been called tooltip 😞

packages/palette/src/elements/BarChart/Bar.tsx
ds300: you could do `top: calc(100% + ${marginBetweenLabelAndBaseLine});`

packages/palette/src/elements/BarChart/Bar.tsx
```diff
@@ -185,6 +188,16 @@ export const Bar = ({
...
+      </AxisLabelX>
     </BarBox>
   )
 }
+
+const AxisLabelX = styled(Sans)`
+  position: absolute;
+  top: calc(100% + ${space(0.5)}px);
+  width: 100%;
+  text-align: center;
```
ds300: I think to be more flexible it would make sense to center this whole box instead of just the text. That would mean the box is not constrained to be the same width as the bar, and then the text will still be centered even if the text is wider than the bar.

That is, unless you want to use word wrapping in these labels? Then you'd need an explicit width like how it's done now.

sepans: Yes good catch looking at the designs we don't want the labels to be constrained by the bar width. cc: @owendodd and @samchieng:

Screen Shot 2019-04-25 at 10 31 03 AM

sepans: pfffff, I thought it was simple but haven't figured it out yet. Is it possible to center align the label without removing `position: absolute;`? If not it requires a good amount of change to the way bars are positioned I guess.

ds300: try
```
left: 50%;
transform: translateX(-50%);
```

sepans: I got that with https://github.com/artsy/palette/pull/392/commits/0ff165567c44cc0a9f0c34e4561e96b3d9e34890#diff-2e24d9a0ed8b8746d5fa5d1333da123dR231:

Screen Shot 2019-04-25 at 12 05 16 PM

The only tricky thing was that I had to add https://github.com/artsy/palette/pull/392/commits/0ff165567c44cc0a9f0c34e4561e96b3d9e34890#diff-710f36809acea9f66cf6ae101cf56b66R13 to remove the extra padding on the right side of the chart created but `width: 200%;`

Maybe I was missing something but by some reason 👆 didn't quite work:

Screen Shot 2019-04-25 at 12 23 00 PM

ds300: Hmm I tried the left + transform trick and it looked fine.

Also while playing around I noticed that these labels aren't visible for me because of overflow issues. They don't add to the height of the bar chart, but they probably should.

It might be best to rethink how we're doing this. I'm imagining a parallel row of equal-sized flex containers underneath the row of bars which center their children using flexbox and align items using `baseline`, does that make sense to you?

```
Flex row ----->

+-------------+
|         +-+ |
| +-+     | | |
| | |     | | |
| | | +-+ | | |
| | | | | | | |
| | | | | | | |
| | | | | | | |
| +-+ +-+ +-+ |
+-------------+
+-------------+
| +-+ +-+ +-+ |
| +-+ +-+ +-+ |
+-------------+

Flex row ----->

```

ds300: actually just tried that out and don't think it'd work because flexbox doesn't center overflowing content 😞

ds300:
@sepans Another issue I see with this is that the x-axis labels are mutually exclusive with the existing min+max label props. It would be cool to capture that in the api design. Could do it by deprecating `minLabel` and `maxLabel` and creating a new polymorphic `xAxisLabels` prop that can capture both situations and any future requirements.

e.g
image

sepans:
@ds300 👆 is a great idea and I went ahead and changed it until  I realized in order to have x-axis labels for some bars an array like `[undefined, '200', undefined, '100'] needs to be passed to the `xAxisLabels`.

Also thinking it over the mutual exclusiveness of `min/maxLabel` and `xAxisLabels` is just a visual constraint and not semantic (e.g. hypothetically if in later design iterations the position of `min/maxLabel` changes they are no longer mutually exclusive) so I won't be super worried if it is not enforced in the component API.

ds300:
Ah yeah that's a great point.

Did you have time to look into the vertical height issue?

sepans:
I tried a couple of different things but haven't found a solution for that yet. 😥

sepans:
@ds300 I tried another hybrid approach which is not perfect but better. I explained it in the commit message and here is the screenshot:

Screen Shot 2019-05-15 at 11 50 39 AM

let me know what you think 🙏

ds300:
Thanks for your patience on this :pray: That looks like a good compromise to get stuff done and move on for now.

I noticed some unintended scroll behaviour.

nested-scroll

I looked into this for a few minutes but couldn't figure out what the issue was. Do you mind having a deeper look?

sepans:
Thanks @ds300 for catching that. It was actually caused by the large font size of the `December 30` and `April` (that `min-height` hack only takes care of the labels up to a certain line height). Since large labels are not in out design system I just decreased the font size to fix it. 

And also noticed an extra `overflow: hidden` tag that that I had added and was not really needed and removed it.

artsyit:
:rocket: PR was released in v4.12.0 :rocket:

PR #393, opened 2019-04-23
only apply cursor pointer to bar if it has onclick

lilyfromseattle:
Made `cursor: pointer` conditional so that it only applies to bars with onClick functions passed in. This is related to this story: https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&modal=detail&selectedIssue=PURCHASE-1035

### DIFF
--- a/packages/palette/src/elements/BarChart/Bar.tsx
+++ b/packages/palette/src/elements/BarChart/Bar.tsx
@@ -14,6 +14,7 @@ const LABEL_OFFSET = 10
 
 interface BarBoxProps {
   isHighlighted?: boolean
+  hasOnClick?: boolean
 }
 
 // the actual visible bit of the bar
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit b7a272a5c3d80871502da60c375558ce6b521b9b

https://deploy-preview-393--artsy-palette.netlify.com

packages/palette/src/elements/BarChart/Bar.tsx
```diff
@@ -29,6 +30,7 @@ export const BarBox = styled(Box)`
   }
   flex: 1;
   cursor: pointer;
+  cursor: ${(props: BarBoxProps) => (props.hasOnClick ? "pointer" : "default")};
```
ds300: You can delete the line above

packages/palette/src/elements/BarChart/Bar.tsx
```diff
@@ -14,6 +14,7 @@ const LABEL_OFFSET = 10
 
 interface BarBoxProps {
   isHighlighted?: boolean
+  hasOnClick?: boolean
```
ds300: Do we need this extra prop? Maybe instead we could check whether an 'onClick' callback was supplied?

artsyit:
:rocket: PR was released in v4.10.4 :rocket:

PR #397, opened 2019-04-25
[Spike] on LineChart and DonutCharts #DontMerge 

sepans:
This is a minimum spike to evaluate the dependencies we need to add to `LineChart` and `DonutChart` to decide if we want to add them to Palette or keep them in another repo, or use a 3rd party charting library.

Things that are in this spike:
- Find the minimum dependencies to draw and animate line and donut chart without creating a lot of complexity. (`d3-shape` and `d3-interpolate`)

Things that are not included in the spike:
- Many of the features such as tooltip and the comonent API are going to be the same as `BarChart`.
- The new version of `react-spring` has a better API but left it out for now.
- Making the chart size responsive (and getting rid of those height/width/margin constants)

In line chart the component interpolates from an array of zeros to data values and generates the svg path for those interpolated value. One possible extension is to keep those values in component state so by changing data passed to the component it also animates to the new values.

In donut chart it interpolates the `endAngle` of the 'pie' from 0 to 2 * PI so the donut shape animates. It is also possible to keep data values in component state to animate the donut to new values when they change similar to line chart.

The use of `react-spring` is very minimal so it can be swapped with any other library that can provide a single seed (changing from 0 to 1) for interpolation.

`d3-interpolate` is also doing super basic mapping so that one could also be replaced by another library or just a function. (the only nice thing `interpolateArray` does is that it is smarter interpolating arrays of different size)


artsyit:

Built with commit 0b35155853b0980c413d668821f23b20a62aeacc

https://deploy-preview-397--artsy-palette.netlify.com

packages/palette-docs/content/docs/elements/data-visualization/DonutChart.mdx
```diff
@@ -0,0 +1,17 @@
+---
+name: DonutChart
+---
+
+<Playground showEditor={true}>
```
damassi: `showEditor` is true by default

damassi:
Seems simple enough, and looks great in the preview 👍

ds300:
Yeah, this is all really easy to follow, great stuff @sepans! What's the bundle size increase?

sepans:
closing in favor of https://github.com/artsy/palette/pull/464

PR #452, opened 2019-05-01
Add storybook for Join component and a couple of more tests.

oxaudo:
The original idea was to skip rendering blank components if those are passed as children to Join. But that is very hard to do without pre-rendering the components first. I've abandoned that idea but added a storybook and a couple of tests in a process for Join component. So those could stay :)

The Storybook entry:
Screen Shot 2019-05-01 at 4 17 06 PM

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -29,7 +29,7 @@
     "watch:emission": "yarn clean:emission && concurrently --raw --kill-others 'yarn compile:emission -w' 'yarn type-declarations -w --outDir ../../../emission/node_modules/@artsy/palette/dist'",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "visual-test": "chromatic test --exit-zero-on-changes"
+    "visual-test": "chromatic test --exit-zero-on-changes --app-code n02zjqmdqq"
   },
   "repository": {
     "type": "git",
--- a/packages/palette/src/elements/Join/Join.tsx
+++ b/packages/palette/src/elements/Join/Join.tsx
@@ -26,15 +26,18 @@ interface JoinProps {
  */
 export const Join: SFC<JoinProps> = ({ separator, children }) => {
   const childArray = React.Children.toArray(children) as any
+  const filteredChildren = childArray.filter(
+    child => child && child.props && child.props.children
+  )
 
-  return childArray.reduce((acc, curr, currentIndex) => {
+  return filteredChildren.reduce((acc, curr, currentIndex) => {
     acc.push(
       React.cloneElement(curr as React.ReactElement<any>, {
         key: `join-${currentIndex}`,
       })
     )
 
-    if (currentIndex !== childArray.length - 1) {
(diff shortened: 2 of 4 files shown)


artsyit:

Built with commit cd9546e2cf122026b5993be1139aeaf2ff4b71f0

https://deploy-preview-452--artsy-palette.netlify.com

packages/palette/src/elements/Join/Join.tsx
```diff
@@ -26,15 +26,18 @@ interface JoinProps {
  */
 export const Join: SFC<JoinProps> = ({ separator, children }) => {
   const childArray = React.Children.toArray(children) as any
+  const filteredChildren = childArray.filter(
+    child => child && child.props && child.props.children
```
damassi: What if it's just `<FooComponent />` without props or children?

oxaudo: So i believe that is exactly the thing that we want to skip. If the component does not render anything whatsoever - it should be skipped. Like in out case <TestComponent/> returns `null`.

If the component renders anything - those elements considered `children` (or at least what it seems like from my testing). Similar with `<div/>` - blank one will be skipped, but once it has any content - it'll be considered with children.

damassi: I wonder if you can assert against `child.type` vs checking against props and if that might be safer? 

Either way, an additional test case using a flat component (`<FooComponent />`) that _does_ have content will settle it.

oxaudo: @damassi - I was wrong... It does not work if `<FooComponent />` just renders but does not have any params or children :(

Discussed this with @zephraph in person a bit - and even though we agree that the idea is nice here - it seems to be super tricky to implement without first rendering it.

I think I'll close this for now. Wondering though if it's helpful to just add the story that illustrates current functionality.

damassi: Yup, we should keep the story since that now gets piped to Chromatic 👍

oxaudo:
Ok @damassi - I've removed the actual change but kept the spec and the story that I've added. 

Will merge on green unless you have some objections here.

oxaudo:
@zephraph - Chromatic stuff is very cool! That it lets you see all the components rendered in one page:)

packages/palette/package.json
```diff
@@ -29,7 +29,7 @@
     "watch:emission": "yarn clean:emission && concurrently --raw --kill-others 'yarn compile:emission -w' 'yarn type-declarations -w --outDir ../../../emission/node_modules/@artsy/palette/dist'",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "visual-test": "chromatic test --exit-zero-on-changes"
+    "visual-test": "chromatic test --exit-zero-on-changes --app-code n02zjqmdqq"
```
oxaudo: This was added with @zephraph blessing to allow chromatic to run properly on PR from forks.

artsyit:
:rocket: PR was released in v4.11.4 :rocket:

PR #464, opened 2019-05-17
Add line chart

sepans:
Addresses https://github.com/artsy/palette/issues/359

Paired with @iskounen 

[Preview](https://deploy-preview-464--artsy-palette.netlify.com/elements/data-visualization/linechart/)

linechart

Added package sizes:

d3-shape: [7.7k](https://bundlephobia.com/result?p=d3-shape@1.3.5)
d3-interpolate: [6.4k](https://bundlephobia.com/result?p=d3-interpolate@1.3.2)
~lodash.maxBy: [4.4k](https://bundlephobia.com/result?p=lodash.maxby@4.6.0)~ (lodash was already there)

todo list:
- [x] Add tooltip
- [x] Move and rename components shared between BarChart and LineChart (where? suggestions @ds300?)
- ~[ ] update react-spring version~ (updating `collapse.ios.tsx` wasn't straight forward. Will do it in another pass)
- [x] more test coverage
- [x] mobile
- ~[ ] `onHover` and `onClick`?~ (not in the spec will add it when needed)
- [x] move svg section of LineChart to a separate component


artsyit:

Built with commit 9e6f10760faa33ad20931bc88e1814c9eb3ceb39

https://deploy-preview-464--artsy-palette.netlify.com

ds300:
> Move and rename components shared between BarChart and LineChart (where? suggestions @ds300?)

Maybe `src/helpers/data-vis/`? @damassi do you have opinions on this?

sepans:
👍 I will move the helper functions to `src/helpers/data-vis/`. What about the `BarHoverLabel`? Should it also go to the same folder or since it is a component it is better to have it in `src/elements/DataVis`? or create `src/elements/shared` or something?

I was also thinking about renaming `BarHoverLabel` to `ChartHoverTooltip` and rename all mentions of `label` to `tooltip` where it refers to that hover thing. Does that sound good?

damassi:
@sepans, @ds300 - why not `src/elements/data-vis/utils`? Keep things nice and scoped.

packages/palette/src/elements/LineChart/Line.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react"
+import { color } from "../../helpers"
+
+/**
+ * Line is the component responsible for rendering a line that goes through all of the data points
+ * in a line chart.
+ */
+export const Line = ({ d }: { d: string }) => {
+  return <path fill="none" stroke={color("black10")} d={d} />
```
damassi: Not sure if `d` follows prior art but 1-character var names don't convey much information, at least passed as props.

packages/palette/src/elements/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  useEffect(() => {
+    window.addEventListener("resize", setContainerWidth)
+
+    return function cleanup() {
+      window.removeEventListener("resize", setContainerWidth)
+    }
+  }, [])
+
+  const hasEnteredViewport = useHasEnteredViewport(wrapperRef)
```
damassi: The width / height code that's above this line can be shared with the other chart, yah?

packages/palette/src/helpers/visualizationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+      }
+    }
+    window.addEventListener("scroll", handleScroll)
+    window.dispatchEvent(new Event("scroll"))
+    return () => {
+      window.removeEventListener("scroll", handleScroll)
+    }
+  })
+  return hasEntered
+}
```
damassi: `useHasEnteredViewport` can be its own file, same with `max` (though the name can probably be expanded a bit). If in the future we have more visualization helpers, we can create a folder of that name. 

(Utils, if they're all in their own files, allow devs to glance at a folder to see whats available, versus having to dig inside of files; it helps keep stuff at the surface.)

packages/palette/src/elements/LineChart/LineChart.tsx
sepans: Maybe for the donut chart if we end up adding it to palette. `BarChart` doesn't need that because it doesn't have `SVG`.

packages/palette/src/elements/DataVis/ChartHoverTooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { media } from "../../helpers/media"
+import { Flex } from "../Flex"
+import { MousePositionContext } from "./MousePositionContext"
+
+const LABEL_OFFSET = 10
+
+/**
+ * Tooltip for bar and line charts
+ */
+export const ChartHoverTooltip = ({ children }) => {
```
damassi:
```ts
export const ChartHoverTooltip: React.FC< ChartHoverTooltipProps> = ...
```

packages/palette/src/elements/DataVis/utils/maxInArray.ts
```diff
@@ -0,0 +1,14 @@
...
+ * specifies which attribute of object to be used for comparison. default is a unity function
+ */
+export const maxInArray = (
+  values: any[],
+  accessor: (item: any) => number = item => item
+): number =>
+  values.reduce((currentMax: number, item: any): number => {
+    const value: number = accessor(item)
+    return value > currentMax ? value : currentMax
+  }, -Infinity) as number
```
damassi: Could https://lodash.com/docs/4.17.11#max or https://lodash.com/docs/4.17.11#maxBy be used instead?

packages/palette/src/elements/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+const margin = space(2)
+const DEFAULT_HEIGHT = 87
+
+/**
+ * LineChart is a component that displays some data points connected by lines.
+ * Useful for visualizing a time series, etc.
+ */
+export const LineChart = ({ points, chartHeight }: LineChartProps) => {
```
damassi:
```ts
export const LineChart: React.FC<LineChartProps> = ...
```

packages/palette/src/elements/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,151 @@
...
+    </ProvideMousePosition>
+  )
+}
+
+interface AxisContainerProps {
+  first: boolean
+  last: boolean
+}
+
+const HoverHandler = ({ children, index, hoverIndex, setHoverIndex }) => {
```
damassi: `React.FC<...>`

packages/palette/src/elements/LineChart/LineChartSVG.tsx
```diff
@@ -0,0 +1,86 @@
...
+  margin: number
+  points: PointDescriptor[]
+  hoverIndex: number
+  hasEnteredViewport: boolean
+}
+
+/**
+ * Component rendering the SVG part of the line chart (e.g line and points)
+ */
+export const LineChartSVG = ({
```
damassi: `React.FC<...>`

packages/palette/src/elements/DataVis/utils/maxInArray.ts
sepans: yes it is exactly the same as `maxBy`. d3 also has similar think but wasn't sure if we want to add dependency vs adding a 4 line function.

damassi: We're already using lodash, so that's preferred since we don't need to maintain it.

packages/palette/src/elements/DataVis/ChartHoverTooltip.tsx
sepans: hmmm I see these different notations of defining component props but don't know what the difference is 🤔

damassi: The version currently being used is just a normal function with an interface. By assigning `React.FC<Props>` we confer additional type information to the component -- that it is a React component, that it is a Functional component, and so on.

damassi:
Chart is looking stellar @sepans 💯

packages/palette/src/helpers/visualizationHelpers.tsx
```diff
@@ -0,0 +1,39 @@
...
+        setHasEntered(true)
+        window.removeEventListener("scroll", handleScroll)
+      }
+    }
+    window.addEventListener("scroll", handleScroll)
+    window.dispatchEvent(new Event("scroll"))
+    return () => {
+      window.removeEventListener("scroll", handleScroll)
+    }
+  })
```
sepans: @ds300 is this effect gets triggered on every render and adds a new event listener for scroll? Should this be `}, [])` (added empty array as second param to `useEffect`) so it only adds eventListener on mount?

packages/palette/src/elements/DataVis/ChartHoverTooltip.tsx
sepans: got it got it! it is type definition for `ChartHoverTooltip` itself and not the function params!

packages/palette/src/elements/BarChart/BarChart.tsx
```diff
@@ -3,7 +3,7 @@ import styled from "styled-components"
 import { color, space } from "../../helpers"
 import { ChartTooltipProps, coerceTooltip } from "../DataVis/ChartTooltip"
 import { ProvideMousePosition } from "../DataVis/MousePositionContext"
-import { maxInArray } from "../DataVis/utils/maxInArray"
+import maxBy from "lodash.maxBy"
```
damassi: @sepans - in Reaction and Force we use `babel-plugin-lodash` to automatically cherry-pick our functions; so rather than add a new dependency we just do 
```js
import { maxBy } from 'lodash'
```
If you can also add https://github.com/artsy/force/blob/master/.babelrc#L26 to palette that would be nice (and would recommend adding it to volt as well.)

packages/palette/src/elements/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * LineChart is a component that displays some data points connected by lines.
+ * Useful for visualizing a time series, etc.
+ */
+export const LineChart: React.FC<LineChartProps> = ({
+  points,
+  chartHeight,
+}: LineChartProps) => {
+  const [width, setWidth] = useState(0)
+  const [height, setHeight] = useState(DEFAULT_HEIGHT)
```
ds300: looks like you don't need `height` as state here? Can just use `chartHeight = DEFAULT_HEIGHT` in the props destructuring above.

packages/palette/src/elements/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const hasEnteredViewport = useHasEnteredViewport(wrapperRef)
+
+  return (
+    <ProvideMousePosition>
+      <Flex flexDirection="column" ref={wrapperRef as any} width="100%">
+        {width && (
+          <>
+            <Flex height={height}>
+              {points.map((point, i) => (
+                <HoverHandler
```
ds300: tidy solution to this problem :+1:

packages/palette/src/elements/LineChart/LineChart.tsx
sepans: 🙏 nice catch!

sepans:
Thanks @ds300 made the changes. The fade-in was a great suggestion. 🙌

ds300:
Sweeet I'm happy to merge then :sparkles: Terrific work! ❤️

artsyit:
:rocket: PR was released in v4.12.3 :rocket:

PR #465, opened 2019-05-20
Add some help for users who might be confused why their `Radio` isn't selected.

pepopowitz:
In https://github.com/artsy/reaction/pull/2439, I found a `Radio` that was setting `selected` based on a prop, but it "wasn't working," because it was wrapped in a `RadioGroup`. In that scenario, `RadioGroup`.`defaultValue` is the correct way to mark the individual `Radio` items as selected.

If you'd like different verbiage, please suggest it!

### DIFF
--- a/packages/palette/src/elements/Radio/Radio.tsx
+++ b/packages/palette/src/elements/Radio/Radio.tsx
@@ -19,7 +19,7 @@ import {
 export interface RadioProps extends FlexProps {
   /** Disable interactions */
   disabled?: boolean
-  /** Select the button on render */
+  /** Select the button on render. If the Radio is inside a RadioGroup, use RadioGroup.defaultValue instead. */
   selected?: boolean
   /** Show hover state on render */
   hover?: boolean


artsyit:

Built with commit c2cab414c7065ebe06e4ba1d2a6f10724c57cbb9

https://deploy-preview-465--artsy-palette.netlify.com

packages/palette/src/elements/Radio/Radio.tsx
```diff
@@ -19,7 +19,7 @@ import {
 export interface RadioProps extends FlexProps {
   /** Disable interactions */
   disabled?: boolean
-  /** Select the button on render */
+  /** Select the button on render. If the Radio is inside a RadioGroup, use RadioGroup.defaultValue instead. */
```
damassi: Can this be added to palette docs as an example?

pepopowitz: There are examples for `RadioGroup`, and most of them show the use of `RadioGroup`.`defaultValue` instead of `Radio`.`selected`.

https://palette.artsy.net/elements/inputs/radiogroup/

artsyit:
:rocket: PR was released in v4.12.2 :rocket:

PR #467, opened 2019-05-22
Add a MiniSelect as a select component

oxaudo:
For now I've created this MiniSelect as an additional Select component. The idea is that once it's merged we can go ahead and update all the uses of SmallSelect with MiniSelect and then just kill SmallSelect option here.

The looks:
Screen Shot 2019-05-22 at 3 31 36 PM

Screen Shot 2019-05-22 at 3 31 20 PM

Screen Shot 2019-05-22 at 3 31 30 PM

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -199,6 +199,7 @@ const MiniSelectContainer = styled.div<SelectProps>`
     background-color: ${color("black10")};
     border-radius: 2px;
     font-size: ${themeGet("typeSizes.sans.2.fontSize")}px;
+    font-weight: 500;
     line-height: ${themeGet("typeSizes.sans.2.lineHeight")}px;
     padding: ${space(0.5)}px ${space(1) + carretSize * 4}px ${space(0.5)}px
       ${space(1)}px;
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -19,6 +19,7 @@ export interface SelectProps extends PositionProps, SpaceProps {
   selected?: string
   disabled?: boolean
   error?: string
+  title?: string
   onSelect?: (value) => void
 }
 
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/Select/Select.tsx
```diff
@@ -71,6 +72,36 @@ export const SmallSelect: SFC<SelectProps> = props => {
...
+          {props.options.map(({ value, text }) => (
+            <option value={value} key={value}>
+              {text}
+            </option>
+          ))}
+        </select>
+      </label>
+    </MiniSelectContainer>
+  )
+}
```
damassi: Can we add a test for this?

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -159,3 +190,34 @@ const SmallSelectContainer = styled.div<SelectProps>`
...
+  }
+
+  &::after {
+    ${caretArrow};
+    content: "";
+    cursor: pointer;
+    margin-left: -${space(1) + carretSize * 2}px;
+    pointer-events: none;
+    position: absolute;
+    top: ${5 + space(0.5)}px;
```
oxaudo: This 5 is a bit of a magic number here. I basically positioned the carret properly without any padding and that is what 5 came from and then added the padding. Not sure what is the more logical way to get to it here.

oxaudo:
Spec here: https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5b6b074edff31c5af3f1f2f1

packages/palette/src/elements/Select/Select.tsx
oxaudo: I was going to but the test for the Select is non existent. Would love to maybe pair with somebody who knows what the plan for this test was.

oxaudo: The Select also missing .ios file, not sure why?

damassi: Likely because this was moved over from Reaction and was never updated

oxaudo: Per discussion with @damassi - will add a separate test file for MiniSelect that tests the props handling and selection function.

packages/palette/src/elements/Select/Select.tsx
damassi: I believe 
```js
5 + space(0.5) === space(1)
```

oxaudo: Absolutely true, but the idea here was that it's some magical number 5 + whatever is currently set as top padding for this element. But since the padding is just hardcoded above - maybe that is an unnecessary complication here.

artsyit:

Built with commit 950f8bdd53fc66f712eedf76648078422ed82bb3

https://deploy-preview-467--artsy-palette.netlify.com

oxaudo:
@damassi - I've added a spec and accepted chromatic changes. Good to merge?

oxaudo:
Merging this. Let me know if anybody has more comments/suggestions.

artsyit:
:rocket: PR was released in v4.12.4 :rocket:

PR #475, opened 2019-05-28
rename MiniSelect to SelectSmall and create separate entry for it in docs

oxaudo:
Current view of `SelectSmall`:
Screen Shot 2019-05-28 at 2 49 04 PM

Did not do anything to `LargeSelect` for now. I think this is a separate refactor/upgrade.
cc @owendodd, @jeffreytr

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -44,6 +44,7 @@ export const LargeSelect: SFC<SelectProps> = props => {
   )
 }
 
+// TODO: Remove this SmallSelect below once all the clients use SelectSmall
 /**
  * A small drop-down select menu
  */
@@ -73,11 +74,11 @@ export const SmallSelect: SFC<SelectProps> = props => {
 }
 
 /**
- * A mini drop-down select menu
+ * A small version of drop-down select menu
  */
-export const MiniSelect: SFC<SelectProps> = props => {
+export const SelectSmall: SFC<SelectProps> = props => {
   return (
-    <MiniSelectContainer {...props}>
+    <SelectSmallContainer {...props}>
       <label>
         {props.title && (
           <Sans size="2" display="inline" mr={0.5}>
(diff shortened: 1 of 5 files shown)


oxaudo:
I added a `WIP` for all those select component sections since there is an active discussion how they are going to look. And I think Select in docs should really be `SelectLarge` but that will probably come with this Select rewrite...

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -73,11 +73,11 @@ export const SmallSelect: SFC<SelectProps> = props => {
 }
 
 /**
- * A mini drop-down select menu
+ * A small version of drop-down select menu
  */
-export const MiniSelect: SFC<SelectProps> = props => {
+export const SelectSmall: SFC<SelectProps> = props => {
```
damassi: @oxaudo - can you add a comment about deprecating the `SmallSelect` component just up above? It seems a bit confusing having these things mixed together. (Also, maybe `SelectSmall` can be moved into its own file.)

owendodd:
👌 Thanks @oxaudo

artsyit:

Built with commit 3ad0911f14c9863b6d131b4cd44d6148d1f2bb6a

https://deploy-preview-475--artsy-palette.netlify.com

packages/palette/src/elements/Select/Select.tsx
oxaudo: Totally. Left a comment. And I do want to work on replacing this SmallSelect next and will kill it here once that is done.

I think it might be ok to keep the implementation in the same file since those are reusing the props interface and couple of other css definitions.

artsyit:

Built with commit 2391a15f3979b2c3aca3d91089bfc6002374320d

https://deploy-preview-475--artsy-palette.netlify.com

oxaudo:
Xm... Not sure what is going with Circle here:(

zephraph:
CircleCI is fine here. For whatever reason one of the steps didn't report its status, but it did successfully complete. 

You're good to merge 👍

artsyit:
:rocket: PR was released in v4.12.6 :rocket:

PR #478, opened 2019-05-29
remove unused SmallSelect and remove margin/padding on label for SelectSmall

oxaudo:
I did a global search in our repos on Github and seems like the only place SmallSelect was used is the reaction and the removal was merged here: https://github.com/artsy/reaction/pull/2475. So killing SmallSelect here.

Also added explicit font definition from Theme for SmallSelect. And made sure that label in this select does not include padding/margin. In Volt the global value for label was sneaking in here:(

### DIFF
--- a/packages/palette-docs/content/docs/elements/inputs/Select.mdx
+++ b/packages/palette-docs/content/docs/elements/inputs/Select.mdx
@@ -65,20 +65,3 @@ of user-benefitting product strategy.
     }}
   </Value>
 </Playground>
-
-### Small select
-
-<Playground>
-  <SmallSelect
-    options={[
-      {
-        text: "Price",
-        value: "price",
-      },
-      {
-        text: "Estimate",
-        value: "estimate",
-      },
-    ]}
-  />
-</Playground>
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -44,35 +44,6 @@ export const LargeSelect: SFC<SelectProps> = props => {
   )
 }
 
-// TODO: Remove this SmallSelect below once all the clients use SelectSmall
-/**
- * A small drop-down select menu
- */
-export const SmallSelect: SFC<SelectProps> = props => {
-  return (
-    <SmallSelectContainer {...props}>
-      <label>
-        <Sans size="2" display="inline" mr={0.5}>
-          Sort:
-        </Sans>
-
-        <select
-          value={props.selected}
-          onChange={event =>
-            props.onSelect && props.onSelect(event.target.value)
-          }
-        >
-          {props.options.map(({ value, text }) => (
-            <option value={value} key={value}>
-              {text}
-            </option>
-          ))}
-        </select>
-      </label>
-    </SmallSelectContainer>
-  )
-}
-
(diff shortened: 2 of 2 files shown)


packages/palette/src/elements/Select/Select.tsx
```diff
@@ -195,12 +195,17 @@ const SmallSelectContainer = styled.div<SelectProps>`
...
+  label {
+    padding: 0;
+    margin: 0;
+  }
+
   select {
     ${hideDefaultSkin};
     background-color: ${color("black10")};
     border-radius: 2px;
+    ${themeGet("fontFamily.sans.medium")};
```
oxaudo: Strangely this `themeGet` here seems to include `font-family: ` and not just the value for font. Not sure why?

damassi: Can you elaborate a bit?

packages/palette-docs/content/docs/elements/inputs/Select.mdx
```diff
@@ -65,20 +65,3 @@ of user-benefitting product strategy.
...
-        text: "Price",
-        value: "price",
-      },
-      {
-        text: "Estimate",
-        value: "estimate",
-      },
-    ]}
-  />
-</Playground>
```
damassi: Is it possible to remove the corresponding `SmallSelect` code? (Need to check apps to see where it's used first, though.)

oxaudo: Oh.. I thought I did:( Maybe somehow did not commit.

artsyit:

Built with commit 122b1ecb571d3033aa64f5783e42cfb3d9b43867

https://deploy-preview-478--artsy-palette.netlify.com

packages/palette-docs/content/docs/elements/inputs/Select.mdx
damassi: Just as long as it doesn't break any consuming apps 👍- Will want to check out reaction, volt, positron

packages/palette/src/elements/Select/Select.tsx
oxaudo: I was expecting to write something like `font-family: ${themeGet("fontFamily.sans.medium")};` but that did not really work and inspecting stuff I noticed that it did not work because the output html looked like `font-family: font-family:Unica77LLWebMedium,'Helvetica Neue',Helvetica,Arial,sans-serif;` so I removed the original `font-family:` and things look right now. It is confusing to me though.

packages/palette-docs/content/docs/elements/inputs/Select.mdx
oxaudo: Yep. Reaction pr that you've merged earlier was addressing the change. And that is the only usage I found.

packages/palette/src/elements/Select/Select.tsx
damassi: That's just a way to quickly get at the main Theme.tsx object, so not too sure.

oxaudo:
Merging!

artsyit:
:rocket: PR was released in v4.12.9 :rocket:

PR #489, opened 2019-06-03
Add icons necessary for the iOS artwork view

sweir27:
This PR adds icons that can be consumed by emission via the `react-native-svg` library, which is what we're already using within emission.

I can follow-up by updating all of the remaining icons to have a proper iOS counterpart, but wanted to keep this targeted for now.

Confirmed that it works via an emission link:
image

### DIFF
--- a/packages/palette/src/svgs/EyeOpenedIcon.ios.tsx
+++ b/packages/palette/src/svgs/EyeOpenedIcon.ios.tsx
@@ -1,0 +1,16 @@
+import React from "react"
+import { G, Path } from "react-native-svg"
+import { color } from "../helpers"
+import { Icon, IconProps } from "./Icon"
+
+/** EyeOpenedIcon */
+export const EyeOpenedIcon: React.SFC<IconProps> = props => {
+  return (
+    <Icon {...props} viewBox="0 0 18 18">
+      <G fill={color(props.fill)} fillRule="nonzero">
+        <Path d="M2.249 8.462a.56.56 0 0 0 .016.601C4.172 11.9 6.406 13.286 9 13.286c2.594 0 4.828-1.387 6.736-4.225a.56.56 0 0 0 .017-.6C14.038 5.56 11.808 4.144 9.01 4.144S3.975 5.56 2.25 8.463zm-1.011-.574C3.163 4.652 5.77 3 9.01 3c3.241 0 5.843 1.653 7.755 4.89.331.561.31 1.257-.053 1.798-2.109 3.139-4.69 4.74-7.713 4.74-3.021 0-5.602-1.6-7.711-4.738a1.681 1.681 0 0 1-.05-1.802z" />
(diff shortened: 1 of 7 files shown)


artsyit:

Built with commit f379f28e04f84193f79f44ffa4d69ecdc894e15e

https://deploy-preview-489--artsy-palette.netlify.com

packages/palette/src/svgs/EyeOpenedIcon.ios.tsx
```diff
@@ -0,0 +1,16 @@
...
+export const EyeOpenedIcon: React.SFC<IconProps> = props => {
+  return (
+    <Icon {...props} viewBox="0 0 18 18">
+      <G fill={color(props.fill)} fillRule="nonzero">
+        <Path d="M2.249 8.462a.56.56 0 0 0 .016.601C4.172 11.9 6.406 13.286 9 13.286c2.594 0 4.828-1.387 6.736-4.225a.56.56 0 0 0 .017-.6C14.038 5.56 11.808 4.144 9.01 4.144S3.975 5.56 2.25 8.463zm-1.011-.574C3.163 4.652 5.77 3 9.01 3c3.241 0 5.843 1.653 7.755 4.89.331.561.31 1.257-.053 1.798-2.109 3.139-4.69 4.74-7.713 4.74-3.021 0-5.602-1.6-7.711-4.738a1.681 1.681 0 0 1-.05-1.802z" />
+        <Path d="M9 11.571a2.857 2.857 0 1 1 0-5.714 2.857 2.857 0 0 1 0 5.714zm0-1.142A1.714 1.714 0 1 0 9 7a1.714 1.714 0 0 0 0 3.429z" />
+      </G>
+    </Icon>
+  )
+}
```
damassi: @sweir27 - out of curiosity is it possible to omit these new icons and _just_ use the new `Icon.ios.tsx` wrapper? The svg internals seem so similar.

sweir27: `svg` isn't natively supported, so I don't think we can re-use the existing components.

It might be possible to use the `jsx`-ed version on web as well as native via this: https://github.com/bakerface/react-native-svg-web but I haven't looked to hard into this.

cc @l2succes to make sure-- when you made https://github.com/artsy/emission/issues/1273, how did you imagine eventually integrating with palette?

l2succes: I was thinking of using it in palette. the way you're currently going about it @sweir27! 👍 

Although I didn't really explore [`react-native-svg-web`](https://github.com/bakerface/react-native-svg-web) which is an interesting solution. But we mentioned before in the front-end platform meeting, sharing implementations isn't the end goal. This is a fine solution

artsyit:
:rocket: PR was released in v4.14.1 :rocket:

PR #490, opened 2019-06-03
Reorganize image props for web, iOS, and shared Image/ResponsiveImage components

pepopowitz:
## Background

While working on #473, I misunderstood the separation of `Image.tsx`, `Image.ios.tsx`, and `Image.shared.tsx`, and initially added a new web-specific prop to the props that are shared between web _and_ iOS. @zephraph identified my mistake before we merged, but it raised a question of clarity on the roles of these different files. `Image.shared.tsx`, which is intended to be used by both web and iOS, contained several props that are web-specific, and this contributed to me adding another web-specific prop in the wrong place.

## These changes 

This PR moves the web-specific props from `Image.shared.tsx` to `Image.tsx`, via interfaces named `WebImageProps` and `WebResponsiveImageProps`. 

## Open questions

- Did I move the correct props? These are all the ones that I thought were web-specific; let me know if I got any wrong.
- I `yarn link`ed palette to force/reaction _and_ emission with these changes, and did not notice any related `tsc` errors afterward. I _think_ this means the changes are safe for both web & iOS - but please let me know if you think I should do something other than `yarn type-check` to verify they don't break anything.

### DIFF
--- a/packages/palette/src/elements/Image/Image.shared.tsx
+++ b/packages/palette/src/elements/Image/Image.shared.tsx
@@ -19,19 +19,12 @@ import {
   WidthProps,
 } from "styled-system"
 
+/** Props for web & iOS images */
 export interface BaseImageProps {
   /** The url for the image */
   src: string
-  /** Alternate text for image */
-  alt?: string
-  /** A11y text label */
-  ["aria-label"]?: string
-  /** The title of the image */
-  title?: string
   /** Apply additional styles to component */
   style?: object
-  /** Flag for if image should be lazy loaded */
-  lazyLoad?: boolean
 }
 
 export interface ImageProps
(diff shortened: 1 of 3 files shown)


damassi:
> I yarn linked palette to force/reaction and emission with these changes, and did not notice any related tsc errors afterward.

Until we get proper integration tests into palette this unfortunately won't ensure (in a 100% sure way) that things are valid cross platform. While performing the upgrade I had to have emission running, which (thankfully) would immediately call out an error on boot. Since these are interfaces which get stripped out at compile time this is fine, but just a heads up when building new components.

Now that Palette is now in an active cross-platform cycle I think we should prioritize getting these tests RN tests in.

artsyit:

Built with commit 1e06dc929de138e205f8ffc9449848e42c340df3

https://deploy-preview-490--artsy-palette.netlify.com

zephraph:
Thinking out loud here: I wonder if we could have a special ts config that would overwrite the import order of files to use the `.ios` versions first...

zephraph:
This LGTM 👍. I'll wait for @damassi's review.

damassi:
Oh yeah, I'm 👍

artsyit:
:rocket: PR was released in v4.14.2 :rocket:

PR #491, opened 2019-06-03
Add ios type-checking

zephraph:
Figured out how to get a type-checking task working for iOS!!


artsyit:

Built with commit cae7090a8b76a1ee07d7b8a183b323c379cf3df5

https://deploy-preview-491--artsy-palette.netlify.com

packages/palette/tsconfig.ios.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "./*": ["./*.ios", "./*"],
+      "../*": ["../*.ios", "../*"]
```
damassi: What's this one for?

zephraph: I meannn... `./` and `../` are different paths technically.

damassi: Oh, just wondering  why  it's needed if the `baseUrl` points to the root?

packages/palette/tsconfig.ios.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "./*": ["./*.ios", "./*"],
```
ds300: :ceiling-cat:

Can I suggest adding a link to https://github.com/microsoft/TypeScript/issues/8328#issuecomment-219583889 in a comment here? This is a fairly advanced bit of config and it might save someone some googling one day.

:ceiling-cat-out:

alloy:
Niiice!

I recall there not being a solution last time around, but that can’t have been in 2016… so not sure why we didn’t go with this last time around 🤔 Anyways, glad that we have this now 👍 

I agree with @ds300 about leaving a comment. Not sure if tsconfig is json5, if not you should add that context to the commit message.

zephraph:
Can't really do a comment (it's not json5), but I'll update the commit message.

zephraph:
Okay @ds300, @alloy I edited the commit to be much more detailed w/ some context.

ds300:
Thanks :pray: 

Btw, pretty sure tsconfig files are json5! At least tsc treats them as such and has done for years.

artsyit:
:rocket: PR was released in v4.14.4 :rocket:

PR #493, opened 2019-06-04
WIP: Experiment with ReadMore iOS component

sweir27:
I'm currently testing this in emission with:
```
<ReadMore renderer={Markdown} content={this.props.artwork.artist.biography_blurb.text} maxChars={140} />
```

Some things to figure out:
- How to render markdown while also sticking the "... Read more" on the same line. I _think_ this should be do-able but we may need to update our `Markdown` component in Emission so that we could potentially render that + our button side-by-side.
- Animations (if we want them)


artsyit:

Built with commit 56ddedbfd4badae4a741a0fac1c67e11bf5c01a8

https://deploy-preview-493--artsy-palette.netlify.com

Issue #358, opened 2019-03-29
[NEW] Element / DataVisualization / DonutChart


owendodd:
### Zeplin link

zpl://screen?sid=5c9e845cca007d05b8b061ee&pid=5acd19ff49a1429169c3128b

### Product team

Galleries

### Design lead

@owendodd 

### Additional info
The one strange thing about this new component is that we need to use a few new shades of gray that fall outside of our existing color palette. I'd suggest hardcoding these as this usage is hyper specific to this component and I don't see those colors being necessary anywhere else. Happy to discuss further though

---

#### For Elements

_*Buttons, links, loaders, inputs, dropdowns, etc*_

- [ ] Is it using up-to-date tokens from above?

- [ ] Have all states been captured? (hover, selected, disabled, focused, normal, thinking, errors)

- [ ] Have all transitions/animations been defined?

- [ ] Is it useable? (Considering touch targets, screen sizes)

- [ ] Is it consistent with the rest of the visual system? (Corner radius, stroke weight, form, shadow, opacity, etc)

- [ ] Does it meet accessibility standards? (is it keyboard navigable, does it have required accessibility markup)

- [ ] Does it render and function properly in Artsy supported browsers?

#### Accessibility

- [ ] Do all images and multimedia have `alt` or `title` tags?

- [ ] Are semantic elements used appropriately (`nav`, `button`, etc)?

- [ ] Are new components keyboard-navigable?

- [ ] Are hover interactions available by other means?

- [ ] Are `aria-` attributes included where appropriate?

- [ ] Has a Chrome Devtools [accessibility audit](https://developers.google.com/web/tools/chrome-devtools/accessibility/reference#audits) been performed?

#### Compatibility

Has the new component been reviewed in [Browserstack](https://live.browserstack.com/dashboard):

- [ ] Desktop Chrome

- [ ] Desktop Edge

- [ ] Desktop Safari

- [ ] Desktop Firefox

- [ ] Android

- [ ] iOS

damassi:
Before development on this begins we should have a larger discussion between design / engineering about where charting is expected to go at Artsy, and whether a formal charting library may be more maintainable over the long term than solutions that we roll in house. // cc @ds300

owendodd:
@damassi Agreed, I'm definitely supportive of us using some sort of library.

damassi:
Here are some non-exhaustive options: 
- https://uber.github.io/react-vis/
- https://formidable.com/open-source/victory/gallery/
[...]

PR #495, opened 2019-06-04
Donut chart

sepans:
Fixes https://github.com/artsy/palette/issues/358

[Design spec](https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5c9e845cca007d05b8b061ee)

[Preview](https://deploy-preview-495--artsy-palette.netlify.com/elements/data-visualization/donutchart/)

The chart takes the entire horizontal space available to it.

TODO:
- [x] labels
- [x] Remove `height` from props. Add `margin` 
- [x] title/description labels
- [x] calculate donut width and label margin based on screen size (or chart width)
- [x] A bit of cleanup
- [x] tests

donut


artsyit:

Built with commit b5e3484038379ce41ae5b46bfdda0a5d3d1b6b2d

https://deploy-preview-495--artsy-palette.netlify.com

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const svg = (
+    <svg key={"svg"} width={width + margin} height={width + margin}>
+      <g transform={`translate(${centerX}, ${centerY})`}>
+        <Spring
+          from={{ num: 0 }}
+          to={hasEnteredViewport ? { num: 1 } : { num: 0 }}
+          delay={500}
+          onRest={() => setRest(true)}
+        >
```
damassi: Can we upgrade to latest react spring? (Sorry for continuing to raise this, but the latest has hook support while maintaining a backwards-compat API.)

packages/palette/src/elements/DonutChart/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./DonutChart"
```
damassi: https://github.com/artsy/palette#%EF%B8%8F-dont-forget-about-ios

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const [hoverIndex, setHoverIndex] = useState(-1)
+  const [rest, setRest] = useState(false)
+
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  const hasEnteredViewport = useHasEnteredViewport(wrapperRef)
+
+  const width = useGetWrapperWidth(wrapperRef)
+
+  const donutWidth = width * 0.12
```
ds300: According to the spec `donutWidth` should be based on screen size. 30px on mobile and 40px on desktop. I noticed this because the donuts looked a bit fatter than the spec to me.

I personally think slightly fatter is fine, but we should clear it with @owendodd

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,164 @@
...
+            {coerceTooltip(axisLabelX)}
+          </Sans>
+        </DonutLabelContainer>
+      )
+    )
+  })
+
+  const svg = (
+    <svg key={"svg"} width={width + margin} height={width + margin}>
+      <g transform={`translate(${centerX}, ${centerY})`}>
```
ds300: According to the spec, the donut is supposed to have a silhouette ring visible before the arcs animate in.

image

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const [rest, setRest] = useState(false)
+
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  const hasEnteredViewport = useHasEnteredViewport(wrapperRef)
+
+  const width = useGetWrapperWidth(wrapperRef)
+
+  const donutWidth = width * 0.12
+  const donutLabelMargin = space(2) / Math.sqrt(2)
```
ds300: why the division by root 2 here?

packages/palette/src/elements/DonutChart/DonutChart.tsx
ds300: I'm happy to leave that for another PR, since we'll need to QA all the existing animations. Would it make sense to add a ticket to the platform backlog?

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,164 @@
...
+  })
+
+  const labels = pie(values).map((slice, index) => {
+    const [x, y] = labelArc.centroid(slice as any)
+    const axisLabelX = points[index].axisLabelX
+    return (
+      axisLabelX && (
+        <DonutLabelContainer
+          key={index}
+          opacity={rest ? 1 : 0}
```
ds300: These labels fade in a bit too late IMO. Feels sluggish. Would be good if they were already appearing just as the donut animation is ending.

packages/palette/src/elements/DataVis/utils/useGetWrapperWidth.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    setContainerWidth()
+
+    window.addEventListener("resize", setContainerWidth)
+
+    return function cleanup() {
+      window.removeEventListener("resize", setContainerWidth)
+    }
+  }, [])
```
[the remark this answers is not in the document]
sepans: Good point! You mean when the rest of the page changes making the wrapper to resize without `DonutChart` itself being re-rendered?

Adding `ref.current.addEventListener("resize", setContainerWidth)` would handle that case right?

packages/palette/src/elements/DonutChart/DonutChart.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+import "jest-styled-components"
+import React from "react"
+import { Theme } from "../../Theme"
+import { Flex } from "../Flex"
+import { Sans } from "../Typography"
+import { DonutChart, DonutChartProps } from "./DonutChart"
+
+jest.useFakeTimers()
+
+// TODO: add points and test
```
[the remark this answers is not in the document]
sepans: 🙈

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: Sorry I should have mentioned it in the PR. We had a conversation on slack about it: https://artsy.slack.com/archives/C9XJKPY9W/p1559851503006000

Happy to fine tune that ratio if you think it needs to be thinner.

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: Good catch 🙌  it is not needed anymore since the `donutLabelMargin` is added to the radius of `labelArc` and `d3` takes care of the angle.

Previously I was calculating `x, y` distances manually and added this, which also was wrong, I should have had `sin`/`cos`.

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: I agree. I can trigger the opacity when `num` is greater than, let's say 0.7 instead of waiting for the spring animation to end.

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: I tried to upgrade `spring` in previous PR but it wasn't super straightforward. Apart from API change the logic has also changed a little bit and 👆 doesn't work. Which to @damassi point makes it more important to upgrade sooner than later.

But to unblock our team sprint work I also prefer to create a ticket (for myself in galleries backlog) and do it in a follow up PR if you all agree.

packages/palette/src/elements/DataVis/utils/useGetWrapperWidth.ts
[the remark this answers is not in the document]
ds300: 😬 I think `"resize"` only works on `window`. There's a ResizeObserver but seems like it's only supported in Chrome. I think that kind of thing is a rabbit hole that is not worth pursuing right now. Maybe just run `setContainerWidth` on window.resize and also at 1s intervals? _Just in case_? At least until someone has a situation where it needs to be smooth.

packages/palette/src/elements/DonutChart/DonutChart.tsx
ds300: Nah it's good 👍

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: Yeah made that change looks much better ❤️

packages/palette/src/elements/DataVis/utils/useGetWrapperWidth.ts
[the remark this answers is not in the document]
sepans: You are right:
>  It is still possible to set onresize attributes or use addEventListener() to set a handler on any element. However, resize events are only fired on the window object (i.e. returned by document.defaultView). -> [doc](https://developer.mozilla.org/en-US/docs/Web/API/Window/resize_event)

Should I just hold off adding the interval until it is really needed? For partner analytics I don't think it is gonna happen but 👆 is also used in `LineChart`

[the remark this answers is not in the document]
ds300: I think it's a 'better safe than sorry' kind of situation. Will take us a few seconds to add and could potentially mitigate embarrassing rendering issues in the future.. Not concerned about perf. If the width doesn't change the state update will not propagate.

packages/palette/src/elements/DonutChart/DonutChart.tsx
sepans: Added here: https://github.com/artsy/palette/pull/495/commits/1e08b48840a6d8747392def6069bbf495b89cb83

sepans:
hehe:

Screen Shot 2019-06-07 at 12 15 54 PM

packages/palette/src/elements/DonutChart/DonutChart.tsx
ds300: Looks great, thanks!

packages/palette/src/elements/DataVis/utils/useGetWrapperWidth.ts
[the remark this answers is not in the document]
sepans: Added it here: https://github.com/artsy/palette/pull/495/commits/f97109bc0289841d0f9735fa4117ca7abeb5c174.

I guess calling `ref.current.getBoundingClientRect()` every second shouldn't be a huge performance deal but I am worried in some edge cases it gets into a loop 😬 

This is how it looks now (storybook). With 100ms interval it is going to be smooth but maybe that is too much.

donuthover

damassi:
Lol @sepans! Perfect

packages/palette/src/elements/DonutChart/DonutChart.tsx
damassi: @sepans - sounds good. If possible, please follow up on 1807 in a timely manner or file in Platform and we can take care of it.

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/**
+ * DonutChart is a component that displays data points with donut shaped arcs.
+ * Good for illustrating numerical proportions.
+ */
+export const DonutChart: React.FC<DonutChartProps> = ({
+  points,
+  margin = space(3),
+}: DonutChartProps) => {
```
damassi: `React.FC<DonutChartProps>` will cover this

packages/palette/src/elements/DonutChart/DonutChart.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const [hoverIndex, setHoverIndex] = useState(-1)
+  const [labelFadeIn, setLabelFadeIn] = useState(false)
+
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  const hasEnteredViewport = useHasEnteredViewport(wrapperRef)
+
+  const width = useGetWrapperWidth(wrapperRef)
```
damassi: Maybe just `useWrapperWidth` since a hook is presumed to return something?

ds300:
:tada: :heart: 💃 🎈

artsyit:
:rocket: PR was released in v4.14.12 :rocket:

PR #496, opened 2019-06-05
Fix error when all linechart values are zero

sepans:
When all values are zero, `maxValue` ends up being zero causing division by zero.

We currently don't support negative values for line chart so this would fix the problem.

### DIFF
--- a/packages/palette/src/elements/LineChart/LineChartSVG.tsx
+++ b/packages/palette/src/elements/LineChart/LineChartSVG.tsx
@@ -39,7 +39,7 @@ export const LineChartSVG: React.FC<LineChartSVGProps> = ({
   const h = height - 2 * margin
 
   // maps value to x/y position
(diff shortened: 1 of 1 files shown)


oxaudo:
Looks legit! Thanks for fixing. Merge on green!

Unrelated note - maybe it worth to add a story with several graphs examples so that they can be included in visual tests. It's cool that it verifies that we don't break anything drastically.

artsyit:

Built with commit 694d4e96c0fd934965eef9448a7ad72eb5c97536

https://deploy-preview-496--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.14.4 :rocket:

sepans:
Yes great idea @oxaudo  will do that! When I did `LineChart` didn't know about the storybook!

PR #498, opened 2019-06-06
[WIP] Upgrade @types/styled-components

alloy:
Got to the part where things need to be different for web vs RN, i.e. RN isn’t a node or web env.

Had to give up for now, as I didn’t want to go further down the rabbit hole. If anybody else feels like taking a next stab, by all means please do 👍


artsyit:

Built with commit cb3add7b4af5324d5eda38ed60bc799c64b7abe6

https://deploy-preview-498--artsy-palette.netlify.com

erikdstock:
I was wondering how palette would approach this ...

zephraph:
Just looking at the CI failures, it looks like it's failing b/c of duplicate definitions between react native and non-react native. 

Unfortunately this is a really hard problem to solve. TypeScript doesn't provide a way to exclude a single type. We can specific which types to include, but then we have to manually specific _all_ the types we use (minus the ones we don't) so it's a fairly untenable situation. 

In the past I've had scripts to automatically delete RN types when building for web. It's not a great solution. We'll have to do some more research around this, but I suspect it'll be really hard. 

Here's a thread on this issue:

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311

And another thread requesting excludeTypes:

https://github.com/microsoft/TypeScript/issues/18588

PR #505, opened 2019-06-07
Removes version 2 of updated luxon files

kierangillen:
- Removes version 2 of updated components using Luxon
- Removes `moment` and `moment-timezone`
- Moves `luxon` to peer and dev dependency only
- These changes had been reverted in https://github.com/artsy/palette/pull/504 but previously approved in https://github.com/artsy/palette/pull/502 and https://github.com/artsy/palette/pull/499

PR #508, opened 2019-06-10
Allow any 1.x version of luxon

kierangillen:
- Hit the merge button too early on https://github.com/artsy/palette/pull/505
- Allow any 1.x version of luxon


artsyit:

Built with commit 1b431c1893b99d7c9e12b8e152fddc83b1eb3170

https://deploy-preview-508--artsy-palette.netlify.com

packages/palette/package.json
```diff
@@ -70,7 +70,7 @@
     "@types/d3-shape": "1.3.1",
     "@types/enzyme": "3.9.1",
     "@types/jest": "23.3.14",
-    "@types/luxon": "1.15.1",
+    "@types/luxon": "1.15",
```
alloy: This is missing the caret. Not sure what that leads to, will this be pinned to 1.15.0?

kierangillen: It looks like all of the @types don't have carets. Is this expected then?

alloy: Ah, I think that’s a renovate thing. In any case, looks like we are installing at least `1.15.1` so it should be fine https://github.com/artsy/palette/blob/12597a2ec97fc760e4902229d9156aa444a9a7ef/yarn.lock#L3415

artsyit:
:rocket: PR was released in v4.14.14 :rocket:

PR #510, opened 2019-06-11
Add pointer-event: none to tooltip content

sepans:
## Problem:
When there are multiple tooltips vertically aligned the tooltip content of the one on the bottom can block the hover area of the one above:

Screen Shot 2019-06-11 at 4 32 49 PM

## Solution: 
adding `pointer-evnets: none` to content. It is also possible to fix this with changing `z-index` along with `opacity` but `pointer-events` is simpler and in fact it is not even possible to get the mouse over the content area when it is visible.


artsyit:

Built with commit f71eaad30a8e25a1a349b099b7d506e4d6bfe82e

https://deploy-preview-510--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.14.14 :rocket:

PR #515, opened 2019-06-15
Button & Spinner Component for React Native

l2succes:
This PR ports the Button and Spinner component from emission into palette. This also sets some guidelines on how to share common code and interfaces between React Native and React Web version of a component.

Note this works best for components that behave essentially the same on both platforms and can thus share a common interface.

Kapture 2019-06-15 at 1 00 25

Design spec:

Screen Shot 2019-06-15 at 1 09 51 AM

PR #516, opened 2019-06-15
Fix import path

damassi:
Fixes an import path error from https://github.com/artsy/palette/pull/515

### DIFF
--- a/packages/palette/src/elements/Button/Button.ios.tsx
+++ b/packages/palette/src/elements/Button/Button.ios.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react"
 import { TouchableWithoutFeedback } from "react-native"
-import { animated, Spring } from "react-spring/dist/native"
+import { animated, Spring } from "react-spring/renderprops-native.cjs"
 import styled from "styled-components/native"
 import { themeProps } from "../../Theme"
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/Button/Button.ios.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from "react"
 import { TouchableWithoutFeedback } from "react-native"
-import { animated, Spring } from "react-spring/dist/native"
+import { animated, Spring } from "react-spring/renderprops-native.cjs"
```
damassi: @l2succes - we should be using hooks for animation now, but in the case of porting over code this is the new import path in `react-spring` 8.x.

artsyit:

Built with commit 3ac5dfe670d9acdef79cefa9891818cf1703cc4c

https://deploy-preview-516--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v4.14.17 :rocket:

packages/palette/src/elements/Button/Button.ios.tsx
l2succes: Got it, sounds good @damassi

PR #518, opened 2019-06-17
Add support for React Native component preview

l2succes:
Related: https://artsyproduct.atlassian.net/browse/GROW-1328

As part of ramping up our React Native support in Palette, we needed a way to preview how components render and behave for mobile. This enables us to do just that by defining a way to add a way to toggle between Web and Mobile rendering mode for a given component.

To enable mobile support for a component, just add a `platforms` property to frontmatter with either `web`, `mobile` or both. 

E.g.

```
---
name: Button
lastPointOfContact: Nicole
platforms: ["web", "mobile"]
---
```

You also have to follow the naming convention and directory structure below. In the case of 
```
elements/
└── Box/
    ├── Box.ios.tsx
    ├── Box.tsx
    └── index.tsx
```

Kapture 2019-06-25 at 10 12 48

### DIFF
--- a/packages/palette-docs/package.json
+++ b/packages/palette-docs/package.json
@@ -101,8 +101,9 @@
     "react": "^16.8.2",
     "react-dom": "^16.8.2",
     "react-helmet": "^5.2.0",
+    "react-native-web": "^0.11.4",
     "react-powerplug": "^1.0.0",
-    "styled-components": "^4.1.3",
+    "styled-components": "^4.3.1",
     "unstated": "^2.1.1"
   },
   "license": "MIT",
--- a/packages/palette-docs/src/components/MobilePreview.jsx
+++ b/packages/palette-docs/src/components/MobilePreview.jsx
@@ -1,4 +1,20 @@
 import React from "react"
 import { IPhone } from "./iPhone"
+import styled from "styled-components"
+import { Container } from "unstated"
 
-export const MobilePreview = ({ children }) => <IPhone>{children}</IPhone>
+export const MobilePreview = ({ children }) => (
+  <IPhone>
+    <InnerContainer>{children}</InnerContainer>
+  </IPhone>
+)
+
+const InnerContainer = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  overflow-y: auto;
+  padding: 0 5px;
+`
--- a/packages/palette-docs/src/components/MobilePreview.jsx
+++ b/packages/palette-docs/src/components/MobilePreview.jsx
@@ -1,0 +1,4 @@
+import React from "react"
+import { IPhone } from "./iPhone"
+
+export const MobilePreview = ({ children }) => <IPhone>{children}</IPhone>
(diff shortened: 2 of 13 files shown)


artsyit:

Built with commit cf0371044fb88d4506875844ba823fc8e885019a

https://deploy-preview-518--artsy-palette.netlify.com

packages/palette-docs/package.json
```diff
@@ -101,8 +101,9 @@
     "react": "^16.8.2",
     "react-dom": "^16.8.2",
     "react-helmet": "^5.2.0",
+    "react-native-web": "^0.11.4",
```
l2succes: Used to render our React Native components in the browser

packages/palette-docs/src/components/MobileRenderer.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <MDXRenderer
+        scope={{
+          CodeEditor,
+          ColorComponent,
+          MobilePreview,
+          Playground,
+          Toggler,
+          ...Elements,
+          ...MarkdownComponents,
+          Box: require("@artsy/palette/dist/elements/Box/Box.ios").Box,
```
l2succes: The idea here is we override the web version of those components with the React Native version at runtime. I couldn't figure out a more elegant way to resolve those components since webpack is not aware of the `ios` platform extension. 

cc @damassi @alloy let me know if you can think of something. I was thinking a script that would generate this for all `ios.tsx` files in the src directory tree.

packages/palette/src/elements/Box/Box.ios.tsx
```diff
@@ -0,0 +1,77 @@
+// @ts-ignore
+import React from "react"
+import { styled as primitives } from "../../platform/primitives.ios"
```
l2succes: The only reason I needed to duplicate this file was to hardcode `primitives.ios` here. Otherwise the web version of Box would break rendering using `react-native-web`

packages/palette-docs/src/components/MobilePreview.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+import { IPhone } from "./iPhone"
+import styled from "styled-components"
+import { Container } from "unstated"
```
damassi: Stray

packages/palette-docs/src/components/MobileRenderer.tsx
damassi: cc @zephraph

packages/palette/src/elements/Box/Box.ios.tsx
damassi: This is fine for now but we should figure this out as to avoid duplication

damassi:
This is looking good @l2succes -- I've reassigned this to @zephraph as i'll be out for a few weeks. 

My one thought is we should continue experimenting so we don't break platform resolutions and need to duplicate. I bet there's a solution close by.

packages/palette-docs/src/components/MobileRenderer.tsx
zephraph: It'll be really hard to do at build time without building twice. While this approach works, my concern is that it would transparently break if we added a new `.ios` file for a component.

damassi:
Going to close this until there's interest in picking this up.

PR #530, opened 2019-07-02
Fixes Spring bug and updates Inline to remove padding and height

kierangillen:
- `Inline` now removes `height` and `padding` 
- Fixes React Spring bug where `transparent` background causes crash, replacing `transparent` with `rgba(0, 0, 0, 0)` is the fix.

Example of inline button:

Screen Shot 2019-07-02 at 2 51 37 PM

### DIFF
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -34,7 +34,7 @@ export class Button extends Component<ButtonProps> {
         return {
           height: inline ? "17px" : "26px",
           size: "2",
-          px: inline ? 0 : 1,
+          px: inline ? 0 : 2,
         }
       case "medium":
         return {
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -28,13 +28,26 @@ export class Button extends Component<ButtonProps> {
   }
 
   getSize(): { height: string; size: "2" | "3t"; px: number } {
+    const { inline } = this.props
     switch (this.props.size) {
       case "small":
-        return { height: "26px", size: "2", px: 2 }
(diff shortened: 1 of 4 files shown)


artsyit:

Built with commit e375a81937f6cc275d955f0e7e48c29f98306aa0

https://deploy-preview-530--artsy-palette.netlify.com

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -28,13 +28,26 @@ export class Button extends Component<ButtonProps> {
...
 
   getSize(): { height: string; size: "2" | "3t"; px: number } {
+    const { inline } = this.props
     switch (this.props.size) {
       case "small":
-        return { height: "26px", size: "2", px: 2 }
+        return {
+          height: inline ? "17px" : "26px",
+          size: "2",
+          px: inline ? 0 : 1,
```
zephraph: This changes the original behavior to have 10px padding instead of 20px. Is that intended?

zephraph: This also shows up in the chromatic test.

l2succes: This only applies to `<Button inline />` I just checked reaction for its usage and there are no instances of it as of now.

l2succes: @samchieng just to double check, does this change sound ok?

zephraph:
I think the chromatic failures are due to the conflicts. Once we fix those and get this branch up-to-date w/ master we should be good.

kierangillen:
@zephraph just assigned you to this and rebased, thanks!

artsyit:
:rocket: PR was released in v4.14.37 :rocket:

PR #536, opened 2019-07-10
Adds longestText prop to Button component

kierangillen:
- Adds a prop to `Button` to pass the longest string to keep the button width fixed at the longest text width. I.e. a button with "Following" and "Follow" states will always retain the width of "Following." This allows for enlarged text sizes and is a fix for otherwise necessitating hard coded button widths.

- Replaces `purple100` for `black100` background color from iOS button transitions on request from design team.

### DIFF
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -192,7 +192,12 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
         >
           {children}
         </VisibleText>
-        <HiddenText pt="1px" weight={weight || "medium"} size={size}>
+        <HiddenText
+          role="presentation"
+          pt="1px"
+          weight={weight || "medium"}
+          size={size}
+        >
           {longestText ? longestText : children}
         </HiddenText>
       </Container>
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -166,6 +166,7 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
       disabled,
       color,
       size,
+      longestText,
       weight,
       onClick,
       ...rest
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit b4490c50f0a99bb45b09de7a2d707a517e9a35fc

https://deploy-preview-536--artsy-palette.netlify.com

packages/palette/src/elements/EntityHeader/EntityHeader.tsx
```diff
@@ -52,7 +52,13 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
       )}
 
       <Flex flexDirection="column" justifyContent="center" width="100%">
-        <Serif size="3" weight="semibold" color="black100">
+        <Serif
```
sweir27: Are these changes meant to be included?

(Also, they look like they belong in the `ios` version of this component, but I think that change was already made).

kierangillen: oh whoops!

kierangillen: Cleaned this up

sweir27:
merge on green

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -183,14 +184,34 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
...
+        <VisibleText
+          pt="1px"
+          weight={weight || "medium"}
+          color={color}
+          size={size}
+        >
           {children}
-        </Sans>
+        </VisibleText>
+        <HiddenText pt="1px" weight={weight || "medium"} size={size}>
```
ds300: Might be worth adding `role="presentation"` here in case we care about accessibility one day 🙃 Don't want screen readers reading both labels out 😬

kierangillen: Good idea, I didn't add because my understanding of screen readers is that they won't read anything visibly hidden but better to be clear anyway.

kierangillen: Added this.

artsyit:
:rocket: PR was released in v4.14.28 :rocket:

PR #540, opened 2019-07-19
PURCHASE-1019 - Adds modal component

kierangillen:
- https://artsyproduct.atlassian.net/browse/PURCHASE-1019


artsyit:

Built with commit 4f1090df47c297f57c88353745e4522377415ed1

https://deploy-preview-540--artsy-palette.netlify.com

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+const FixedButtonWrapper = styled(Box)`
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  padding-left: ${space(4)}px;
+  padding-right: ${space(4)}px;
+  padding-top: ${space(2)}px;
+  padding-bottom: ${space(3)}px;
```
damassi: These can be moved declaratively inline -- `px={4} pt={2} pb={3}`

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+const ModalOuterWrapper = styled(Box)<TransitionElementProps>`
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 100;
+  width: 100vw;
+  height: 100vh;
+  background-color: rgba(229, 229, 229, 0.5);
+  transition: opacity 250ms ease;
```
damassi: Why mix react-spring + CSS animation?

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+      (previousHasLogo && hasLogo !== previousHasLogo)
+    ) {
+      setSpringContentAnimation({
+        opacity: 0,
+        onRest: () => updateVisibleContent(),
+      })
+    }
+  }, [hasLogo, title, FixedButton, children])
+
+  const content = () => {
```
damassi: #minor - this is technically a FC component, so should be `const SomeComponent = () => { ... }` (caps case)

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  hasLogo?: boolean
+  FixedButton?: JSX.Element
+  show?: boolean
+  onClose?: () => void
+}
+
+interface TransitionElementProps {
+  show?: boolean
+  isAnimatingOpen?: boolean
+  isAnimatingClosed?: boolean
```
damassi: How are these props used?

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+import { Spacer } from "../Spacer"
+import { Serif } from "../Typography"
+
+interface ModalProps {
+  title?: string
+  isWide?: boolean
+  hasLogo?: boolean
+  FixedButton?: JSX.Element
+  show?: boolean
+  onClose?: () => void
```
damassi: Alpha would be nice; highlight list, `cmd + shift + p > sort ...`

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,238 @@
...
+  width: ${props => (props.isWide ? "900px" : "440px")};
+`
+
+const ModalScrollContent = styled(Box)<ModalProps>`
+  height: 100%;
+  width: 100%;
+  overflow: scroll;
+  padding-left: ${space(4)}px;
+  padding-right: ${space(4)}px;
+  padding-top: ${space(2)}px;
```
damassi: `px={4} pt={2}`

packages/palette/src/elements/Modal/index.tsx
kierangillen: I'd like to use `react-spring` sparingly and therefore it's only used when an animation finished callback is necessary and only on a single element.

packages/palette/src/elements/Modal/index.tsx
kierangillen: Cool updated this.

packages/palette/src/elements/Modal/index.tsx
kierangillen: Fixed this

packages/palette/src/elements/Modal/index.tsx
kierangillen: Oh whoops they were artifacts, deleted these.

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,234 @@
...
+    setFixedButtonState(FixedButton)
+    setHasLogoState(hasLogo)
+    // Replaces content
+    setVisibleContent(ModalContent)
+    // Fades in new content
+    setSpringContentAnimation({
+      opacity: 1,
+      onRest: null,
+    })
+  }
```
damassi: This function can be moved up, closer to where it's called. Then the renderable content (`ModalContent`, `return (...)`) can be grouped nicely side-by-side.

damassi:
Two qs: 

1) The fade-in / fade-out. Strictly necessary? There's a lot of complexity around fading existing content out and then fading new content in that would vanish if we _only_ faded content in. Then, when new props were passed, opacity starts at zero and moves to 1. No libraries needed and everything remains declarative, no state is needed, etc. Fading stuff out and then in is a nice touch, but fade in only can also be done elegantly. This then leads to...

2) It makes testing easier. 

Thoughts?

kierangillen:
@damassi I agree it'd be nicer programmatically for sure. I went for the complexity based on the fade out being explicitly requested in the spec. Can check with the designer.

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,234 @@
+import React, { SFC, useEffect, useRef, useState } from "react"
```
ds300: why is the implementation in `index.tsx` and the export in `Modal.tsx`? Isn't it the other way around in most cases? Also I'd advocate for not even having an index file, FWIW.

packages/palette/src/elements/Modal/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  show?: boolean
+}
+
+const usePrevious = value => {
+  const ref = useRef(null)
+  useEffect(() => {
+    ref.current = value
+  }, [value])
+  return ref.current
+}
```
damassi: Can this be moved to our [utils directory](https://github.com/artsy/palette/tree/master/packages/palette/src/helpers)?

kierangillen: Sure just moved it

packages/palette/src/elements/Modal/index.tsx
kierangillen: I just removed the index file -- but this is the convention we've used for every palette component so on one hand it may be good to be consistent.

packages/palette/src/elements/index.tsx
```diff
@@ -16,7 +16,7 @@ export * from "./Input"
 export * from "./Join"
 export * from "./Menu"
 export * from "./Message"
-export * from "./Modal"
+export * from "./Modal/Modal"
```
damassi: Totally understand the index-files-are-weird argument but we do have a convention in palette and it's used throughout. If we want to change that convention we should do so in a separate PR that a) argues reasoning; and b) updates everything, rather than slipping in one-off changes like the above.

ds300: Yeah naming/structural consistency is more important IMO :+1:

We could just delete all the index files and get the best of both worlds at the cost of one weird diff, but I don't feel like it matters much either way. Happy for this one index file to be reinstated as long as it's just a single export statement like the others.

artsyit:
:rocket: PR was released in v4.14.34 :rocket:

oxaudo:
This is cool. Do you by chance know if there are plans to move ModalDialog as well @kierangillen ?
I've just used it recently in volt directly from Reactions - think it would be cooler to get it from Reaction but was not following on design plans...

kierangillen:
@oxaudo Hey I'm not sure! Maybe talk to your PM? I'm down to move as much to Palette as possible.

PR #541, opened 2019-07-19
Bar chart zero state

sepans:
### Before:

Screen Shot 2019-07-19 at 5 13 28 PM

### After:

Screen Shot 2019-07-19 at 5 32 43 PM

### Questions:

- Is it by design that the highlighted bar has at least has `10px` height? It is defined [here](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/BarChart/Bar.tsx#L164-L166):

Screen Shot 2019-07-19 at 5 34 09 PM

- Removed [this](https://github.com/artsy/palette/compare/master...sepans:bar-chart-zero?expand=1#diff-710f36809acea9f66cf6ae101cf56b66L14) `flex: 1`. Doesn't seem to be doing anything except making charts grow from top instead of bottom when there is no `highlightLabel`

bar_grow

### DIFF
--- a/packages/palette/src/elements/BarChart/BarChart.story.tsx
+++ b/packages/palette/src/elements/BarChart/BarChart.story.tsx
@@ -226,3 +226,94 @@ storiesOf("Components/BarChart", module)
       </Box>
     )
   })
+
+  .add("Zero state with highlight and x axis label", () => {
+    return (
+      <Box width="50%">
+        <BarChart
+          bars={[
+            {
+              value: 0,
+              axisLabelX: "Sep 30",
+            },
+            {
+              value: 0,
+            },
+            {
+              value: 0,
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 4002842fa174a5564f939e03598b703d46f9acd6

https://deploy-preview-541--artsy-palette.netlify.com

ds300:
Hey, I'll answer your questions now and do a review right after!

> Is it by design that the highlighted bar has at least has 10px height?

Yes :) https://artsy.slack.com/archives/C9YNS4X32/p1554200961150900

> Removed this flex: 1. Doesn't seem to be doing anything except making charts grow from top instead of bottom when there is no highlightLabel

Haha weird. Thanks for fixing!

artsyit:
:rocket: PR was released in v4.14.31 :rocket:

PR #546, opened 2019-07-25
PURCHASE-1361 Fixes button breaking viewport

kierangillen:
- Fixes Buttons breaking the width of their container based on the hardcoded width. Now computes width from the view width.

## After fix

Screen Shot 2019-07-25 at 1 20 43 PM

### DIFF
--- a/packages/palette/src/elements/EntityHeader/EntityHeader.ios.tsx
+++ b/packages/palette/src/elements/EntityHeader/EntityHeader.ios.tsx
@@ -1,5 +1,4 @@
 import React, { SFC } from "react"
-import { Dimensions } from "react-native"
 import { Avatar } from "../Avatar"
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit eadad78d3c3d73bc05da78a229c83c4b383fe467

https://deploy-preview-546--artsy-palette.netlify.com

ds300:
As mentioned in person, I found a way to fix this which removes some code and allows the component to be variable-width, but I don't quite understand why it works. You think it's worth investigating, @kierangillen ? I'd want to check that it works in other contexts and on various devices.

kierangillen:
#mergeongreen

artsyit:
:rocket: PR was released in v4.14.39 :rocket:

PR #547, opened 2019-07-25
Modal improvements

kierangillen:
- Adds styling for mobile
- Replaces animation on `children` change for optional `contentAnimationKey` prop
- Adds more styling and allows for escaping via the keyboard


artsyit:

Built with commit 39be6acb5314e7ba0f553ad33c12af68d0e2969d

https://deploy-preview-547--artsy-palette.netlify.com

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -11,6 +11,8 @@ import { Serif } from "../Typography"
 
 interface ModalProps {
   FixedButton?: JSX.Element
+  // contentAnimationKey should change if the modal displays new content
+  contentAnimationKey?: string
```
damassi: This is a bit opaque naming wise -- maybe `refresh` or something a little more user friendly?

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -51,11 +55,34 @@ export const Modal: SFC<ModalProps> = ({
...
+    }
+  }
+
+  const updateBodyScrollBlock = () => {
+    if (show) {
+      document.body.style.overflowY = "hidden"
+    } else {
+      document.body.style.overflowY = "visible"
+    }
+  }
```
damassi: 67 - 81 should be wrapped in a `useEffect` hook

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -51,11 +55,34 @@ export const Modal: SFC<ModalProps> = ({
...
+
   useEffect(() => {
+    updateBodyScrollBlock()
+    updateEscapeKeyListener()
```
damassi: It might be worth moving the contents of these functions inside of here with a comment, and consolidating around the `if (show)` conditional, rather than duplicating. It reads clear enough without the need for functions.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -160,7 +184,7 @@ const ModalOuterWrapper = styled(Box)<TransitionElementProps>`
   position: fixed;
   top: 0;
   left: 0;
-  z-index: 100;
+  z-index: 9999;
```
damassi: 🤔 it's better to be precise with z-indexes

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -51,11 +55,34 @@ export const Modal: SFC<ModalProps> = ({
...
 
+  const handleEscapeKey = event => {
+    if (event && event.key === "Escape") {
+      onClose()
+    }
+  }
+
+  const updateEscapeKeyListener = () => {
+    if (show) {
+      document.addEventListener(KEYBOARD_EVENT, handleEscapeKey, true)
```
damassi: Don't think this needs to be moved into a `KEYBOARD_EVENT` var. Rule of three's  / doesn't add much descriptive value.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -11,8 +11,8 @@ import { Serif } from "../Typography"
 
 interface ModalProps {
   FixedButton?: JSX.Element
-  // contentAnimationKey should change if the modal displays new content
-  contentAnimationKey?: string
+  // refreshKey should change if the modal displays new content
+  refreshKey?: string
```
damassi: maybe `refreshModalContent`?

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -51,12 +53,30 @@ export const Modal: SFC<ModalProps> = ({
...
+    if (event && event.key === "Escape") {
+      onClose()
+    }
+  }
+
   useEffect(() => {
-    if (!show) {
+    if (show) {
+      // Binds key event for escape to close modal
+      document.addEventListener("keyup", handleEscapeKey, true)
```
damassi: Possible to add a test for this?

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -92,13 +104,11 @@ export const Modal: SFC<ModalProps> = ({
...
-        opacity: 0,
-        onRest: () => updateVisibleContent(),
-      })
-    }
-  }, [children])
+    setSpringContentAnimation({
+      opacity: 0,
+      onRest: () => updateVisibleContent(),
+    })
+  }, [refreshModalContent])
```
damassi: This feels kinda weird to me, like it might confuse devs further down the road who are wondering why things aren't updating right. I understand the reasoning (pass this vs checking diffs against specific content props) but it's a little less intuitive.

kierangillen: Yeah ideally we'd check vs `children` but because the `children` here are so variable it's hard to pinpoint an exact prop to check against. If we check `children` vs `previousChildren` without a specific key then then it's always returned true.

damassi: Yah, `children` being a complex object / function it's not comparable (only shallow comparisons are possible). 🤔Yeah, I don't have a good answer here.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { SFC, useEffect, useState } from "react"
 import { animated, useSpring } from "react-spring"
 import styled from "styled-components"
-import { color, space, usePrevious } from "../../helpers"
```
damassi: If `usePrevious` is no longer being used we can delete it

packages/palette/src/elements/Modal/__tests__/Modal.test.tsx
```diff
@@ -54,4 +60,17 @@ describe("Modal", () => {
...
+    const onClose = () => (show = false)
+    mount(
+      <Modal show={show} onClose={onClose}>
+        Aenean eu leo quam. Pellentesque ornare sem lacinia quam venenatis
+        vestibulum.
+      </Modal>
+    )
+    document.dispatchEvent(new KeyboardEvent("keyup", { key: "Escape" }))
+    expect(show).toEqual(false)
+  })
```
damassi: 👍

artsyit:
:rocket: PR was released in v4.14.37 :rocket:

PR #555, opened 2019-08-01
Add tests for SVG chart animations

sepans:
It turned out that the SVG chart animations (currently donut and line charts) are sensitive to the version of `react-spring` it is using. This hopefully catches that.

- Why not chromatic? The chart may render correctly without animation and chromatic wouldn't catch that
- Why `mock-raf` dependency? It is a small dev dependency but definitely can make our own, though it will be _very_ similar.

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -93,6 +93,7 @@
     "jest": "23.6.0",
     "jest-styled-components": "7.0.0-2",
     "lint-staged": "8.1.5",
+    "mock-raf": "^1.0.1",
     "prettier": "1.17.0",
     "react": "16.8.4",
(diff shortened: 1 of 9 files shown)


artsyit:

Built with commit 43a461333d9cd98e83d90a68badee7a8060bbc4c

https://deploy-preview-555--artsy-palette.netlify.com

zephraph:
@sepans while we're here, would you mind adding a delay to the chromatic tests so that we don't have to keep manually accepting the results when the animation hasn't completed?

https://docs.chromaticqa.com/delay

artsyit:
:rocket: PR was released in v4.14.44 :rocket:

PR #556, opened 2019-08-02
Doesn't render modal in DOM if it's offscreen

kierangillen:
- Removes the `Modal` from the DOM unless it should be visible.

### DIFF
--- a/packages/palette/src/elements/Modal/Modal.tsx
+++ b/packages/palette/src/elements/Modal/Modal.tsx
@@ -16,6 +16,7 @@ import { Serif } from "../Typography"
  * updates immedately.
  */
 interface ModalProps {
+  oa
   FixedButton?: JSX.Element
   refreshModalContentKey?: string
   hasLogo?: boolean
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 73e7151ec2768f36b0d7ea3d897398bf753f62fb

https://deploy-preview-556--artsy-palette.netlify.com

artsyit:

Built with commit ddad0dea87383ad46cb6c2fe189094a1861208b4

https://deploy-preview-556--artsy-palette.netlify.com

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -86,6 +87,7 @@ export const Modal: SFC<ModalProps> = ({
 
   useEffect(() => {
     if (show) {
+      setRenderModal(true)
```
erikdstock: If this effect depends on `show` (L118) and our return value looks like ```
<>{renderModal && (<ModalOuterWrapper show={show}> ...
```

Does that mean that the `ModalOuterWrapper` effects achieved by its `show` prop would be lost (since the modal would simply no longer render after `useEffect` first with `show=false`?

kierangillen: Hey it's a bit hard to read/get what you're saying here, maybe we can review in person?

erikdstock: Talked in person, verified on deploy preview, all good.

artsyit:
:rocket: PR was released in v4.14.41 :rocket:

PR #563, opened 2019-08-14
New CSSGrid element

pepopowitz:
This PR introduces a new element named `CSSGrid`. It enables us to easily render elements with the style `display:grid` and responsive props.

It's named `CSSGrid` instead of `Grid` because [we already have a `Grid` element](https://palette.artsy.net/tokens/grid/). The name `CSSGrid` clearly differentiates it from our other `Grid` element.

I added docs to the Layout section, because I think this element fits better with the `Flex` element instead of the `Grid` element. 

Example usage is seen in [the story](https://github.com/artsy/palette/compare/master...pepopowitz:css-grid-element#diff-065d2efcfb6e5da648844bc18be7bdfaR10):

```
      <CSSGrid
        width={[220, 420, 680]}
        gridGap={[2, 3, 4]}
        gridTemplateColumns={[
          "repeat(2, 1fr)",
          "repeat(3, 1fr)",
          "repeat(4, 1fr)",
        ]}
      >
        {...contents of grid...}
      </CSSGrid>
```

In this example, the grid will display 2 columns at xs, 3 columns at sm, and 4 columns at anything larger. As seen in this gif:

css-grid-responsive

---

## Questions

* I _think_ I handled iOS correctly. I added a `CSSGrid.ios.tsx` that doesn't export anything. As far as I can tell, React Native doesn't support CSS grid, so `CSSGrid.tsx` would break in emission. Could someone confirm that I've addressed this properly?


artsyit:

Built with commit f99f831e421b338e1f9c26e431978b6d195ace80

https://deploy-preview-563--artsy-palette.netlify.com

packages/palette/src/elements/CSSGrid/CSSGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+  GridTemplatesAreasProps,
+  GridTemplatesColumnsProps,
+  GridTemplatesRowsProps,
+} from "styled-system"
+import { Box } from "../Box"
+
+/**
+ * All props available to the CSSGrid component
+ */
+export type CSSGridProps = GridAutoColumnsProps &
```
damassi: We should use an interface, rather than a type

packages/palette-docs/content/docs/elements/layout/CSSGrid.mdx
```diff
@@ -0,0 +1,31 @@
+---
+name: CSSGrid
+---
+
+The CSSGrid component is a layout container for content that exposes props
+related to CSS grid.
```
damassi: Can we move the other Grid to `layout` as well so that these items live side-by-side, and add cross references in that Grid's docs to this and vice versa?

kierangillen:
Don't worry about the `.ios.tsx` file being blank, in general we shouldn't be adding content here, unless it's easily replicated, if we don't have a current use for it.

packages/palette-docs/content/docs/elements/layout/Grid.mdx
```diff
@@ -1,13 +1,20 @@
...
 import _ from "lodash"
 
+The Grid element is a layout container for rendering a Bootstrap-style twelve
+column grid.
+
 See [styled-bootstrap-grid](https://github.com/dragma/styled-bootstrap-grid) for
 full documentation.
 
+Palette also offers a [CSSGrid element](/elements/layout/cssgrid/), which can be
+used for layouts based on the `display:grid` CSS feature.
```
damassi: 👍

artsyit:
:rocket: PR was released in v4.15.0 :rocket:

PR #564, opened 2019-08-14
Creates TimeRemaing.ios.tsx and updates TimeRemaining API

kierangillen:
- Updates the `TimeRemaing` API with new props: `trailingText` & `timeEndedDisplayText`
- Creates a `useCurrentTime` hook which managers the current time in an interval. If either `TimeRemaining` or `StaticCountdownTimer` isn't passed a specific `currentTime` prop the time will be counted down using the `DateTime.local()` time.
- `TimeRemaining` now defaults to `00d 00h 00m 00s` display with an optional override using `timeEndedDisplayText`
- TimeRemaining now has the option to pass `labelWithTimeRemaining` or `labelWithoutTimeRemaining` to display a label below the timer. This will allow this component to seamlessly replace the current Reaction component ( https://github.com/artsy/reaction/blob/master/src/Components/v2/Timer.tsx )


packages/palette-docs/content/docs/components/StaticCountdownTimer.mdx
```diff
@@ -6,18 +6,7 @@ name: StaticCountdownTimer
...
     note="Before it's too late."
-    countdownStart="2019-01-01T12:00:00.000-04:00"
-    countdownEnd="2019-01-14T15:30:00.000-04:00"
-    currentTime="2019-01-05T12:00:30.000-04:00"
-  />
-</Playground>
-
-<Playground>
-  <StaticCountdownTimer
-    action="You forgot to do the thing 😵"
```
damassi: Why remove this?

kierangillen: Just fixed this accidentally pushed it, you're too fast on the trigger :)

packages/palette-docs/content/docs/components/TimeRemaining.mdx
```diff
@@ -4,16 +4,16 @@ name: TimeRemaining
...
     highlight="purple100"
   />
 </Playground>
 
 <Playground title="Out of time">
   <TimeRemaining
-    countdownEnd="2019-01-03T20:15:00.000-04:00"
-    currentTime="2019-01-05T12:00:00.000-04:00"
+    currentTime="2019-01-02T20:15:00.000-04:00"
+    endDate="2019-01-03T20:15:00.000-04:00"
```
damassi: I think a previous comment got eaten by rebase, but why change `countdownEnd`? I'm cool with the change if certain there aren't a lot of consumers that would need to be updated, but maybe change to `endTime` to be in alignment with `currentTime`, or even `endDateTime` (and `currentDateTime`) -- either way.

packages/palette/src/elements/ProgressBarTimer/ProgressBarTimer.tsx
```diff
@@ -16,7 +16,7 @@ export const ProgressBarTimer: React.SFC<{
   highlight = "purple100",
 }) => {
   const secondsRemaining = DateTime.fromISO(countdownEnd).diff(
-    DateTime.fromISO(currentTime),
+    DateTime.fromISO(currentTime.toString()),
```
damassi: Why was this change needed?

packages/palette/src/elements/TimeRemaining/TimeRemaining.ios.tsx
```diff
@@ -0,0 +1,77 @@
+import { DateTime, Duration } from "luxon"
```
damassi: Are there significant design differences between the web and ios implementations? Is it possible to share one component rather than duplicate functionality?

packages/palette/src/elements/TimeRemaining/TimeRemaining.ios.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const now = currentTime ? currentTime : DateTime.local()
+  const [duration, setDuration] = useState(
+    Duration.fromISO(
+      DateTime.fromISO(endDate)
+        .diff(now)
+        .toString()
+    )
+  )
+
+  useInterval(() => {
```
damassi: Why `useInterval` vs `setInterval`?

packages/palette/src/elements/TimeRemaining/TimeRemaining.tsx
```diff
@@ -1,39 +1,64 @@
...
+        ) : (
+          <>
+            {padWithZero(Math.max(0, Math.floor(duration.as("days"))))}d
+            {SEPARATOR}
+            {padWithZero(Math.max(0, Math.floor(duration.as("hours") % 24)))}h
+            {SEPARATOR}
+            {padWithZero(Math.max(0, Math.floor(duration.as("minutes") % 60)))}m
+            {SEPARATOR}
+            {padWithZero(Math.max(0, Math.floor(duration.as("seconds") % 60)))}s
+            {trailingText && ` ${trailingText}`}
```
damassi: The previous form, where the time values are assigned to variables is more readable  and also inspectable via debugger, vs depending on a react render cycle.

packages/palette/src/elements/TimeRemaining/TimeRemaining.ios.tsx
kierangillen: You mean why is it named that way or why don't I just use a regular setInterval?

damassi: > why don't I just use a regular setInterval
 ^

artsyit:

Built with commit 3fb0a2a942c448f86ba53fa69a4818fa5d89e370

https://deploy-preview-564--artsy-palette.netlify.com

packages/palette/src/elements/ProgressBarTimer/ProgressBarTimer.tsx
kierangillen: If `currentTime` is passed as a `DateTime` instead of a `string` the `DateTime.fromISO` wont be able to parse it unless it's converted to a string.

packages/palette/src/elements/TimeRemaining/TimeRemaining.ios.tsx
kierangillen: No there isn't, I should be able to move it into a shared file, they're identical.

packages/palette/src/elements/TimeRemaining/TimeRemaining.ios.tsx
kierangillen: the `useInterval` hook provides a state object that gets set every second which refreshes the render. Using a regular `setInterval` I'd need to add state to the component, the hook handles the render update and also cleans up the interval on unmount. Best in my opinion for one function to handle everything and keeps us from adding state, hooks and duplicating code.

damassi: nice 👍

packages/palette/src/elements/TimeRemaining/TimeRemaining.tsx
kierangillen: Ok switched back to the previous format

README.md
```diff
@@ -64,7 +64,7 @@ $ yarn start
 
 When adding a new component to Palette, it's important to be aware that this library is used on the web as well as in React Native, via Emission, and therefore must follow a few rules in terms of structure, namely:
 
-> Components exported from the main `/elements/index.tsx` must have a corresponding `.ios.tsx` file, even if the component isn't yet used in React Native.
+> If a mobile component has platform-specific features, that code has to live in a `Component.ios.tsx` file. Or if a web component has browser-only features, then a `Component.ios.tsx` file must be created, or React Native will error out. If the code between a web and native component is identical, a `.ios.tx` file isn’t needed. If some code can be shared between platforms, then that shared code should live in a `Component.shared.tsx` file.
```
pepopowitz: Some editing suggestions to give this paragraph a little more breathing room:

```suggestion
- If a mobile component has platform-specific features, that code must live in a `Component.ios.tsx` file.
- If a web component has browser-only features, then an empty `Component.ios.tsx` file must be created. Without this file, React Native will error out.
- If the code between a web and native component is identical, a `.ios.tx` file is not needed.
- If some code can be shared between platforms, then that shared code should live in a `Component.shared.tsx` file.
```

In the previous paragraph we say "must follow a few rules in terms of structure", so I think it makes sense to turn this into a list of rules instead of one paragraph.

packages/palette-docs/content/docs/components/TimeRemaining.mdx
kierangillen: Reverted this.

packages/palette/src/elements/ProgressBarTimer/ProgressBarTimer.tsx
pepopowitz: This seems like a good thing to wrap some tests around.

README.md
damassi: 👍

packages/palette/src/utils/useInterval.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef } from "react"
+
+/** useInterval */
+export function useInterval(callback, delay) {
+  const savedCallback = useRef(null)
+
+  // Remember the latest callback.
+  useEffect(() => {
```
pepopowitz: The code in this hook isn't totally obvious to me. Are there any tests we could put around this hook that could describe how these hooks are all intended to work together? Nothing that tests the exact implementation directly, but something that renders a test component that uses `useInterval`, and gets a callback fired the expected number of times within a window of time.

packages/palette/src/utils/useCurrentTime.ts
```diff
@@ -0,0 +1,20 @@
+import { DateTime } from "luxon"
+import { useState } from "react"
+import { useInterval } from "./useInterval"
+
+/**
+ * useCurrentTime
+ */
+export function useCurrentTime(currentTime) {
```
pepopowitz: Do the existing tests cover this hook pretty well? If not, I think tests against `useCurrentTime` directly would be valuable.

artsyit:
:rocket: PR was released in v4.14.48 :rocket:

PR #565, opened 2019-08-15
Small CSS adjustments to modal for scroll

kierangillen:
- Small css update to improve modal scroll

### DIFF
--- a/packages/palette/src/elements/Modal/Modal.tsx
+++ b/packages/palette/src/elements/Modal/Modal.tsx
@@ -234,7 +234,7 @@ const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   padding-bottom: ${space(2)}px;
   max-height: ${props =>
     !!props.FixedButton
-      ? `calc(100vh - (80px ${space(9)}px));`
+      ? `calc(100vh - (80px + ${space(9)}px));`
       : "calc(100vh - 80px)"};
   ${media.xs`
     max-height: ${props =>
--- a/packages/palette/src/elements/Modal/Modal.tsx
+++ b/packages/palette/src/elements/Modal/Modal.tsx
@@ -179,7 +179,7 @@ export const Modal: SFC<ModalProps> = ({
 }
 
 const FixedButtonWrapper = styled(Box)`
-  position: fixed;
+  position: relative;
   bottom: 0;
   left: 0;
   width: 100%;
@@ -213,7 +213,7 @@ const ModalElement = styled(AnimatedView)<TransitionElementProps>`
   top: 50%;
   left: 50%;
   height: ${props => (props.height ? props.height : "auto")};
-  max-height: calc(100% - 80px);
+  max-height: calc(100vh - 80px);
   min-height: 58px;
   overflow: hidden;
   background-color: ${color("white100")};
@@ -230,8 +230,10 @@ const ModalElement = styled(AnimatedView)<TransitionElementProps>`
 const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   height: 100%;
   width: 100%;
-  overflow: scroll;
-  padding-bottom: ${props => (!!props.FixedButton ? "112px" : space(2) + "px")};
+  max-height: ${props =>
+    !!props.FixedButton ? "calc(100vh - 172px)" : "calc(100vh - 80px)"};
+  overflow: auto;
+  padding-bottom: ${space(2)}px;
 `
 
 const CloseIconWrapper = styled(Box)`
--- a/packages/palette/src/elements/Modal/Modal.tsx
+++ b/packages/palette/src/elements/Modal/Modal.tsx
@@ -10,8 +10,6 @@ import { Flex } from "../Flex"
 import { Spacer } from "../Spacer"
 import { Serif } from "../Typography"
 
-const FIXED_BUTTON_HEIGHT = "92px"
-
 /**
  * refreshModalContentKey should change if the modal displays new content and should fade
  * in/fade out with content update. If refreshModalContentKey does not change, the content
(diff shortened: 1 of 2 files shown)


artsyit:

Built with commit 7650d039305295e39692964855bd58211d038b80

https://deploy-preview-565--artsy-palette.netlify.com

erikdstock:
Looks good from my browser in the deploy preview- scrollbar disappears on smaller content. There is a faint line above the action button if it is attached- I noticed this at the same time as the scrollbar began appearing, but I don't know if it should be a blocker to this fix.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -230,8 +230,10 @@ const ModalElement = styled(AnimatedView)<TransitionElementProps>`
 const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   height: 100%;
   width: 100%;
-  overflow: scroll;
-  padding-bottom: ${props => (!!props.FixedButton ? "112px" : space(2) + "px")};
+  max-height: ${props =>
+    !!props.FixedButton ? "calc(100vh - 172px)" : "calc(100vh - 80px)"};
```
damassi: this is so so minor, but `Boolean(props.FixedButton)` is more readable at a glance

kierangillen: fair, but `!!` is also ubiquitous.

kierangillen:
@erikdstock that line above the `FixedButton` is part of the design.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -213,7 +213,7 @@ const ModalElement = styled(AnimatedView)<TransitionElementProps>`
   top: 50%;
   left: 50%;
   height: ${props => (props.height ? props.height : "auto")};
-  max-height: calc(100% - 80px);
+  max-height: calc(100vh - 80px);
```
damassi: Ah, was thinking that this seemed  kinda weird no a large screen but checked zeplin and this is the spec 👍

kierangillen:
@erikdstock Okay just made a small update for the mobile styling. This is good to merge now.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -10,6 +10,8 @@ import { Flex } from "../Flex"
 import { Spacer } from "../Spacer"
 import { Serif } from "../Typography"
 
+const FIXED_BUTTON_HEIGHT = "92px"
```
damassi: Why not `space(9)`?

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -231,9 +233,17 @@ const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   height: 100%;
   width: 100%;
   max-height: ${props =>
-    !!props.FixedButton ? "calc(100vh - 172px)" : "calc(100vh - 80px)"};
+    !!props.FixedButton
+      ? "calc(100vh - (80px + " + FIXED_BUTTON_HEIGHT + "))"
+      : "calc(100vh - 80px)"};
   overflow: auto;
   padding-bottom: ${space(2)}px;
+  ${media.xs`
```
damassi: `max-heights`' should be grouped together, and `space(9)` with an inline comment is probably fine, don't need a big constant for less than three uses

packages/palette/src/elements/Modal/Modal.tsx
kierangillen: It's not a reference to a styled system spacing, but a specific pixel perfect measured height. Not a biggie but I would have opted for the specific height over the styled system height as the latter is used for keeping consistency, which was not the goal here. Updated using the `space` API as I'm mostly impartial.

packages/palette/src/elements/Modal/Modal.tsx
kierangillen: I think the constant helps identify why I'm adding the extra space here despite it being used sparingly.

packages/palette/src/elements/Modal/Modal.tsx
damassi: Makes sense; thinking here is that at scale, the less variables in the system (`space(9)` * n == 1 variable; `92px` == 1 variable) makes for more consistency visually (at scale). Since 92 is just two pixels away from our theme value, rounding down can be considered and effort towards the greater good.

dleve123:
x-posting so that issue doesn't get lost in the shuffle: https://artsy.slack.com/archives/C9XJKPY9W/p1565891762023900

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -230,8 +230,16 @@ const ModalElement = styled(AnimatedView)<TransitionElementProps>`
 const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   height: 100%;
   width: 100%;
-  overflow: scroll;
-  padding-bottom: ${props => (!!props.FixedButton ? "112px" : space(2) + "px")};
+  overflow: auto;
+  padding-bottom: ${space(2)}px;
+  max-height: ${props =>
+    !!props.FixedButton
+      ? "calc(100vh - (80px + " + space(9) + "))"
```
erikdstock:
```suggestion
      ? "calc(100vh - (80px + " + space(9) + "px" +  "))"
```

zephraph: I think it might be a little easier to grok if it was

```
`calc(100vh - (80px ${space(9)}px))`
```

You can nest <code>`</code> when it's inside an interpolation `${ ... }`

erikdstock: This fixes the issue that @dleve123 found/x-posted above (which only affects the modal when using the `FixedButton`)

kierangillen: Thanks for the catch, just updated.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -234,11 +234,11 @@ const ModalScrollContent = styled(Box)<ModalScrollContentProps>`
   padding-bottom: ${space(2)}px;
   max-height: ${props =>
     !!props.FixedButton
-      ? "calc(100vh - (80px + " + space(9) + "))"
+      ? `calc(100vh - (80px ${space(9)}px));`
       : "calc(100vh - 80px)"};
   ${media.xs`
     max-height: ${props =>
-      !!props.FixedButton ? "calc(100vh - " + space(9) + ")" : "100vh"};
+      !!props.FixedButton ? `calc(100vh - (80px ${space(9)}px));` : "100vh"};
```
erikdstock: We missed this in our conversations but there should be a `+` between `80px` and the interpolated `space` value.

erikdstock:
Checked in desktop and mobile, looks good with correct scrollbar behaviour; merging.

artsyit:
:rocket: PR was released in v4.14.47 :rocket:

PR #570, opened 2019-08-20
WIP: Add placeholder api to Select

erikdstock:
This pr implements select placeholder values based on the suggestion of @ds300 [here](https://github.com/artsy/reaction/pull/2708#issuecomment-522952081).

TODOS:
- [ ] Resolve a `Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.` I tried to do this but was unable to make it work without getting new warnings about controlled v uncontrolled components or breaking the input altogether.
- [ ] A few more tests. 

_Edit about this bit: The `color` property needs to be set on the `select` element, not the option, and then it works- but it does not change when a different option is selected. Need to investigate further._
I tried to make the placeholder text a lighter shade of gray by adding this style to the select container. It showed correctly in the devtools but the text color did not apply.
```
  option:disabled {
    color: ${color("purple100")};
  }
```
image

### DIFF
--- a/packages/palette/src/elements/Select/Select.test.tsx
+++ b/packages/palette/src/elements/Select/Select.test.tsx
@@ -1,6 +1,7 @@
 import { mount } from "enzyme"
 import React from "react"
 import { SelectSmall } from "../Select"
+import { LargeSelect } from "./Select"
 
 describe("Select", () => {
   describe("SelectSmall", () => {
@@ -25,11 +26,12 @@ describe("Select", () => {
 
     it("renders with a placeholder value", () => {
       const wrapper = mount(
-        <SelectSmall options={options} selected="secondOption" />
+        <LargeSelect options={options} placeholder="I'm a placeholder" />
       )
-
-      expect(wrapper.find("option").length).toBe(2)
-      expect(wrapper.find("select").props().value).toBe("secondOption")
+      expect(wrapper.find("select").text()).toMatch(
+        new RegExp("^I'm a placeholder")
+      )
+      expect(wrapper.find("option").length).toBe(3)
     })
 
     it("triggers callback on change", () => {
--- a/packages/palette/src/elements/Select/Select.test.tsx
+++ b/packages/palette/src/elements/Select/Select.test.tsx
@@ -23,6 +23,15 @@ describe("Select", () => {
       expect(wrapper.find("select").props().value).toBe("secondOption")
     })
 
+    it("renders with a placeholder value", () => {
+      const wrapper = mount(
+        <SelectSmall options={options} selected="secondOption" />
+      )
+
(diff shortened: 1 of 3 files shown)


artsyit:

Built with commit ec26fc5d6eb3b44178976fa465a949d46fd759b5

https://deploy-preview-570--artsy-palette.netlify.com

packages/palette/src/elements/Select/Select.test.tsx
```diff
@@ -1,19 +1,31 @@
...
+    },
+    {
+      text: "Second option that is really long",
+      value: "secondOption",
+    },
+  ]
+  describe("LargeSelect", () => {
+    it("renders with a placeholder value", () => {
+      const wrapper = mount(
+        <LargeSelect options={options} placeholder="I'm a placeholder" />
```
damassi: Why not set `placeholder` to just be `defaultValue`? We should be able to set it in this way anyways, and it will silence the warning from React.

packages/palette/src/elements/Select/Select.test.tsx
```diff
@@ -23,6 +35,16 @@ describe("Select", () => {
       expect(wrapper.find("select").props().value).toBe("secondOption")
     })
 
+    it("renders with a placeholder value", () => {
+      const wrapper = mount(
+        <SelectSmall options={options} placeholder="I'm a placeholder" />
+      )
+      expect(wrapper.find("select").text()).toMatch(
+        new RegExp("^I'm a placeholder")
```
damassi: Why use  a regex here?

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -34,6 +35,11 @@ export const LargeSelect: SFC<SelectProps> = props => {
         disabled={props.disabled}
         onChange={event => props.onSelect && props.onSelect(event.target.value)}
       >
+        {placeholderActive(props) && (
+          <option selected value="" disabled>
```
damassi: What happens if value is passed  in here _with_ `disabled`?

erikdstock: It starts out selected, but after you choose something you can't go back to the placeholder value because it is disabled as an option.

packages/palette/src/elements/Select/Select.test.tsx
erikdstock: I should probably do something better- this was just from copying the adjacent tests around.

damassi:
@erikdstock - what is the status on this? Still WIP or should this be closed?

erikdstock:
Thanks for the poke @damassi - closing for now, but we should revisit this. cc @ds300 @dblandin

PR #571, opened 2019-08-21
Add Modal component examples to Storybook

dblandin:
This PR ports the Palette documentation examples over to Storybook for visual regression testing.

screenshot 1566399731

### DIFF
--- a/packages/palette/.storybook/config.js
+++ b/packages/palette/.storybook/config.js
@@ -2,6 +2,8 @@ import React from "react"
 import { configure, addDecorator, addParameters } from "@storybook/react"
 import { create } from "@storybook/theming"
 import { Theme } from "../src/Theme"
+import { injectGlobalStyles } from "../src/helpers/injectGlobalStyles"
+
 import "storybook-chromatic"
 
 // automatically import all files ending in *.story.tsx.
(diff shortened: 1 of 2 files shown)


dblandin:
Should Storybook / doc updates have a `Version: Trivial` label? Seems like it shouldn't cut a new a release...

Also ran into a bit of weirdness around the rendering in Storybook which you can see in the fixed button example:

screenshot 1566399835

Didn't have much luck fixing it. Maybe @kierangillen and I can pair on that.

artsyit:

Built with commit 0acda5b79c8e42283d1aa010e34772be6d6aff3c

https://deploy-preview-571--artsy-palette.netlify.com

damassi:
@dblandin , @kierangillen - something we need to do is wire up a similar mechanism to [storiesOf](https://github.com/artsy/reaction/blob/master/src/__stories__/storiesOf.js#L11-L15) from reaction. There's no theme there by default, hence the styling problems. 

The other way to do it would be to use 

```
.addDecoorator(storyFn => <Theme>{storyFn()}</Theme>
```
directly in the story.

dblandin:
Addressed with https://github.com/artsy/palette/pull/571/commits/173cd44a6e8586d4b4a628d8628afb9b38b0d136!

We already had a decorator defined that wrapped the story with a `<Theme>` component but it looks like we also needed to pull in Palette's global styles. Let me know if there's a better way of going about this.

Modal examples look better though :+1: 

screenshot 1566409562

Issue #576, opened 2019-09-05
[NEW] Elements / Loaders / PageLoader


samchieng:
Full page loading feedback

Design Spec:
https://zpl.io/2ZXj3kL

PR #577, opened 2019-09-06
[PageLoader] Add PageLoader component

damassi:
Addresses https://github.com/artsy/palette/issues/576

Adds a new `<PageLoader>` component: 

Screen Shot 2019-09-05 at 5 11 09 PM

Additionally, adds a StoryBooks entry for `ProgressBar` and a new prop `showBackground` for showing / hiding the background that the progressBar sits upon. 

Note that the actually loading aspect is faked -- we don't when we'll get a response back from the server, but there's a nice trig curve to make it feel natural: 

loader


artsyit:

Built with commit 890917a0eceb6f52083ddcda0adfb94e14c56f70

https://deploy-preview-577--artsy-palette.netlify.com

packages/palette/src/elements/PageLoader/PageLoader.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  defaultProps = {
+    showBackground: true,
+  }
+
+  currentProgress = this.props.percentComplete || 0
+  step = 0.5
+  interval
+
+  componentDidMount() {
```
kierangillen: This looks great. I’m thinking we should consider a 200-400ms delay before showing the loader. Personally I think it’s best if the progress bar isn’t displayed on every page load but only those that aren’t snappy. What do you think?

kierangillen: Or maybe this interval achieves this but was this a consideration?

damassi: Good idea -- I was going to add a light fade in when it appears so I'll work it in then 👍

kierangillen:
#mergeongreen

artsyit:
:rocket: PR was released in v4.17.0 :rocket:

PR #597, opened 2019-10-01
Add dialog to palette

oxaudo:
We've used the `ModalDialog` from `Reaction` in several places in `Volt` and have more tasks that would rely on `Dialog`. And we've discussed that it would be more clean/proper/supported to move the `ModalDialog` into palette and refactor the Volt usages of it as a first step.

I've noticed that in `Reaction` both `Modal` and `ModalDialog` relied on multiple shared wrapper classes and when `Modal` implementation was moved to `Palette` all of those classes were inlined into `Modal` implementation(https://github.com/artsy/palette/pull/540).

Also looking into specs for both seems like `Modal` defines mostly a wrapper (https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5c75ad115c1db5628cc03c2a) and the definition for `Dialog` is basically an instance of `Modal` with some predefined properties and very defined content.

So this PR tries to create `Dialog` with some very specific properties but the implementation is just calling Modal with a very specific content.

Want to ask your opinion @damassi, @kierangillen, @ds300 and @pepopowitz on what do you think of this approach? We might need to add additional width to the `Modal` options since spec defines `Dialog` in different width variation. Also I need to check with @tinn if it's ok for us to have a bit more margin on `Dialog` so it's matching spaces on `Modal`. Another task here is to fill out specs some more.

I'll proceed with all those tasks if we confirm that it's a proper approach here.

Thanks all!


oxaudo:
Here is the comparison. 

Current use:
Screen Shot 2019-10-01 at 2 53 51 PM

Same thing configured with the WIP version here:
Screen Shot 2019-10-01 at 2 56 36 PM

artsyit:

Built with commit bdd387d96a94722ad39c1cf0d33844491a46a0ec

https://deploy-preview-597--artsy-palette.netlify.com

packages/palette/src/elements/Dialog/Dialog.story.tsx
```diff
@@ -0,0 +1,59 @@
...
+      )
+    },
+    { chromatic: { delay: 500, diffThreshold: 0.2 } }
+  )
+  .add(
+    "With detail",
+    () => {
+      return (
+        <Dialog
+          show
```
damassi: `show` shouldn't be required, but rather default to true with the option to set it to false

packages/palette/src/elements/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Dialog.
+ * Spec: https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5b9a7bdaa2fa5551fff7c014
+ */
+export const Dialog: SFC<DialogProps> = ({
+  title,
+  detail,
+  show,
```
damassi: `show = true`

tinn:
@oxaudo Thanks for showing the comparison! I understand that we want the margin to match, but the actions appear too high up with 30px for the width that we're aiming for. Keeping the 20px margin would feel more balanced; however, I'm okay with it with this larger margin if we think it's the better, more consistent path.

A bigger priority is swapping the actions so that the dismissive action (Cancel) is to the left of the confirmation action (Delete).

packages/palette/src/elements/Dialog/Dialog.story.tsx
```diff
@@ -0,0 +1,59 @@
...
+          title="Information"
+          detail="This extra informaton is important."
+          primaryCta={{
+            action: () => ({}),
```
damassi: Same here -- this shouldn't be required so as to keep the API requirements slender, and instead the ` undefined` action handled inside the implementation.

packages/palette/src/elements/Dialog/__tests__/Dialog.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+          action: () => ({ primaryCtaAction }),
+          text: "Continue",
+        }}
+        title="Condimentum Ridiculus"
+        show
+      />
+    )
+    expect(component.html()).toContain("Condimentum Ridiculus")
+  })
+})
```
damassi: Can we add a few more tests that cover all of the different behavior? 
- `primaryCta` / secondary actions and labels
- showing / hiding 
- dismissal

oxaudo:
@tinn - the buttons words/actions are configurable. What it set is that there are at least one button and max of 2.

And the padding - we can make it configurable as well, it's just a bit more work -:)

Will add it in a bit as well if you think it's a distraction to the experience.

packages/palette/src/elements/Dialog/__tests__/Dialog.test.tsx
oxaudo: @damassi - totally. I mainly PR'ed this in current state to get some feedback on the idea of reusing `Modal` with a very predefined values for `Dialog` implementation. Was not sure if it'll be better to go this way or try to abstract needed classes and then include them in both `Modal` and `Dialog` implementation as it was done in `Reaction`. 

I think the code looks simpler with `Dialog` using `Modal` directly and don't see any major drawbacks with  going this way, but maybe I'm missing something here?

damassi: No this is great! The design looks super solid / ideal 👍

packages/palette/src/elements/Dialog/Dialog.story.tsx
oxaudo: I think the idea here is that at least one button should be required for a `Dialog`. We can default the text of it to something generic like `ok` and action is to just close the `dialog` but that is one of the distinctions of a `Dialog` if I understand it correctly. `Dialog` is something that blocks the flow with at least one action that should be actively acknowledged (clicked) to move with the flow.

tinn: Yeah! A dialog presents the user critical information that require their attention/make a decision–purposely interruptive of their flow.

tinn:
> @tinn - the buttons words/actions are configurable. What it set is that there are at least one button and max of 2.
> 
> And the padding - we can make it configurable as well, it's just a bit more work -:)
> 
> Will add it in a bit as well if you think it's a distraction to the experience.

Great. Thanks for the clarification @oxaudo I don't think it's a critical thing (distractive to the UX) for now, but I can see us addressing it later in the future as a refinement task. Perhaps this is a case where we move forward with this current padding now and bring it up during a design systems meeting.

jonallured:
Yeah, what @damassi said: this approach seems ideal and makes total sense - thanks for taking on this work!! ❤️

oxaudo:
Ok @jonallured and @damassi - I've updated the Width of the `Dialog` and added some specs. I think ideally `Modal` could have taken just `modalWidth` property that can be either `wide`, `narrow` or `normal`. But I did not want to break/go and update the current usages of modal, so just added `modalWidth` as additional property that would overwrite `isWide` if present. Hope that makes sense?

Also filled out additional specs.

And here is the latest look (cc @tinn ):
Screen Shot 2019-10-02 at 3 10 40 PM

packages/palette-docs/content/docs/elements/dialogs/Dialog.mdx
```diff
@@ -0,0 +1,98 @@
+---
+name: Dialog
+status: wip
```
damassi: `wip` can be removed

packages/palette-docs/content/docs/elements/dialogs/Dialog.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+Dialogs appear over content and disable all other functionality. The user must
+take an action within the dialog before they can continue.
+
+Dialogs must include at least one action button and no more than two.
+
+Dialogs appear at the centerpoint of the screen.
+
+
+## Dialog with minimal text and one continue action
```
damassi: linebreak

packages/palette-docs/content/docs/elements/dialogs/Dialog.mdx
```diff
@@ -0,0 +1,98 @@
...
+            text: "Continue",
+          }}
+        />
+      </>
+    )}
+  </State>
+</Playground>
+
+## Dialog with some details about the action
+<Playground title="Dialog with details">
```
damassi: same

oxaudo:
@damassi - those new chromatic files are mine, but not sure that is causing the other changes there(the loader and css grid). Any idea?

packages/palette/src/elements/Dialog/Dialog.story.tsx
```diff
@@ -0,0 +1,56 @@
...
+            text: "Continue",
+          }}
+          secondaryCta={{
+            action: () => ({}),
+            text: "Cancel",
+          }}
+        />
+      )
+    },
+    { chromatic: { delay: 500, diffThreshold: 0.2 } }
```
damassi: These items can be removed

damassi:
Chromatic is finicky. I've been playing with things like 
```
{ chromatic: { delay: 500, diffThreshold: 0.2 } }
```
but that doesn't seem to help. Since we know that nothing has changed in those files feel free to accept changes.

tinn:
@oxaudo Looks great!

packages/palette/src/elements/Dialog/Dialog.tsx
jonallured: This is done, marking resolved.

artsyit:
:rocket: PR was released in v4.17.11 :rocket:

PR #600, opened 2019-10-03
Improve selects

jonallured:
While working on something unrelated over on Volt, I had a chance to compare the `Select` component here with what we made before Palette even existed! I found that there was some work to do, so this PR is me taking a crack at it. The result of this PR would be me able to migrate code from Volt's select to Palette's.

So with this LargeSelect learns all sorts of cool new tricks:

* it supports title, description and error messages like Input
* it correctly computes it's border color the same way as Input
* it can be marked as required
* a name attribute can be supplied to the underlying select element
* the keys for the options try a little harder to be unique on the page

I've also updated the docs to more closely match Input so that all the various states can be played around with.

Here's some pretty pictures:

Screen Shot 2019-10-03 at 3 58 37 PM
Screen Shot 2019-10-03 at 3 58 40 PM

### DIFF
--- a/packages/palette/src/elements/Input/Input.test.tsx
+++ b/packages/palette/src/elements/Input/Input.test.tsx
@@ -49,98 +49,52 @@ describe("Input", () => {
 
 describe("computeBorderColor", () => {
   it("defaults to returning black 10", () => {
-    const props = {
-      disabled: null,
-      error: null,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(null,  null)
     expect(color).toEqual("black10")
   })
 
   it("returns black10 when disabled", () => {
-    const props = {
-      disabled: true,
-      error: false,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(true, false)
     expect(color).toEqual("black10")
   })
 
   it("returns black10 when disabled and error", () => {
-    const props = {
-      disabled: true,
-      error: true,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(true, true)
     expect(color).toEqual("black10")
   })
 
   it("returns black10 when disabled and hover", () => {
-    const props = {
-      disabled: true,
-      error: false,
-      pseudo: "hover",
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(true, false, "hover")
     expect(color).toEqual("black10")
   })
 
   it("returns black10 when disabled and focus", () => {
-    const props = {
-      disabled: true,
-      error: false,
-      pseudo: "focus",
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(true, false, "focus")
     expect(color).toEqual("black10")
   })
 
   it("returns red100 when error", () => {
-    const props = {
-      disabled: false,
-      error: true,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(false, true)
     expect(color).toEqual("red100")
   })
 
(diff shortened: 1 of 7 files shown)


damassi:
@jonallured can you update the description with screenshots?

jonallured:
@damassi you got it dude!

packages/palette/src/elements/Input/Input.test.tsx
```diff
@@ -49,98 +49,52 @@ describe("Input", () => {
 
 describe("computeBorderColor", () => {
   it("defaults to returning black 10", () => {
-    const props = {
-      disabled: null,
-      error: null,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(null, null)
```
damassi: Using unnamed arguments is a bit opaque -- can we preserve passing an object along?

packages/palette/src/elements/Input/Input.test.tsx
```diff
@@ -49,98 +49,52 @@ describe("Input", () => {
...
     expect(color).toEqual("black10")
   })
 
   it("returns black10 when disabled", () => {
-    const props = {
-      disabled: true,
-      error: false,
-    }
-    const color = computeBorderColor(props)
+    const color = computeBorderColor(true, false)
```
damassi: Meaning, what is "true", what is "false"? Objects with keys also allow for order independence.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -55,7 +56,11 @@ interface StyledInputProps extends React.HTMLProps<HTMLInputElement> {
 /**
  * func to compute border color
  */
-export const computeBorderColor = ({ disabled, error, pseudo = null }) => {
+export const computeBorderColor = (
+  disabled: boolean,
+  error: boolean,
+  pseudo?: string
+): Color => {
```
damassi:
```js 
export const computeBorderColor: {
  disabled: boolean,
  error: boolean,
  pseudo?: string
} = ({ 
  disabled, 
  error,
  pseudo
}) => {
  ... 
}
```
this also follows closely with the react props model, which these days is the dominant pattern for working with functions.

artsyit:

Built with commit 9f6bae91f247b1a62dca0f120f5a315fe8e63933

https://deploy-preview-600--artsy-palette.netlify.com

packages/palette/src/elements/Input/Input.tsx
jonallured: I opted for the interface, good feedback!!! 🤗

jonallured:
@damassi wow, great catch on that arrow being weird. Turns out I needed another container to keep the layout the same. I think I've got all your feedback addressed but lmk if I missed something!

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -52,10 +53,18 @@ interface StyledInputProps extends React.HTMLProps<HTMLInputElement> {
...
+  error: boolean
+  pseudo?: string
+}
+
 /**
  * func to compute border color
  */
-export const computeBorderColor = ({ disabled, error, pseudo = null }) => {
+export const computeBorderColor = (inputStatus: InputStatus): Color => {
+  const { disabled, error, pseudo } = inputStatus
```
damassi: 👍

damassi:
Mind accepting / adding changes in Chromatic? Then feel free to merge on 🍏

jonallured:
@damassi just went through Chromatic for the first time, that was really cool!! ❤️

artsyit:
:rocket: PR was released in v4.17.13 :rocket:

PR #601, opened 2019-10-04
bug fix for Dialog

oxaudo:
Realizing that I have a flaw here with the dialog. Because it's a stopping point in the flow that usually used for 'confirm'/'cancel' interactions - with confirm being primary action and cancel secondary - we definitely don't want to call primary action with dialog close and in most cases we would not want to do any additional actions on close.

I should have a spec for that but not sure how to incorporate `State` for show in spec.

### DIFF
--- a/packages/palette/src/elements/Dialog/Dialog.tsx
+++ b/packages/palette/src/elements/Dialog/Dialog.tsx
@@ -14,6 +14,7 @@ interface DialogProps {
   detail?: React.ReactNode
   primaryCta: CtaProps
   secondaryCta?: CtaProps
+  onClose?: () => void
   show?: boolean
   title: string
 }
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 26d3a9be0474980bf6016fcadde0698a657ac4dd

https://deploy-preview-601--artsy-palette.netlify.com

jonallured:
Oh nice catch! Yeah I was trying to come up with a good way to test this, but I'm not seeing it. Since the `hideCloseButton` prop is being passed to `Modal` there's not really a good way to trigger closing without clicking a button, right? I'm going to merge this and if we think of a better way to test it, we can always follow up with a test.

artsyit:
:rocket: PR was released in v4.17.13 :rocket:

PR #604, opened 2019-10-11
PURCHASE-1518: Update Bell and BellFill Icons in Palette

lilyfromseattle:
# [PURCHASE-1518] Update Bell and BellFill Icons in Palette

Adjusted the height by a few pixels on the Bell and BellFill icons because they were too tall compared to the other icons on the iOS Artwork Page.

__Before:__
Screen Shot 2019-10-11 at 3 15 04 PM

__After:__
Screen Shot 2019-10-15 at 2 15 28 PM

[PURCHASE-1518]: https://artsyproduct.atlassian.net/browse/PURCHASE-1518

### DIFF
--- a/packages/palette/src/svgs/BellIcon.ios.tsx
+++ b/packages/palette/src/svgs/BellIcon.ios.tsx
@@ -6,7 +6,7 @@ import { Icon, IconProps } from "./Icon"
 /** BellIcon */
 export const BellIcon: React.SFC<IconProps> = props => {
   return (
-    <Icon {...props} viewBox="0 0 18 18">
+    <Icon {...props} width="14px" height="14px" viewBox="0 0 18 18">
       <Path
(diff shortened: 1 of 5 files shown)


artsyit:

Built with commit 4f884c3344bf51a6a63ac4c38175ce608824cb6e

https://deploy-preview-604--artsy-palette.netlify.com

ds300:
:ceiling-cat: I think we should be adjusting the SVG path code, rather than the default width/height props. We should be aiming for a situation where if you put all the icons in a row and give them all the same explicit width/height they should look consistently sized.

Looks like there are some svgs attached to that jira ticket, they might contain the updated paths.

lilyfromseattle:
I updated the paths and changed the after screenshot. Should be good to go now.

packages/palette/src/svgs/BellIcon.ios.tsx
```diff
@@ -6,7 +6,7 @@ import { Icon, IconProps } from "./Icon"
 /** BellIcon */
 export const BellIcon: React.SFC<IconProps> = props => {
   return (
-    <Icon {...props} viewBox="0 0 18 18">
+    <Icon {...props} width="14px" height="14px" viewBox="0 0 18 18">
```
sweir27: @lilyfromseattle I would expect to see some changes in these files in the `d=` section (where the path is defined).

I don't think we need to update the width/height here-- that can be done from clients who consume this icon.

PR #608, opened 2019-10-23
[PURCHASE-1577] Tag component

sepans:
[PURCHASE-1577](https://artsyproduct.atlassian.net/browse/PURCHASE-1577)

[Design](https://app.zeplin.io/project/5acd19ff49a1429169c3128b/screen/5d96541467c03315a0266529)

TODO:
- [x] Fix mdx file


artsyit:

Built with commit 365609115d062944d34ccdbeb68e152dcf434987

https://deploy-preview-608--artsy-palette.netlify.com

samchieng:
@sepans ooooo yes! i think the only thing missing is there should be a `2px` border-radius for both tags

sepans:
🙏 thanks @samchieng totally missed that.

zephraph:
This is 👍. 

A thought for the future: should there be a default limit on the number of tags to display?

I think we can merge without that today though.

artsyit:
:rocket: PR was released in v5.1.0 :rocket:

samchieng:
@zephraph prob depends on the context, but i can't imagine a situation where we'd ever show more than up to 8 by default, so if its helpful we can use this number for now.

Just thinking on the Artwork page, we'll probably only get 6 max

PR #609, opened 2019-10-24
Add maxWidth to selects

mdole:
Pretty much title. We want to have the ability to set maxWidth for some work on the /artworks page in Volt.

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -3,6 +3,8 @@ import styled, { css } from "styled-components"
 import {
   maxWidth,
   MaxWidthProps,
+  minWidth,
+  MinWidthProps,
   PositionProps,
   space as styledSpace,
   SpaceProps,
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit a01b33566aafdfd38e6a05a012d01b375c02b5ef

https://deploy-preview-609--artsy-palette.netlify.com

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { SFC } from "react"
 import styled, { css } from "styled-components"
 import {
+  maxWidth,
+  MaxWidthProps,
```
oxaudo: I think we should have an option to set both `maxWidth` and `minWidth` here. And let clients figure out which one they want.

oxaudo:
@mdole - also bunch of those chromatic changes seem unrelated. I do still scan and approve them usually just to make sure that nothing looks completely off.

There might be a better process with this though. Maybe @damassi or @zephraph can fill us in on the latest procedure with those.

packages/palette/src/elements/Select/Select.tsx
mdole: makes sense! done

artsyit:
:rocket: PR was released in v5.1.1 :rocket:

PR #610, opened 2019-10-28
Adds Tags container animation

sepans:
Fixes [PURCHASE-1615](https://artsyproduct.atlassian.net/browse/PURCHASE-1615)

Mostly inspired from [Collapse component](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Collapse/Collapse.tsx) H/T @ds300

I first struggled to find a way to animate the height of a flexbox that has `flex-wrap` to wrap its element using only css but looked at the collapse component and realized that approach is perfectly applicable here but open to other simpler ideas.

Also I don't think this causes any problem with **SSR** since the height is `auto` by default and only changes when there is a user interaction but I might be missing something.

<details>
  <summary>screenshot</summary>

tags-animation

</details>

TODO: 
- [x] Add `.ios.tsx` file (in https://github.com/artsy/palette/commit/ddf783a9bde4f4c6bc650a30f5d95c0534450213)

### DIFF
--- a/packages/palette/src/elements/Tags/Tags.tsx
+++ b/packages/palette/src/elements/Tags/Tags.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useEffect, useRef, useState } from "react"
 import { color } from "../../helpers"
 import { styledWrapper } from "../../platform/primitives"
 import { Box } from "../Box"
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Tags/Tags.tsx
```diff
@@ -24,28 +24,73 @@ export const Tags: React.FC<TagsProps> = ({
   displayNum = tags.length,
 }) => {
   const [expanded, setExpanded] = useState(false)
+  const [boxHeight, setBoxHeight] = useState("auto")
+  const flexContainer = useRef<HTMLDivElement>(null)
+  const animatingBox = useRef<HTMLDivElement>(null)
+  // after animating AnimatingBox height set back height to auto (to handle screen resize etc.)
+  useEffect(() => {
+    animatingBox.current.addEventListener("transitionend", e => {
+      const isAnimatingBoxTransition = Array.from(
```
sepans: The transitions from `Tag` elements bubble up here so had to check if the `transitionend` is from the `AnimatingBox` element.

packages/palette/src/elements/Tags/Tags.tsx
```diff
@@ -24,28 +24,73 @@ export const Tags: React.FC<TagsProps> = ({
...
   })
+
+  /**
+   * By default AnimationBox height is auto. for open/close transition:
+   * 1. explicitly set the height to be the same as the flex container inside it,
+   * 2. display new tags (setExpanded(true))
+   * 3. set the height to the new flexContainer height so the animation triggers
+   */
+  const toggleMore = () => {
+    const BORDER_OFFSET = 2
```
sepans: Without this 2 px `BORDER_OFFSET` the container height jumps a little bit when height is changed from px value to `auto` which I guessed it was because of top/bottom borders (?!).

artsyit:

Built with commit 2ebc30beb7a92d83dd1e03f65f737ab1069e0403

https://deploy-preview-610--artsy-palette.netlify.com

damassi:
@sepans - why not use the collapse component directly?

packages/palette/src/elements/Tags/Tags.tsx
```diff
@@ -24,28 +24,73 @@ export const Tags: React.FC<TagsProps> = ({
...
+    setBoxHeight(oldHeight + "px")
+
+    setExpanded(!expanded)
+
+    // wait for a tick
+    setTimeout(() => {
+      const newHeight = flexContainer.current.offsetHeight
+        ? flexContainer.current.offsetHeight - BORDER_OFFSET
+        : "auto"
+      setBoxHeight(`${newHeight}px`)
```
ds300: `'autopx'`? I've got a feeling `flexContainer.current.offsetHeight` will always be truthy here and above  🤔

ds300:
@damassi I think because collapse only animates between `height: auto` and `height: 0px` while this needs to animate between two different states of `height: auto`. Might be possible to combine the logic for both.

sepans:
@damassi because the `+ More` button could be in the middle of the row depending on the width of the parent, size of tags and [number](https://github.com/artsy/palette/pull/610/files#diff-62f69d59656282ca33c9a0215bc6f41dR17) of tags to be displayed, but `Collapse` is a div so new tags appear in a new line if put inside `Collapse` and not possible to preserve the `flex-wrap` behavior.

damassi:
Would be cool if we could maybe add a new prop to `Collapse` to handle both cases (since in both cases we're "collapsing" content in a similar way) but either way makes sense 👍

packages/palette/src/elements/Tags/Tags.tsx
sepans: fixed in https://github.com/artsy/palette/pull/610/commits/ddf783a9bde4f4c6bc650a30f5d95c0534450213

sepans:
@damassi yes it would be dryer using `Collapse` but I feel making collapse component support that case will make it too complex as opposed to splitting the complexity between the two which makes it them easier to maintain. But if you and @ds300 think otherwise would be happy to try that out.

packages/palette/src/elements/Tags/Tags.tsx
ds300: I think you could probably do

```ts
const isAnimatingBoxTransition = e.target === animatingBox.current
```

artsyit:
:rocket: PR was released in v5.1.5 :rocket:

PR #611, opened 2019-10-29
Make Tags cross platform

ds300:
The introduction of the `Tags` component broke emission because it was using the web version of `styled` so I switched it to use our cross-platform `styledWrapper` helper.

### DIFF
--- a/packages/palette/src/elements/Tags/Tags.tsx
+++ b/packages/palette/src/elements/Tags/Tags.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react"
-import styled from "styled-components"
 import { color } from "../../helpers"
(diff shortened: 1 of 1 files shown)


artsyit:

Built with commit 9bf32fc26784bb6a9796590694f791c917e1f26c

https://deploy-preview-611--artsy-palette.netlify.com

sepans:
Cool. Thanks. Didn't know about `styledWrapper` 🚀

ds300:
@sepans 

> should I just add the .ios.tsx file?

Yeah looks like that would be the best thing to do. Might be tricky to add support for the collapsing content. Might be able to combine that with the collapsing ReadMore logic but for now happy to just have a blank .ios.tsx file

PR #613, opened 2019-11-06
want the Box to support minWidth

oxaudo:
I need it for volt layout. In general I think it's good if we try to stay consistent with supporting min/max if we support at least one. Takes some time to figure out why it's not taking max if it's taking min -:)


artsyit:

Built with commit 7de5df2f18ad25311f6ceafd157147a9fb8ff9db

https://deploy-preview-613--artsy-palette.netlify.com

artsyit:
:rocket: PR was released in v5.1.6 :rocket:

PR #614, opened 2019-11-08
[Don't merge] add id to select interface

mdole:
Not 100% sure this is the right direction yet but putting it up so that I can reference it from the relevant Volt PR :)


artsyit:

Built with commit 99593567e12466b677cfa5297a4c8b57cbe5dc73

https://deploy-preview-614--artsy-palette.netlify.com

mdole:
Closing since we decided to use `name` instead of `id` in Volt!

PR #629, opened 2020-02-05
[Docs] Add gatsby-plugin-s3 for deploying app to S3 bucket

damassi:
We're leaving Netlify and instead leaning on our own S3 bucket. This adds a new build step (`yarn deploy-docs`) which will build and upload Palette's docs to the `artsy-palette` bucket. We use a [Gatsby Plugin](https://www.gatsbyjs.org/docs/deploying-to-s3-cloudfront/) to accomplish this. 

Still need to wire up to CI in a separate PR as well as wire up subdomain, but in the meantime we can view Palette via underlying URL here:  http://artsy-palette.s3-website-us-east-1.amazonaws.com/

**Follow-up Tasks:**
- Setup CI for deploy's on merge from master
- Remove Netlify app access from Palette repo


package.json
```diff
@@ -2,13 +2,14 @@
   "name": "palette",
   "private": true,
   "scripts": {
+    "build-docs": "yarn compile-palette && yarn workspace artsy-palette-docs build",
+    "compile-palette": "yarn workspace @artsy/palette compile",
+    "deploy-docs": "yarn workspace artsy-palette-docs deploy",
```
damassi: Main deploy script

packages/palette-docs/gatsby-config.js
```diff
@@ -58,6 +58,12 @@ module.exports = {
         manualInit: true,
       },
     },
+    {
+      resolve: `gatsby-plugin-s3`,
+      options: {
+        bucketName: "artsy-palette",
```
anandaroop: What I've previously seen for S3 based hosting is that [the bucket name has to match the hostname](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingCustomURLs) of the site being served — so we might need `palette.artsy.net` as the name of the bucket.

On the other hand, I'm not sure if that's required if we make a CloudFront distribution out of this bucket — something we might consider in order to get features like https, caching etc.

damassi: Yup, we'll be putting this behind cloudfront 👍

zephraph:
I think this solution is good to roll with, but I'd eventually like to pull this into our circle orbs config to have a process we could also use for future static sites (and reaction's storybook site)

There's already an orb specifically for S3: https://circleci.com/orbs/registry/orb/circleci/aws-s3

I don't think that needs to be a blocker or done in this PR though generally given that we would like this back up sooner rather than later.

damassi:
@zephraph - I assigned to you because I was thinking you might have something up your sleeve. This plugin seemed easy enough but if we have an orb (and we still _do_ need to update Reaction) then it would be better to be consistent.

zephraph:
I say we wire up CI and ship this as the first version. Can set up reaction with the circle plugin and loop back and update this later.

damassi:
Sounds good 👍

artsyit:

PR #631, opened 2020-02-11
Deploy docs to s3 on master build

zephraph:
This is a follow-up to https://github.com/artsy/palette/pull/629

cc @damassi 

This sets up CI to deploy the docs (assuming that the `auto-release` set succeeds). 

I'm assuming we need to add some env vars to this as well?

## TODO

- [x] Add credentials for pushing to s3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,7 @@ workflows:
       - yarn/run:
           name: deploy-docs
           script: deploy-docs
+          context: hokusai
           requires:
             - yarn/auto-release
           filters:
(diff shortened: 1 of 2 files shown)


damassi:
Yup, need envs -- gatsby plugin automatically looks for that file I believe. Locally, I grabbed the values from force via hokusai.

zephraph:
@eessex we need to have some aws creds for the upload. Do you think we should reuse the hokusai context or should we create special credentials just for this project?

eessex:
Yes, we can use the hokusai context to get credentials!

zephraph:
Okay, this is good to go.

artsyit:

Issue #633, opened 2020-02-12
Implement Algolia Search


damassi:
From Algolia customer support:

--- 
	
Congratulations, your search is now ready!
I've successfully configured the underlying crawler and it will now run every 24h.

You're now a few steps away from having it working on your website:

- Copy the following CSS/JS snippets and add them to your page
```
<!-- at the end of the HEAD -->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
```

```
<!-- at the end of the BODY -->
<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
<script type="text/javascript"> docsearch({
apiKey: 'e4ea4437446d07b0549e0db7928d92d1',
indexName: 'artsy_palette',
inputSelector: '### REPLACE ME ####',
debug: false // Set debug to true if you want to inspect the dropdown
});
</script>
```

- Add a search input in your page if you don't have any yet. Then update the inputSelector value in JS snippet
to a CSS selector that targets your search input field.
- Optionally customize the look and feel by following the DocSearch documentation
(See https://community.algolia.com/docsearch/styling.html)
- You can also check your configuration in our GitHub repository
(See https://github.com/algolia/docsearch-configs/blob/master/configs/artsy_palette.json).
- < @zephraph email address > can get access to the full Algolia analytics for your DocSearch index by creating an account, following this link: https://www.algolia.com/users/invitation/accept?invitation_token=XtnS5MqprMJxDEPaEY3G

Please open a pull request if want to leverage your configuration!

Feel free to get back to us if you have any issues or questions regarding the integration.

We'd also be happy to get your feedback and thoughts about DocSearch - so we can continue to improve it.

Have a nice day :)

-----

@zephraph - would you mind logging into Algolia admin with your email (per the above) and updating to artsy@it?

zephraph:
@damassi, done.

damassi:
Updated 1pass?

pepopowitz:
[...]

PR #642, opened 2020-02-21
Incorporate Algolia-based search

pepopowitz:
Addresses #633 

This PR implements a local plugin hooking up Algolia docsearch. It is based on [an external plugin](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch) that didn't quite work for our scenario. (Specifically, the search box only worked on initial page render, due to [a known issue in that plugin](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/issues/2).) I followed the lead of [a comment on that issue](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/issues/2#issuecomment-566519493), but moved all functionality into a local plugin instead of using the external one. We were only using about 12 lines of the external version, and about 30 lines were not working for us, so it felt wrong to keep using the external one.

## In action

palette-search

### Things to note in the gif

* It takes you to the prod site when clicking on a result locally, I'm assuming because the URLs from algolia include the origin. A minor annoyance that I don't think is worth finding a way around, but let me know if you feel otherwise.
* The target URL includes `#__gatsby` at the end, again because it's in the Algolia URL. I'm not sure how it ends up there or how to get rid of it? I _think_ it's probably just another minor annoyance.

### Original PR body

~This PR adds [a plugin for hooking up Algolia docsearch](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch). Unfortunately, the plugin doesn't work very well for our scenario - it only connects the search box to Algolia on initial page render. As soon as you start clicking around the site, the search box becomes functionless. [This is a known issue in that plugin](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/issues/2). It is a result of our search box re-rendering at each new route, something the plugin doesn't account for. I dug a bit to try to fix it within the plugin so I could submit a PR, but I was unsuccessful.~

~Instead, I followed the lead of [a comment on that issue](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/issues/2#issuecomment-566519493). I'm still using the plugin to render the [JS](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/blob/9d6cbc5a627ee3687222422cae5e2c7c0cf6c924/src/gatsby-ssr.js#L21) and [CSS](https://github.com/ironstar-io/gatsby-plugin-algolia-docsearch/blob/9d6cbc5a627ee3687222422cae5e2c7c0cf6c924/src/gatsby-ssr.js#L13) inclusions into the DOM, but connecting the actual search box with an effect in the `SearchBox` component.~

~As a result, we're only using about 12 lines of code from the plugin. In addition, we _have_ to provide settings for the plugin to work, but they aren't actually used by the plugin (or the workaround - I don't _think_ I can access the options for another plugin.) My preference at this point would be to ditch the plugin in favor of implementing those 12 lines ourselves, but I'd love to hear your opinions.~

## Remaining work on this PR: 

- [x] See what kind of styling I can do to the search results
- [x] Ditch the plugin that only partially works for our scenario, in favor of implementing the few lines that _do_ work in our own codebase. Maybe?
- [x] Figure out how to get the algolia settings from gatsby-config.js.


packages/palette-docs/src/components/Sidebar/SearchBox.tsx
```diff
@@ -0,0 +1,15 @@
+import { Input } from "@artsy/palette"
+import React, { useEffect } from "react"
+
+export function SearchBox() {
+  useEffect(() => {
+    console.log("doing the thing with ", window.docsearchSettings)
+    docsearch({
```
pepopowitz: This is where the box _actually_ gets hooked up.

damassi:
> My preference at this point would be to ditch the plugin in favor of implementing those 12 lines ourselves, but I'd love to hear your opinions.

I'm down with that! We can always iterate, same with the other current annoyances.

packages/palette-docs/plugins/gatsby-plugin-local-algolia-docsearch/gatsby-ssr.js
```diff
@@ -0,0 +1,30 @@
+const React = require("react")
+
+exports.onRenderBody = (
```
damassi: 👍

pepopowitz: I don't know how to make this plugin work with ES module syntax, but it isn't something that bothers me. If it bothers anyone else, maybe we could pair on it.

damassi: ES module syntax doesn't work completely in gatsby yet. Caused a bit of stress trying to figure that out when i was working on this before because it _kinda_ worked at times, and was confusing.

pepopowitz:
I think we can use a `stop_urls` setting to exclude them, but I have to figure out how/where to actually use that setting. I'll merge this and look into exclusions as follow-up work.

artsyit:
:rocket: PR was released in `v7.1.1` :rocket:

pepopowitz:
FYI @damassi I believe that [this PR](https://github.com/algolia/docsearch-configs/pull/1729), when merged, should remove the changelog from our search results.

PR #643, opened 2020-02-26
Add new context w/ proper permissions to upload to S3

zephraph:
cc @eessex, @damassi 

I created a new IAM role w/ write-only permissions to palette's S3 bucket. This should fix our problem w/ master failing to upload site artifacts.  

<details>
  <summary>📦 Published PR as canary version: <code>7.1.2-canary.643.9742.0</code></summary>
  <br />
  
  :sparkles: Test out this PR locally via:
  
  ```sh
  npm install @artsy/palette@7.1.2-canary.643.9742.0
  # or 
  yarn add @artsy/palette@7.1.2-canary.643.9742.0
  ```
</details>

PR #644, opened 2020-02-26
Disable acl to bypass requiring the addition of extra permissions

zephraph:
Adding `acl: null` to not require a special ACL permissions grant in IAM. 

I also updated the S3 permissions slightly. 

See #643 for more context.


artsyit:
:rocket: PR was released in `v7.1.2` :rocket:

PR #650, opened 2020-03-11
Update update

damassi:
Updates the bottom update tag to something more generic, as we never actually update that date.


artsyit:
:rocket: PR was released in `v7.1.4` :rocket:

PR #653, opened 2020-03-19
Add name prop to TextArea

pepopowitz:
This PR adds a `name` prop to the TextArea input. 

If a client wasn't using TypeScript, they would have no problem passing a `name` into a `<TextArea>`. The `name` would be attached to the `textarea` element in the DOM, just as they'd expect. However, if the client was using TypeScript, they would get a compilation error. This PR doesn't change how the name gets emitted in the DOM - it just allows for TypeScript clients to pass it in.

## Why?

Adding the `name` property enables for easier testing. `capybara` tests have an idiom where you fill in fields in a form with `fill_in 'field_name', with: value` - but they expect a `name` attribute on the field. Enzyme tests also become easier, as elements can be targeted via a selector like `.find('textarea[name="my-textarea"])`.

It also enables us to use `<label for="name">` elements to describe a textarea for accessibility - though this PR **does not** include modifying the `title` prop to be emitted as a `label` element. 

## Potential followup items

* Modify the `title` prop to be emitted as a `<label for="` element
* Modify the `TextAreaProps` to `extend React.HTMLProps<HTMLTextAreaElement>` similar to [how the `Input` component is built](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Input/Input.tsx#L9). This is not totally trivial, as `HTMLTextAreaElement` includes a `title` prop of a different type than ours.

### DIFF
--- a/packages/palette-docs/content/docs/elements/inputs/TextArea.mdx
+++ b/packages/palette-docs/content/docs/elements/inputs/TextArea.mdx
@@ -5,13 +5,18 @@ name: TextArea
 ## Basic
 
 <Playground>
-  <TextArea placeholder="Start typing..." onChange={console.log} />
+  <TextArea
+    name="basic-text-area"
+    placeholder="Start typing..."
+    onChange={console.log}
+  />
 </Playground>
 
 ## Required
 
 <Playground>
   <TextArea
+    name="required-text-area"
     title="Note"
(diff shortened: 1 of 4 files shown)


damassi:
👌

artsyit:
:rocket: PR was released in `v7.3.0` :rocket:

PR #655, opened 2020-03-24
Use a stable image for CSSGrid story.

pepopowitz:
Because we were using a random image from picsum.com for the CSS Grid story, every build resulted in visual diffs that we had to approve.

This PR specifies the objectively best image available on picsum instead.

image

A closer view:

image


zephraph:
This looks great! Thanks so much!!

zephraph:
I also love that image 😄

pepopowitz:
Thanks for the nudge! I've clicked "approve" on that diff many times but never stopped to think "why do I keep doing this?"

artsyit:
:rocket: PR was released in `v7.4.1` :rocket:

PR #656, opened 2020-03-24
changed casing of fillRule and clipRule for no artwork icon

lilyfromseattle:
changed casing of `fillRule` and `clipRule for no artwork icon from `fill-rule` and `clip-rule` to camel case to silence console errors


artsyit:
:rocket: PR was released in `v7.4.1` :rocket:

PR #657, opened 2020-03-25
Adjust left/right padding on small buttons for mWeb (to match iOS styling)

lilyfromseattle:
[PURCHASE-1781]

The Small button size: I know we built all btns with extra padding on right/left, but it feels unresolved/buggy or something in these smaller screen sizes. Possible to consider eliminating this extra 20px of east-west padding? 

https://files.slack.com/files-pri/T02531TU5-FTNEPUDFA/image.png

we adjusted the left/right padding for small buttons in palette for iOS with the new artwork page, so we should be able to change it on web too

__Before:__
Screen Shot 2020-03-25 at 12 43 15 PM

__After:__
Screen Shot 2020-03-25 at 12 43 23 PM

[PURCHASE-1781]: https://artsyproduct.atlassian.net/browse/PURCHASE-1781

### DIFF
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -39,7 +39,7 @@ export class Button extends Component<WebButtonProps> {
         return {
           height: inline ? "17px" : "26px",
           size: "2",
-          px: inline ? 0 : 1.5,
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in `v7.4.2` :rocket:

PR #660, opened 2020-03-29
Add fallback image support to avatar

zephraph:
This adds the ability to specify a fallback when the image doesn't load for an avatar. I updated the palette docs with a realistic example of how we might use it. 

Note that this should still be ported to iOS. 

image

This adds two props to avatar:

1. `renderFallback` -- Basically a render method to render a component if the image fails to load
2. `onError` -- The is the same callback that's present on images and is what powers the fallback logic. You can also hook into this to do arbitrary things (like remove the `src` from the props so that it falls back to initials if you have those defined).  

A tiny related article: https://medium.com/@Mdmoin07/image-fallback-for-broken-images-react-react-native-261f35b02f45

<details>
  <summary>📦 Published PR as canary version: <code>7.6.0-canary.660.10068.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@7.6.0-canary.660.10068.0
  # or 
  yarn add @artsy/palette@7.6.0-canary.660.10068.0
  ```
</details>

### DIFF
--- a/packages/palette-docs/content/docs/elements/images/Avatar.mdx
+++ b/packages/palette-docs/content/docs/elements/images/Avatar.mdx
@@ -1,17 +1,14 @@
 ---
 name: Avatar
 ---
+
 All entities use a circular avatar.
 
 <Playground title="Medium">
   <Flex>
     <Avatar size="md" src="https://picsum.photos/110/110/?random" />
     <Spacer ml={1} />
-    <Avatar
-      size="md"
-      lazyLoad
-      src="https://picsum.photos/110/110/?random"
-    />
+    <Avatar size="md" lazyLoad src="https://picsum.photos/110/110/?random" />
   </Flex>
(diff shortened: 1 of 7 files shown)


packages/palette-docs/content/docs/elements/images/Avatar.mdx
```diff
@@ -26,3 +23,26 @@ All entities use a circular avatar.
 <Playground title="With initials">
   <Avatar size="xs" initials="CD" />
 </Playground>
+
+<Playground title="With fallback">
+  <Avatar
+    size="xs"
+    renderFallback={({ diameter }) => (
```
damassi: 👌

zephraph:
Added a test 👍

artsyit:
:rocket: PR was released in `v7.6.0` :rocket:

PR #665, opened 2020-04-10
[Deletion] Remove page loader

damassi:
The pageloader logic is more appropriate for Reaction, so removing from palette. PreloadBar remains however.


artsyit:
:rocket: PR was released in `v8.0.0` :rocket:

PR #666, opened 2020-04-12
Allow titles to be modifiable for SVGs

zephraph:
Previously all our titles were hard-coded so if you wanted to use an icon for a slightly different purpose you'd still get the default title on hover. I also tried to go through and make sure all the titles had consistent casing. 

It was a bit of a pain, but I think our icons will be better off for it. 

<details>
  <summary>📦 Published PR as canary version: <code>8.1.0-canary.666.10210.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@8.1.0-canary.666.10210.0
  # or 
  yarn add @artsy/palette@8.1.0-canary.666.10210.0
  ```
</details>


packages/palette/scripts/new_icons_script.rb
```diff
@@ -44,22 +47,22 @@ def fix_attrs(node)
...
 end
 
-def format(slug, title)
+def format(slug)
   input_path = "new_icons/output/min/#{slug}.min.svg"
   xml = File.read(input_path)
   doc = Nokogiri::XML(xml)
   svg = doc.at_css('svg')
-  title = "<title>#{title}</title>"
+  title = "<title>{title}</title>"
```
damassi, line 55: Assuming that `title` was never used as an argument and this is just for interpolation later?

zephraph: Title was previously used as an argument, but we no longer want to interpolate at the script level. `<title>{title}</title>` is the actual JSX we want to produce. The title value comes from the prop `title` and is given a default value of whatever the provided title is.

artsyit:
:rocket: PR was released in `v8.1.0` :rocket:

PR #671, opened 2020-04-17
Update menu component style props

xtina-starr:
This addresses some minor style issues with the drop down menu:
- Ensure menu extends to the fullwidth of the screen.
- Remove top and bottom padding for fullwidth drop down menu.

**Before:**
Screen Shot 2020-04-17 at 2 57 19 PM

**After:**
Screen Shot 2020-04-17 at 3 20 30 PM

<details>
  <summary>📦 Published PR as canary version: <code>8.1.3-canary.671.10374.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@8.1.3-canary.671.10374.0
  # or 
  yarn add @artsy/palette@8.1.3-canary.671.10374.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Menu/Menu.tsx
+++ b/packages/palette/src/elements/Menu/Menu.tsx
@@ -14,6 +14,7 @@ import { Sans } from "../Typography"
 interface MenuProps {
   children?: React.ReactNode
   onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  margin?: SpaceProps["m"]
   title?: string
   width?: number | string
 }
@@ -21,12 +22,13 @@ interface MenuProps {
 /** Menu */
 export const Menu: React.FC<MenuProps> = ({
   children,
+  margin = "2px",
   title,
   width = 230,
   ...props
 }) => {
   return (
-    <MenuContainer width={width} m="2px" {...props}>
+    <MenuContainer width={width} m={margin} {...props}>
       <BorderBox p={0} pb={1} background="white">
         <Flex flexDirection="column" width="100%">
           {title && (
--- a/packages/palette/src/elements/Menu/Menu.tsx
+++ b/packages/palette/src/elements/Menu/Menu.tsx
@@ -15,6 +15,7 @@ interface MenuProps {
   children?: React.ReactNode
   onClick?: (event: React.MouseEvent<HTMLElement>) => void
   margin?: SpaceProps["m"]
+  padding?: SpaceProps["p"]
   title?: string
   width?: number | string
 }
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Menu/Menu.tsx
```diff
@@ -14,20 +14,24 @@ import { Sans } from "../Typography"
 interface MenuProps {
   children?: React.ReactNode
   onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  margin?: SpaceProps["m"]
```
sweir27, line 17: I'm not exactly sure how these palette components are normally set up, but instead of having custom `margin` and `padding` here, could we have the `Menu` component accept `m` and `p` and pass those in directly (with some defaults)?

packages/palette/src/elements/Menu/Menu.tsx
```diff
@@ -14,20 +14,24 @@ import { Sans } from "../Typography"
...
+  padding = 1,
   title,
   width = 230,
   ...props
 }) => {
   return (
-    <MenuContainer width={width} m="2px" {...props}>
-      <BorderBox p={0} pb={1} background="white">
+    <MenuContainer width={width} m={margin} {...props}>
+      <BorderBox p={0} pb={padding} background="white">
```
sweir27, line 34: I think the `padding` name is a little confusing since, to a consumer of this component, I would expect it to affect the overall padding of the outermost component.

If this serves a specific purpose, could we rename? (`outerPadding` or `innerPadding` or something like that?)

packages/palette/src/elements/Menu/Menu.tsx
xtina-starr: Do you mean remain the `margin` and `padding` props to `m` and `p` respectfully?

These props currently have default values, but I can rename.

packages/palette/src/elements/Menu/Menu.tsx
xtina-starr: Yeah it is a bit confusing. I think I can refactor this a bit.

packages/palette/src/elements/Menu/Menu.tsx
sweir27: I realize I made this comment before my second one 🙈.

It looks like the `margin` value here changes the outermost margin on the component, so renaming to `m` makes sense to me.

It seems like `padding` is a separate case!

packages/palette/src/elements/Menu/Menu.tsx
```diff
@@ -43,9 +43,7 @@ export const Menu: React.FC<MenuProps> = ({
             </Box>
           )}
 
-          <Flex flexDirection="column" pt={title ? 0 : padding}>
-            {children}
-          </Flex>
+          <Flex flexDirection="column">{children}</Flex>
```
xtina-starr, line 46: Okay this attempts to simplify conditionally adjusting the padding. Previously we were giving the box container for the title 20px of top padding. Then conditionally adding 0px of top padding to the menu content flex container when there was a title present and 10px when there wasn't. This refactor allows `BorderBox`  to handles the top and bottom padding. 

Does this make sense @sweir27 ?

sweir27: Makes a lot more sense! Nice refactor.

xtina-starr:
@sweir27 you cool with me merging this?

artsyit:
:rocket: PR was released in [`v8.1.3`](https://github.com/artsy/palette/releases/tag/v8.1.3) :rocket:

PR #673, opened 2020-04-23
[FX-1907] make background color darker for legibility with light text colors

xtina-starr:
Addresses: [FX-1907](https://artsyproduct.atlassian.net/browse/FX-1907)

image

Adjusts the back-ground color to be more legible when we use lighter colors.

Screen Shot 2020-04-23 at 5 55 30 PM

<details>
  <summary>📦 Published PR as canary version: <code>8.2.1-canary.673.10444.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@8.2.1-canary.673.10444.0
  # or 
  yarn add @artsy/palette@8.2.1-canary.673.10444.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Menu/Menu.tsx
+++ b/packages/palette/src/elements/Menu/Menu.tsx
@@ -80,7 +80,7 @@ export const MenuItem: React.FC<MenuItemProps> = ({
   ...props
 }) => {
   return (
-    <MenuLink href={href} {...props}>
+    <MenuLink href={href} {...props} hasLighterColor={Boolean(textColor)}>
       <Box px={px} py={py}>
         <MenuLinkText size={fontSize} weight={textWeight} color={textColor}>
           {children}
@@ -90,7 +90,7 @@ export const MenuItem: React.FC<MenuItemProps> = ({
   )
 }
 
-const MenuLink = styled.a`
+const MenuLink = styled.a<{ hasLighterColor: boolean }>`
   cursor: pointer;
   display: flex;
   text-decoration: none;
@@ -101,7 +101,8 @@ const MenuLink = styled.a`
   ${display};
 
   &:hover {
-    background-color: ${color("black5")};
+    background-color: ${p =>
+      p.hasLighterColor ? color("black10") : color("black5")};
   }
 
   ${Sans} {
--- a/packages/palette/src/elements/Menu/Menu.tsx
+++ b/packages/palette/src/elements/Menu/Menu.tsx
@@ -66,6 +66,7 @@ interface MenuItemProps extends BoxProps {
   py?: SpaceProps["py"]
   textColor?: string
   textWeight?: "medium" | "regular"
+  hasLighterTextColor?: boolean
 }
 
 /** MenuItem */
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Menu/Menu.tsx
```diff
@@ -80,7 +80,7 @@ export const MenuItem: React.FC<MenuItemProps> = ({
   ...props
 }) => {
   return (
-    <MenuLink href={href} {...props}>
+    <MenuLink href={href} {...props} hasLighterColor={Boolean(textColor)}>
```
sweir27, line 83: This makes sense!

Two questions:
- Is there a way to derive this from the `textColor` prop? (Or another prop?) Like-- if the text color is lighter than X, make the background hover color Y.
- I'm not super familiar with everywhere this component is used, but is it possible we want this change to be across-the-board and can avoid the conditional?

xtina-starr: @sweir27 I was trying not to introduce a new prop/litter this component with props for these one off things. 

I like the idea of inferring the background based off the `textColor`. I think we'd need to collaborate with the design team to get some specs for that. Could we create a ticket for this and I commit to revisiting during upkeep time?

sweir27: 👍 Right now it looks like this prop will be `true` if _any_ `textColor` is provided to the component. That seems potentially confusing (especially if someone passes in `black100` for that value).

It might be worth following up quickly with the design team to see if they can provide some direction, or I'd rename this to `hasLighterTextColor` and accept the prop from the outermost component so it can be precisely controlled!

xtina-starr:
Here is the corresponding reaction PR: https://github.com/artsy/reaction/pull/3447

packages/palette/src/elements/Menu/Menu.tsx
xtina-starr: I updated this PR to pass in `hasLighterTextColor` which is a boolean and is used to determine the background-color.

packages/palette/src/elements/Menu/Menu.tsx
```diff
@@ -77,10 +78,15 @@ export const MenuItem: React.FC<MenuItemProps> = ({
...
   textColor,
+  hasLighterTextColor,
   ...props
 }) => {
   return (
-    <MenuLink href={href} {...props}>
+    <MenuLink
+      href={href}
+      {...props}
+      hasDarkerBacgroundColor={hasLighterTextColor}
```
sweir27, line 88: #superminor I think it would be slightly clearer/simpler to keep the prop naming here (so `hasLighterTextColor={hasLighterTextColor}`). But that's not a blocker!

There is a little typo (`Bacground`) 🙈. Please at least fix that and/or make the naming change, and then merge away. 😄 Thanks for bearing with me on this!!

xtina-starr: Haha oops! for sure!

artsyit:
:rocket: PR was released in [`v8.2.1`](https://github.com/artsy/palette/releases/tag/v8.2.1) :rocket:

PR #674, opened 2020-05-04
[FX-1936]: Low-level modal component

dzucconi:
Re: [FX-1936](https://artsyproduct.atlassian.net/browse/FX-1936)

Context here is that we should have a Modal primitive that handles some of the givens like locking scroll in a cross-platform manner, as well as handles any a11y concerns (like trapping focus in the modal)

This makes use of [react-focus-on](https://github.com/theKashey/react-focus-on) which checks all the boxes.

-----

This shouldn't introduce any change in behavior but I'd like to just generally check things out integrated in Force before merging. Also I'm concerned about the specs in consumers since the previous version of enzyme here would cause specs to fail (uncertain as to why).


packages/palette/package.json
```diff
@@ -120,6 +120,7 @@
     "d3-shape": "^1.3.5",
     "luxon": "^1.15",
     "rc-slider": "^8.6.2",
+    "react-focus-on": "^3.4.1",
```
dzucconi, line 123: So, I'm always reluctant to introduce any UI-specific dependency. Here I think it makes sense because doing all of this and doing it correctly in a cross-platform way is super tricky.

Alternatively if we wanted to punt on the a11y stuff for now, we can fix/improve the scroll locking and still extract the ModalBase, which would solve the immediate issue at hand.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -227,19 +205,6 @@ const FixedButtonWrapper = styled(Flex)`
...
-  position: fixed;
-  top: 0;
-  left: 0;
-  z-index: 9999;
-  width: 100vw;
-  height: 100vh;
-  background-color: rgba(229, 229, 229, 0.5);
-  transition: opacity 250ms ease;
-  opacity: ${props => (props.show ? "1" : "0")};
-  pointer-events: ${props => (props.show ? "auto" : "none")};
```
dzucconi, line 240: This fading etc, never worked as far as I can tell?

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -288,11 +253,24 @@ const ModalScrollContent = styled(Flex)<ModalScrollContentProps>`
   `};
 `
 
-const CloseIconWrapper = styled(Box)`
+// TODO: Consider extracting the equivalent of `TouchableWithoutFeedback`
```
dzucconi, line 256: So yeah, something I find very useful is basically implementing a `Box` component with a button underlying it. Something like https://github.com/auspices/eos/blob/master/src/Clickable/Clickable.ts

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -288,11 +253,24 @@ const ModalScrollContent = styled(Flex)<ModalScrollContentProps>`
...
+  right: 0;
   cursor: pointer;
+  padding: ${space(2)}px ${space(2)}px 0 0;
+  background-color: transparent;
+  border: 0;
+  color: ${color("black100")};
+
+  &:focus {
+    path {
+      fill: ${color("black60")};
```
dzucconi, line 271: In general I think it's a bit simpler for Icons to just be `currentColor` by default which makes them simpler to control from the outside.

packages/palette/src/elements/Modal/Modal.tsx
```diff
@@ -288,11 +253,24 @@ const ModalScrollContent = styled(Flex)<ModalScrollContentProps>`
...
+// TODO: Icon should support `currentColor` as a valid color
+const CloseButton = styled.button`
   position: fixed;
-  top: ${space(2)}px;
-  right: ${space(2)}px;
+  appearance: none;
+  top: 0;
+  right: 0;
   cursor: pointer;
+  padding: ${space(2)}px ${space(2)}px 0 0;
```
dzucconi, line 264: Do we have a better way of doing this? (Other than attrs)

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,79 @@
...
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  /* TODO: Use a token? Or some other alternative to wildstyle z-index? */
+  z-index: 9999;
```
dzucconi, line 17: Do we have a centralized z-map anywhere? I tend to think of this as application specific and it's sorta weird to put in a UI library IMO. Though it can be annoying having to provide them.

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  ${({ overlay }) =>
+    overlay &&
+    css`
+      /* TODO: Use a token? */
+      background-color: rgba(229, 229, 229, 0.5);
```
dzucconi, line 23: I know a lot of the stuff in the modal looks pre-tokens? So maybe there's a helper for this sort of thing?

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return () => {
+      document.body.removeChild(current)
+    }
+  }, [])
+
+  return createPortal(
+    <FocusOn onClickOutside={onClose} onEscapeKey={onClose}>
+      <ModalWrapper
+        // TODO: Remove cast as any. Something is wrong with styled's typings?
+        ref={wrapperEl as any}
```
dzucconi, line 66: Not sure what's up with this since `styled.whatever` should be able to accept a ref with the underlying DOM type. I'm assuming it's just related to the older version here?

packages/palette/src/elements/Modal/__tests__/Modal.test.tsx
```diff
@@ -125,9 +128,7 @@ describe("Modal", () => {
         vestibulum.
       </Modal>
     )
-    setTimeout(() => {
-      document.dispatchEvent(new KeyboardEvent("keyup", { key: "Escape" }))
-      expect(show).toEqual(false)
-    })
```
dzucconi, line 131: This spec was failing and just being swallowed by the setTimeout

packages/palette/src/elements/Modal/Modal.tsx
zephraph: We should probably eventually update that.

packages/palette/src/elements/Modal/ModalBase.tsx
zephraph: I don't think we do, but if we did I definitely think it should go in the theme file. You can go ahead and move it there if you want.

packages/palette/src/elements/Modal/ModalBase.tsx
zephraph: Very likely a version conflict. We really need to upgrade styled components but that's a whole thing.

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "palette",
   "private": true,
+  "engines": {
+    "node": "10.x"
+  },
```
damassi, line 6: Is this necessary? We shouldn't have to lock our node version for palette.

packages/palette/src/elements/Modal/Modal.story.tsx
```diff
@@ -1,46 +1,55 @@
...
-      <Modal show onClose={() => null}>
-        Some example content
-      </Modal>
-    )
+    return <Story show>Some example content</Story>
   })
   .add("With title", () => {
     return (
-      <Modal show title="Modal Title" onClose={() => null}>
+      <Story show title="Modal Title">
```
damassi, line 26: Nice

packages/palette/src/elements/Modal/Modal.tsx
damassi: This is how we've been doing it, outside of props / attrs.

packages/palette/src/elements/Modal/ModalBase.tsx
damassi: We're def over 900, because of old force code.

packages/palette/src/elements/Modal/ModalBase.tsx
damassi: FYI, we generally attach portal conteent to [this ID](https://github.com/artsy/force/blob/master/src/desktop/components/main_layout/templates/react_blank_index.jade#L18) (which is spread across a few different templates).

package.json
dzucconi: I think engines is probably the wrong place for this. An nvmrc is better. But you can't yarn install here in 12 because palette-docs can't build. It actually prevents `yarn add` because it's trying to rebuild on add. (Due to the older sharp in the gatsby image plugins).

I'll switch this to nvmrc so it reflects the local development environment. But we should log some kind of ticket for cleaning up docs to make things installable.

packages/palette/src/elements/Modal/ModalBase.tsx
dzucconi: @damassi will update — what's the benefit of portaling into a specific DOM node — so it's predictable and ensures there's only ever a single modal? What I don't love about it is that it creates an assumption about the structure of the DOM in this library that we can't guarantee. So the proper thing to do then would be: either it's configurable, or we search for that specific node then fall back to just appending.

packages/palette/package.json
zephraph: Yeah, understand that. 

Just to document the decision, did you consider any alternatives? Was there a specific reason for choosing this dep?

dzucconi: * One possible alternative is [react-modal](https://reactcommunity.org/react-modal/) — which I think makes too many assumptions about your UI, and is a single-purpose dependency (it's just a modal).
* Alternatives for something as "simple" as scroll-locking are hit and miss. For instance the popular [body-scroll-lock](https://github.com/willmcpo/body-scroll-lock) will break under certain SSR environments currently.
* react-focus-on is a composition of a couple of different packages, each of which is independently useful (see [first paragraph in README](https://github.com/theKashey/react-focus-on))
* All of those sub-packages are by the same author/of similar quality/include type definitions
* react-focus-on also supports scattered locks through it's `shards` API

packages/palette/src/elements/Modal/Modal.tsx
dzucconi: Why do we specify the spacing scale as numbers and not px literals if they are only ever px? Simpler to do math with?

dzucconi: We could do a helper like https://github.com/ModaOperandi/tokens/blob/master/src/helpers.ts#L23 which would at least make this less verbose. I'd be interested to know if we can ascertain if we're in a string literal or not from inside the function then automatically append the `px`

packages/palette/src/elements/Modal/ModalBase.tsx
damassi: That sounds good; yeah, this was more just historical info to be aware of, how we handle our current auth models, etc

packages/palette/src/elements/Modal/ModalBase.tsx
zephraph: Well, some of the things (drop shadows, etc) may not be specified in tokens. Not that they shouldn't be, just something we haven't standardized yet.

packages/palette/src/elements/Modal/Modal.tsx
damassi: Would be lying if I said we _didn't_ use the `space` helper a good deal, but mostly we've used react props piped through to styled-system which gets around things.

The decision around the space scale came from the design team directly; we just copied their system: https://palette.artsy.net/tokens/spacing/

dzucconi: Actually come to think of it; it'd be more proper to just use a `themeGet` from within so that you access the current theme. (So for instance, if we switch to a separate font/space scale for mobile, this method would prohibit us from swapping theme since the function is outside the context tree)

zephraph:
Hey @dzucconi, might be worth poking at the UI tests. Click the details link and it'll take you to that page. You can sign in w/ GitHub. These tests run visual snapshots of our storybooks. 

There are some modal tests that are rendering a button initially in storybooks that were rendering the modal previously. If you think we don't need visual tests for those, you can [skip them](https://www.chromatic.com/docs/ignoring-elements#ignore-stories-and-elements)

packages/palette/src/elements/Modal/Modal.tsx
damassi: Yeah ideally we'd be able to compose that somewhere [in here](https://github.com/artsy/palette/blob/master/packages/palette/src/helpers/space.ts#L7) so that a later refactor wouldn't be too drastic, but not sure how without direct access to context.

dzucconi:
@zephraph yeah I see I'm just trying to see how involved integration in Reaction is. Honestly I think I might back out of this PR, it breaks _alot_ of things in Reaction.

Options are just not have this modal inherit from ModalBase just yet, or just fix scroll locking independently and punt on a11y.

dzucconi:
Ah you know what, I was running the test and thinking all of the `act` warnings were errors. Only actually broke 2 things. Let's see.

dzucconi:
@zephraph Also realizing this is going to be a proper breaking change, since specs are asserting on modal contents being inside of the wrapper component. Is the protocol to just to mark the major change then attempt to upgrade the consumers?

zephraph:
@dzucconi, yep, that's right. Just mark it as major.

zephraph:
@dzucconi I'll leave it to your best judgement on whether to continue or not. Accessibility _is_ important and having it properly handled the first go around would be amazing... that said, if you feel like there's a lot of breaking changes here, then we could roll that up into a second PR to unblock any on-going work.

dzucconi:
Closed in favor of https://github.com/artsy/palette/pull/675

PR #680, opened 2020-05-11
Remove Modal `forcedScroll` prop / Default `overflow` to `auto`

dblandin:
Slack thread: https://artsy.slack.com/archives/C9XJKPY9W/p1576267394001200

This will be a breaking change and will required some compatibility updates in @artsy/reaction.

### Before

Peek 2020-05-13 12-46

### After

Peek 2020-05-13 12-44

### DIFF
--- a/packages/palette/src/elements/Modal/Modal.tsx
+++ b/packages/palette/src/elements/Modal/Modal.tsx
@@ -25,7 +25,6 @@ interface ModalProps {
   onClose: () => void
   show?: boolean
   title?: string
-  forcedScroll?: boolean // TODO: Find out if we can globally switch to `overflow: auto`
   /*
    * Hide the X button if we don't want the user to be able to exit the modal without another action closing the modal
    */
@@ -41,7 +40,6 @@ interface TransitionElementProps {
 interface ModalScrollContentProps {
   hasLogo?: boolean
   modalWidth?: ModalWidth
-  forcedScroll?: boolean
 }
 
 export enum ModalWidth {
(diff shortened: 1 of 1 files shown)


damassi:
@dblandin - Mind sequentially listing exactly what compatability updates are needed to take this to completion in the PR description?

dblandin:
> Mind listing in the PR description exactly what compatibility updates are needed to take this to completion? Then the work required isn't all in yr head and if need be someone else could pick it up.

Sure thing! The main change is to remove all uses of the `forcedScroll` Modal prop.

I'll echo a comment I made in Slack: another option might be to keep the `forcedScroll` property and set it explicitly to true to keep the `overflow: scroll` behavior for existing modals, but I still get the feeling that this is accidental, not intentional, behavior so I'm hesitant to do that.

In that case, we could remove any `forcedScroll={false}` properties, as that will be the new default behavior, and set `forcedScroll={true}` to any existing modals where we're unsure of whether it's needed or not.

@damassi Do you have any preference here?

damassi:
Nope, no preference! `forcedScroll` seems weird to me, and not sure why it was ever needed.

dblandin:
> Nope, no preference! forcedScroll seems weird to me, and not sure why it was ever needed.

I think this `overflow: scroll` setting has always been confusing and the `forcedScroll` property was introduced to get around it temporarily: https://github.com/artsy/palette/pull/623

dblandin:
Going to go ahead and merge this in.

artsyit:
:rocket: PR was released in [`v9.0.0`](https://github.com/artsy/palette/releases/tag/v9.0.0) :rocket:

PR #681, opened 2020-05-12
Mixes in missing flex-related functions

dzucconi:
I was mistaken about `flexBasis` which is here and works fine.

Missing were `alignSelf`, `flex`, `justifyItems`, `justifySelf` — if you're interested in these, which brings this inline with styled-system's current flexbox: https://github.com/styled-system/styled-system/blob/master/packages/flexbox/src/index.js

### DIFF
--- a/packages/palette/src/elements/Flex/Flex.tsx
+++ b/packages/palette/src/elements/Flex/Flex.tsx
@@ -5,24 +5,33 @@ import {
   AlignContentProps,
   alignItems,
   AlignItemsProps,
+  alignSelf,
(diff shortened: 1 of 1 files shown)


dzucconi:
This isn't immediately pressing to me so this can be take it or leave it. I think realistically we just need to work on the styled-system upgrade at large.

damassi:
LGTM 👍

artsyit:
:rocket: PR was released in [`v8.2.4`](https://github.com/artsy/palette/releases/tag/v8.2.4) :rocket:

PR #683, opened 2020-05-13
[Modal] Do not fire external `onClose` callback on first render

dblandin:
Don't call `onClose` callback if we're within the initial render.


artsyit:
:rocket: PR was released in [`v9.0.1`](https://github.com/artsy/palette/releases/tag/v9.0.1) :rocket:

PR #684, opened 2020-05-15
[COM-16] - add unstable_trackIn as a prop to Sans and Serif

oxaudo:
Not 100% sure that I understand the ask correctly. But that does add `unstable_trackIn` to Sans and Serif and puts a sample in the doc file. It's collapsed by default.

Looks like that:
Screen Shot 2020-05-15 at 2 59 26 PM
Screen Shot 2020-05-15 at 2 59 43 PM

### DIFF
--- a/packages/palette/src/elements/Typography/Typography.test.tsx
+++ b/packages/palette/src/elements/Typography/Typography.test.tsx
@@ -64,6 +64,16 @@ describe("Typography", () => {
           themeProps.fontFamily.sans.mediumItalic
         )
       })
+
+      it("supports unstable_trackIn to adjust letter-spacing", () => {
+        const sans = renderer.create(
+          <Sans size="3t" unstable_trackIn>
+            Hello world
+          </Sans>
+        ).root
+        const text = sans.findByType(Text as React.ComponentClass<any>)
+        expect(text.props.style).toEqual({ letterSpacing: "-0.03em" })
+      })
     })
 
     describe("concerning font-size & line-height", () => {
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Typography/Typography.tsx
```diff
@@ -186,6 +198,7 @@ function createStyledText<P extends StyledTextProps>(
             fontFamilyType && themeProps.fontFamily[fontType][fontFamilyType]
           }
           {...determineFontSizes(fontType, size)}
+          {...(unstable_trackIn ? { letterSpacing: "-0.03em" } : {})}
```
oxaudo, line 201: So I guess ideally this should be passed to `determineFontSizes` and only applied if there is a proper set up for it I think. But for now applying it directly here.

packages/palette/src/elements/Typography/Typography.tsx
```diff
@@ -106,6 +109,7 @@ export const Text = primitives.Text<TextProps>`
   ${space};
   ${textAlign};
   ${verticalAlign};
+  ${letterSpacing};
```
damassi, line 112: If this is mixed in then a user would be able to do `<Sans letterSpacing="foo">` as a publically available prop, which we want to avoid.

packages/palette/src/elements/Typography/Typography.tsx
damassi, line 201: instead of setting the prop via the mixin up above, we can lock it down a bit:

Up above where we pluck out the props, can do:
```
...
unstable_trackIn,
element,
style,
...textProps
...
```
(note how we're now pulling out `style`)

And then here
```
style={{
  ...(unstable_trackin ? { letterSpacing: "-0.03em" } : {}) 
  ...style
}}
```
Hope this makes sense...

artsyit:
:rocket: PR was released in [`v9.1.0`](https://github.com/artsy/palette/releases/tag/v9.1.0) :rocket:

packages/palette/src/elements/Typography/Typography.test.tsx
```diff
@@ -140,6 +150,16 @@ describe("Typography", () => {
...
 
+      it("supports unstable_trackIn to adjust letter-spacing", () => {
+        const sans = renderer.create(
+          <Serif size="3t" unstable_trackIn>
+            Hello world
+          </Serif>
+        ).root
+        const text = sans.findByType(Text as React.ComponentClass<any>)
+        expect(text.props.style).toEqual({ letterSpacing: "-0.03em" })
+      })
```
damassi, line 161: 👍

PR #685, opened 2020-05-18
No unstable_trackIn in eigen

ds300:
@oxaudo this change you made failed in eigen (artsy/eigen#3318) because the `style` prop is allowed to be an array in react-native but this code assumed that it was always an object. In this PR I just turned the `unstable_trackIn` feature off in react-native, since I guess it's a web-only thing?

### DIFF
--- a/packages/palette/src/elements/Typography/Typography.tsx
+++ b/packages/palette/src/elements/Typography/Typography.tsx
@@ -31,6 +31,7 @@ import {
   TextAlignProps,
 } from "styled-system"
 
+import { isReactNative } from "../../helpers/isReactNative"
 import { determineFontSizes } from "./determineFontSizes"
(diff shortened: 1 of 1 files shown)


damassi:
@ds300 - this is a temporary fix for our new typography style guide, it (should) apply to mobile and when the new typography components are completed this will be replaced.

oxaudo:
Oh - sorry @ds300 - definitely did not test this on Eigen. For now it's totally fine that it applies to web only but as @damassi pointed out - this is a temporary fix either way.

Thanks for addressing.

artsyit:
:rocket: PR was released in [`v9.2.2`](https://github.com/artsy/palette/releases/tag/v9.2.2) :rocket:

PR #686, opened 2020-05-18
Update the Banner component to work well with a content that takes more space

yuki24:
This PR has a few changes:

 * Make `props.message` for the `<Banner>` component optional. It can take `props.children` so we may not always have to specify `props.message`, so I think it makes sense to make it optional.
 * Fix the close button position when the context of a `<Banner>` expands the height of the component (see the before/after below).
 * Add a new Dismissable Banner section to the doc to clarify the necessity of this update.

## before

Screen Shot 2020-05-18 at 6 11 03 PM

## After

Screen Shot 2020-05-18 at 6 10 43 PM

## Doc

doc

### DIFF
--- a/packages/palette-docs/content/docs/elements/dialogs/Banner.mdx
+++ b/packages/palette-docs/content/docs/elements/dialogs/Banner.mdx
@@ -61,3 +61,18 @@ use colored backgrounds with white text.
 </Box>
 
 <Spacer my={6} />
+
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`v9.2.0`](https://github.com/artsy/palette/releases/tag/v9.2.0) :rocket:

PR #687, opened 2020-05-19
[CSGN-203] Add PaymentIcon

pepopowitz:
Partially addresses https://artsyproduct.atlassian.net/browse/CSGN-203

For the iOS Sell Tab landing page, we want a generic "credit card" icon that matches our other icons. I opted to name this new icon `PaymentIcon` to avoid confusion with the existing `CreditCardIcon`, which currently serves up one of 5 flavors based on the type of card.

The new icon looks like this:

image

Follow-up work takes place in eigen, to use this new icon.


artsyit:
:rocket: PR was released in [`v9.3.0`](https://github.com/artsy/palette/releases/tag/v9.3.0) :rocket:

PR #689, opened 2020-05-20
Add disabledtext to radiogroup

pvinis:
For context, this is something we need for the PR https://github.com/artsy/reaction/pull/3558 which converts the price from a range to a radio.

This is what range looks like:
Screenshot 2020-05-19 at 14 01 45

This is what radio looks like now:
Screenshot 2020-05-20 at 21 28 49

This PR also adds https://github.com/storybookjs/storybook/tree/master/addons/knobs to make our DX just a tiny bit nicer. With knobs we can edit props directly in the storybook like:
Screenshot 2020-05-20 at 21 30 39

### DIFF
--- a/packages/palette/.storybook/addons.js
+++ b/packages/palette/.storybook/addons.js
@@ -1,3 +1,4 @@
 import "@storybook/addon-actions/register"
+import "@storybook/addon-knobs/register"
 import "@storybook/addon-links/register"
 import "@storybook/addon-viewport/register"
--- a/packages/palette/src/elements/RadioGroup/RadioGroup.test.tsx
+++ b/packages/palette/src/elements/RadioGroup/RadioGroup.test.tsx
@@ -99,6 +99,17 @@ describe("RadioGroup", () => {
     ).toBe(true)
   })
 
+  it("displays a specific text when disabled", () => {
+    const wrapper = mount(
+      <RadioGroup disabled disabledText="i am disabled right now mate">
+        <Radio value="SHIP">Provide shipping address</Radio>
+        <Radio value="PICKUP">Arrange for pickup</Radio>
+      </RadioGroup>
+    )
+
+    expect(wrapper.find("Sans").text()).toBe("i am disabled right now mate")
+  })
+
   it("allows Radios within the group to be deselectable", () => {
     const wrapper = mount(
       <RadioGroup defaultValue="PICKUP" deselectable>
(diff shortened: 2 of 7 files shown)


packages/palette/.storybook/addons.js
```diff
@@ -1,3 +1,4 @@
 import "@storybook/addon-actions/register"
+import "@storybook/addon-knobs/register"
```
damassi, line 2: Nice

sweir27:
This looks great!!

packages/palette/src/elements/RadioGroup/RadioGroup.tsx
```diff
@@ -90,9 +93,21 @@ export class RadioGroup extends React.Component<
...
+      disabled,
+      disabledText,
+      onSelect,
+      defaultValue,
+      children,
+      ...others
+    } = this.props
     return (
       <Flex flexDirection="column" {...others}>
+        {disabled && disabledText !== undefined && (
```
damassi, line 106: Don't need to explicitly check for undefined here. Also slight fix for form (for instances when it is needed):

```
typeof variableName !== "undefined"
```

When checking for undefined

packages/palette/src/elements/RadioGroup/RadioGroup.test.tsx
```diff
@@ -99,6 +99,17 @@ describe("RadioGroup", () => {
...
 
+  it("displays a specific text when disabled", () => {
+    const wrapper = mount(
+      <RadioGroup disabled disabledText="i am disabled right now mate">
+        <Radio value="SHIP">Provide shipping address</Radio>
+        <Radio value="PICKUP">Arrange for pickup</Radio>
+      </RadioGroup>
+    )
+
+    expect(wrapper.find("Sans").text()).toBe("i am disabled right now mate")
```
damassi, line 110: Thanks for adding the test 👍

pvinis:
`auto` will take care of the release I guess, right?

damassi:
@pvinis - will want to update the label on the right from `patch` to `minor`, otherwise yup -- should release a new version and then Renvovate should take the lead in updating the rest of our apps with said change.

artsyit:
:rocket: PR was released in [`v9.7.0`](https://github.com/artsy/palette/releases/tag/v9.7.0) :rocket:

PR #690, opened 2020-05-20
Adds HTML component

dzucconi:
Adds simple component to render HTML sourced from Markdown into.

PR #691, opened 2020-05-21
Adds HTML.ios.tsx

dzucconi:
Re: https://github.com/artsy/palette/pull/690#commitcomment-39336360

@ds300 Just so my mental model of how this works is clear — I'm assuming this is breaking Eigen due to `dangerouslySetInnerHTML` — but how does the existence of that code break Eigen if it's not being executed?


artsyit:
:rocket: PR was released in [`v9.6.0`](https://github.com/artsy/palette/releases/tag/v9.6.0) :rocket:

ds300:
@dzucconi this was the failure https://app.circleci.com/pipelines/github/artsy/eigen/1779/workflows/d10171a9-487b-438f-b9bd-8ec3b3ba3ee3/jobs/10378

Looking at the stack trace, it failed when trying to create a new styled component, presumably because it was using the web version of the styled-components api rather than the native version.

ds300:
and thanks for sorting this 🙌 😄

PR #693, opened 2020-05-26
[FX-1936] Implements ModalBase (take II)

dzucconi:
So here I'm re-implementing the lower-level modal but without forcing it onto the existing one. The intent is to use this for the mobile action sheets in Force (!), and implement any newer modally behavior on top of this. I intend to migrate the existing modal to use this as well, but I'm going to punt on that migration for the time being.

Took a slightly lighter touch here as well with only using a library for the scroll-locking, and implemented the focus management stuff myself.

Keyboard navigation demo (we really need focus rings on Buttons!):

---

I'm gonna yalc this over to Force before converting to a real PR.

<details>
  <summary>📦 Published PR as canary version: <code>9.8.2-canary.693.11035.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@9.8.2-canary.693.11035.0
  # or 
  yarn add @artsy/palette@9.8.2-canary.693.11035.0
  ```
</details>

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -73,6 +73,7 @@
     "@types/luxon": "1.15.1",
     "@types/node": "10.14.6",
     "@types/react": "16.8.7",
+    "@types/react-dom": "^16.9.8",
     "@types/react-lazy-load-image-component": "1.3.0",
     "@types/react-test-renderer": "16.8.1",
     "@types/semver": "5.5.0",
@@ -102,7 +103,7 @@
     "react-native": "0.55.4",
     "react-native-svg": "9.6.4",
     "react-test-renderer": "16.8.6",
-    "regenerator-runtime": "0.11.1",
+    "regenerator-runtime": "^0.13.5",
     "simple-progress-webpack-plugin": "1.1.2",
     "static-extend": "0.1.2",
     "storybook-chromatic": "1.3.3",
@@ -123,10 +124,12 @@
     "react-lazy-load-image-component": "^1.4.1-beta.0",
     "react-live": "^1.12.0",
     "react-powerplug": "^1.0.0",
+    "react-remove-scroll": "^2.3.0",
     "react-spring": "^8.0.27",
     "styled-bootstrap-grid": "1.0.4",
     "styled-system": "^3.1.11",
-    "trunc-html": "^1.1.2"
+    "trunc-html": "^1.1.2",
+    "use-cursor": "^1.2.0"
   },
   "husky": {
     "hooks": {
--- a/packages/palette/src/elements/Modal/ModalBase.story.tsx
+++ b/packages/palette/src/elements/Modal/ModalBase.story.tsx
@@ -1,0 +1,85 @@
+import { storiesOf } from "@storybook/react"
+import React, { useState } from "react"
+import { Box } from "../Box"
+import { Button } from "../Button"
+import { Input } from "../Input"
+import { Join } from "../Join"
+import { Spacer } from "../Spacer"
+import { Sans } from "../Typography"
+import { ModalBase, ModalBaseProps } from "./ModalBase"
+
+const Example: React.FC<
+  ModalBaseProps & { dialogChildren?: JSX.Element; bodyChildren?: JSX.Element }
+> = ({ bodyChildren, dialogChildren, ...rest } = {}) => {
+  const [open, setOpen] = useState(false)
+  const label = open ? 'opened' : 'open'
+  const handleClose = () => setOpen(false)
+
+  return (
+    <>
+      <Button variant="secondaryGray" onClick={() => setOpen(true)}>
+        {label}
+      </Button>
+
+      {bodyChildren}
+
+      {open && (
+        <ModalBase onClose={handleClose} {...rest}>
+          <Box
+            background="black"
+            p={4}
+            width="100%"
+            height="100%"
+            style={{ border: "2px solid red" }}
+          >
+            <Box textAlign="center">
+              <Sans size="6" color="white100">
+                <Join separator={<Spacer my={1} />}>
+                  <>Some example content. Click outside to close.</>
+                  <Button variant="primaryWhite" onClick={handleClose}>
+                    Or click here to close.
+                  </Button>
+                  <Input placeholder="Just an example for focusing" />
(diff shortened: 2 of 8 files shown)


packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,167 @@
...
+  ZIndexProps & {
+    children?: React.ReactNode
+    dialogProps?: FlexProps
+    onClose?(): void
+  }
+
+/**
+ * It seems we've landed on this value as the 'top'
+ */
+export const DEFAULT_MODAL_Z_INDEX = 9999
```
dzucconi, line 52: Now that Reaction is inside of Force we can actually extract a comprehensive z-index map. I'll give that a shot as well.

packages/palette/src/elements/Modal/ModalBase.story.tsx
```diff
@@ -0,0 +1,85 @@
...
+  ModalBaseProps & { dialogChildren?: JSX.Element; bodyChildren?: JSX.Element }
+> = ({ bodyChildren, dialogChildren, ...rest } = {}) => {
+  const [open, setOpen] = useState(false)
+
+  const handleClose = () => setOpen(false)
+
+  return (
+    <>
+      <Button variant="secondaryGray" onClick={() => setOpen(true)}>
+        {{ true: "opened", false: "open" }[`${open}`]}
```
damassi, line 21: Mind pushing this into a self-describing variable? Looking closely its clear this tis logging state, but the form might be easier to parse at a glance later if it was labeled `{buttonState}` (or something).

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,167 @@
...
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  align-items: center;
+  justify-content: center;
+  ${systemZIndex}
+`
+
+const Dialog = styled(Flex).attrs({ role: "dialog" })`
```
damassi, line 30: tyy

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -0,0 +1,167 @@
...
+  useEffect(() => {
+    const { current } = appendEl
+
+    const focusedElBeforeOpen = document.activeElement as HTMLElement
+
+    // Append the dialog
+    document.body.appendChild(current)
+
+    // Gets the focusable elements
+    const __focusableEls__ = [].slice.call(
```
damassi, line 128: Out of curiosity what does the `__` signify?

damassi: And can this use [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from)?

dzucconi:
I'm finding working with refs to be unintuitive here because of the non-strict nulls. This codebase is probably an easy first target to convert. At a glance like 90% of the issues are just ref nullchecks.

packages/palette/src/elements/Modal/ModalBase.tsx
dzucconi: I should probably just change this to a single leading underscore which is probably more conventional. Just "temporary/private", since I don't want to wait for the callback on the setState to get at the actual els for the immediate focus handling.

dzucconi: Ah yeah, I couldn't think of that method off the top of my head 😆

packages/palette/src/utils/useCursor.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (max === 0) return 0
+  if (cursor >= 0) return cursor % max
+  const modulo = cursor % max
+  if (modulo === 0) return Math.abs(modulo) // -0
+  return modulo + max
+}
+
+/**
+ * useCursor
+ * Cycles an index within some max length
```
damassi, line 13: Don't you have something like this extracted into a lib? Might be better to use that unless you can imagine this hook being used often elsewhere in palette code. If so, might be good to add a light description to this comment with a real world application for those unfamilar with `cursor` as a concept.

dzucconi: Yeah — [I just inlined it since it's just simple/small](https://github.com/dzucconi/use-cursor). You're probably right.

I find I lean on this pattern for handling literally anything in a UI with next/previous since it's one of those things that is easy to screw up or forget, so I can for instance see myself using it in the future.

packages/palette/package.json
```diff
@@ -123,10 +124,12 @@
     "react-lazy-load-image-component": "^1.4.1-beta.0",
     "react-live": "^1.12.0",
     "react-powerplug": "^1.0.0",
+    "react-remove-scroll": "^2.3.0",
     "react-spring": "^8.0.27",
     "styled-bootstrap-grid": "1.0.4",
     "styled-system": "^3.1.11",
-    "trunc-html": "^1.1.2"
+    "trunc-html": "^1.1.2",
+    "use-cursor": "^1.2.0"
```
damassi, line 132: 👍

dzucconi:
I'll just add within this PR, good call 👍

dzucconi:
Going to try to lower the sensitivity on these bar chart visual specs.

packages/palette/src/elements/Modal/__tests__/ModalBase.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      await keydown("Tab", false)
+      expect(document.activeElement.id).toEqual("foo")
+      // Shift+tab backwards
+      await keydown("Tab", true)
+      expect(document.activeElement.id).toEqual("baz")
+      await keydown("Tab", true)
+      expect(document.activeElement.id).toEqual("bar")
+      await keydown("Tab", true)
+      expect(document.activeElement.id).toEqual("foo")
+      await keydown("Tab", true)
```
damassi, line 71: Nice

artsyit:
:rocket: PR was released in [`v9.9.0`](https://github.com/artsy/palette/releases/tag/v9.9.0) :rocket:

PR #694, opened 2020-05-27
Adjusts Chromatic settings

dzucconi:
Trying to see if I can get the spurious diff reviews to go away.

### DIFF
--- a/packages/palette/src/elements/BarChart/BarChart.story.tsx
+++ b/packages/palette/src/elements/BarChart/BarChart.story.tsx
@@ -64,7 +64,7 @@ storiesOf("Components/BarChart", module)
         </Box>
       )
     },
-    { chromatic: { delay: 500, diffThreshold: 0.2 } }
+    { chromatic: { delay: 750 } }
   )
 
   .add(
@@ -123,7 +123,7 @@ storiesOf("Components/BarChart", module)
         </Box>
       )
     },
-    { chromatic: { delay: 500, diffThreshold: 0.2 } }
+    { chromatic: { delay: 750 } }
   )
 
   .add(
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/BarChart/BarChart.story.tsx
```diff
@@ -342,5 +342,5 @@ storiesOf("Components/BarChart", module)
         </Box>
       )
     },
-    { chromatic: { delay: 500, diffThreshold: 0.2 } }
```
dzucconi, line 345: @damassi why were all the charts made to be more sensitive?

packages/palette/src/elements/Modal/Modal.story.tsx
```diff
@@ -4,43 +4,63 @@ import { Button } from "../Button"
...
+    "Wide modal",
+    () => {
+      return (
+        <Modal show title="Modal Title" isWide onClose={() => null}>
+          Some example content
+        </Modal>
+      )
+    },
+    { chromatic: { delay: 500 } }
+  )
```
dzucconi, line 66: OK well these look good now and just have to be accepted this one time

dzucconi:
Ah OK I kept bumping the delay on the bar chart but they rest consistently at `left: 29.7031px` now in the diiff. Will lower those.

packages/palette/src/elements/BarChart/BarChart.story.tsx
damassi: Lol are they? Not sure 🤦

dzucconi: 😆 https://www.chromatic.com/docs/threshold

> Chromatic’s default threshold is .063 which balances high visual accuracy with low false positives (for example, from artifacts like anti-aliasing). 0 is the most accurate. 1 is the least accurate.

I think all of the places where these were failing weren't due to the diffThreshold but rather views animating in. (So like the bar chart rising, the modal moving into the frame, etc)

artsyit:
:rocket: PR was released in [`v9.8.1`](https://github.com/artsy/palette/releases/tag/v9.8.1) :rocket:

Issue #696, opened 2020-05-27
Investigate inability to utilize styled(Button)


dzucconi:

damassi:
+1

PR #697, opened 2020-05-27
Fixes mixed-in classnames coming from styled-components

dzucconi:
Closes https://github.com/artsy/palette/issues/696

Weird that this uses class names instead of just props for handling these states?

### DIFF
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -187,6 +187,7 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
       longestText,
       weight,
       onClick,
+      className,
       ...rest
     } = this.props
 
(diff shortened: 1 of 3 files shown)


dzucconi:
Lemme add a quick spec for this too 👓

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -187,6 +187,7 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
       longestText,
       weight,
       onClick,
+      className,
```
damassi, line 190: Nice and easy

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -195,10 +196,12 @@ export class ButtonBase extends Component<ButtonBaseProps & SansProps> {
 
     return (
       <Container
-        {...rest}
-        className={[loadingClass, disabledClass].join(" ")}
+        className={[loadingClass, disabledClass, className]
```
damassi, line 199: Yeah, no clue now.... I see that we're using props [here](https://github.com/artsy/palette/blob/9e98e47615c0197db2ea980586097dae196abe29/packages/palette/src/elements/Button/Button.tsx#L262) but not too sure about the reasoning behind this line.

artsyit:
:rocket: PR was released in [`v9.8.1`](https://github.com/artsy/palette/releases/tag/v9.8.1) :rocket:

PR #701, opened 2020-05-29
[Delete] Remove Menu component and move to Force

damassi:
Depends on https://github.com/artsy/force/pull/5679 being merged first.

This removes our menu component as its only used in the force navbar and is more appropriate there.


dzucconi:
Woops, somehow I did not see this PR. Will rebase in a bit.

eessex:
We are using this in horizon now! https://github.com/artsy/horizon/blob/master/app/javascript/components/MainLayout.tsx#L72

dzucconi:
@eessex alright thanks for pointing that out 🤷‍♂️  — we'll probably want to undo that extraction in force @damassi? Closing for now.

PR #702, opened 2020-05-29
adding `endash` and a `range` helper

pvinis:
After the discussion [here](https://artsy.slack.com/archives/C9SATFLUU/p1590685677022700) and looking in force, we have a loooot of places where we do ranges by just adding `minus` instead of `endash`.

While changing it in one place (the pricing filter), I noticed that I actually missed the test because it looks very similar in some fonts. To solve this problem for the devs, I thought we can have the `endash` in palette, as well as a `range` func to just us like `range("from", "to")` that will output `"from – to"` (with an endash in the middle), but we can still just use it like `` `from ${endash} to` ``


packages/palette-docs/content/docs/elements/data-visualization/BarChart.mdx
```diff
@@ -253,7 +253,7 @@ not relativity to other graphs.
           {
             value: 3000,
             highlightLabel: {
-              title: "$30,000–$80,000",
+              title: range("$30,000", "$80,000"),
```
damassi, line 256: Nice

packages/palette/src/elements/Typography/Typography.test.tsx
```diff
@@ -1,7 +1,8 @@
 import React from "react"
 import renderer from "react-test-renderer"
+import { range } from "../../helpers"
 import { themeProps } from "../../Theme"
-import { Display, Sans, Serif, Text } from "../Typography"
+import { Display, endash, Sans, Serif, Text } from "../Typography"
```
damassi, line 5: Also nice idea!

packages/palette/src/helpers/typography.ts
```diff
@@ -0,0 +1,7 @@
+import { endash } from "../elements/Typography"
+
+/**
+ * A helper func to create a range string out of two strings, by putting an endash between them.
+ */
+export const range = (left: string, right: string): string =>
+  `${left} ${endash} ${right}`
```
damassi, line 7: Totally up to your discretion, but maybe this can live directly in the `Typography` component? Then, later, if we find a lot of util-like typography functions being used we can extract into a file.

pvinis: I kinda like to have them separate, seeing the rest of the structure of the repo. I find it harder to split a big file than having more small files 🤔.

artsyit:
:rocket: PR was released in [`v9.11.0`](https://github.com/artsy/palette/releases/tag/v9.11.0) :rocket:

PR #703, opened 2020-05-29
Clickable

dzucconi:
Nothing too exciting here. I find this kind of thing useful — we shouldn't be attaching click handlers to divs without things like `role="button" tabIndex={0}` + handling keyboard invocation, etc. This makes it much easier to just use buttons where things are actually buttons. In the use-case that triggered this PR, I'd like to make an otherwise normal header clickable/tappable.


packages/palette-docs/content/docs/components/Clickable.mdx
```diff
@@ -0,0 +1,14 @@
...
+
+`Clickable` Clickable is a utility component useful for wrapping things like
+`div`s without having to deal with the requirements to make the `div`
+accessible. It is the web-equivalent of `TouchableWithoutFeedback`. It supports
+the same system props as `<Box />`
+
+<Playground>
+  <Clickable onClick={() => alert("Clicked")}>
+    <Sans size="4">Click</Sans>
+  </Clickable>
```
damassi, line 13: Really like this idea

packages/palette/src/elements/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,21 @@
...
+ * ClickableProps
+ */
+export type ClickableProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
+  BoxProps
+
+/**
+ * Clickable is a utility component useful for wrapping things like <div>s
+ * without having to deal with the requirements to make the <div> accessible.
+ */
+export const Clickable = styled.button<ClickableProps>`
```
damassi, line 15: might want to import `styled` from our [primitives](https://github.com/artsy/palette/blob/master/packages/palette/src/platform/primitives.ts#L16) as this component can be cross-platform perhaps.

packages/palette/src/elements/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react"
```
damassi, line 1: Will want to create an iOS stub `Clickable.ios.tsx` so that Eigen doesn't automatically import .js, encounter possible web content, and break. Eventually, if this is shared / extended, we can collapse these two files into one.

packages/palette/src/elements/Clickable/Clickable.tsx
dzucconi: I could also just alias `TouchableWithoutFeedback` which should basically be the same thing?

packages/palette/src/elements/Clickable/Clickable.tsx
dzucconi: Yeah there's an empty one in here

damassi: cool, just wanted to be sure 👍

packages/palette/src/elements/Clickable/Clickable.tsx
damassi: Re primitives, it's just a pattern where where import from there because thats how our universal component system works. We haven't been diligent with this in the past and its led to errors in Eigen, but it's also admittedly hard to remember.

dzucconi:
I think this kind of component isn't relevant for Eigen since it's a pretty web-specific it solves. The only time this would be an issue is if some other primitive used on Eigen used this in which case it should alias TouchableWithoutFeedback. I'm going to merge for now and we can alias if needed.

artsyit:
:rocket: PR was released in [`v9.12.0`](https://github.com/artsy/palette/releases/tag/v9.12.0) :rocket:

PR #706, opened 2020-06-01
Update avatar component to use Sans

xtina-starr:
While working on the Collect Resume header, @jpotts244 and I noticed that the avatar component uses Sans instead Serif (which the component has currently). This PR updates the Avatar component to Sans which we want globally.

More context in this [slack thread](https://artsy.slack.com/archives/C9YNS4X32/p1591042672438200?thread_ts=1591042646.437900&cid=C9YNS4X32).

**Before:**
Screen Shot 2020-06-01 at 5 45 43 PM

**After:**
Screen Shot 2020-06-01 at 5 46 01 PM

### DIFF
--- a/packages/palette/src/elements/Avatar/Avatar.shared.tsx
+++ b/packages/palette/src/elements/Avatar/Avatar.shared.tsx
@@ -1,14 +1,14 @@
 import React, { ImgHTMLAttributes } from "react"
 import { color } from "../../helpers/color"
 import { styledWrapper } from "../../platform/primitives"
-import { SerifSize } from "../../Theme"
+import { SansSize } from "../../Theme"
 import { Flex } from "../Flex"
-import { Serif } from "../Typography"
+import { Sans } from "../Typography"
 
 export interface SizeProps {
   [key: string]: {
     diameter: string
-    typeSize: SerifSize
+    typeSize: SansSize
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Avatar/Avatar.shared.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { SansSize } from "../../Theme"
 import { Flex } from "../Flex"
-import { Serif } from "../Typography"
+import { Sans } from "../Typography"
 
 export interface SizeProps {
   [key: string]: {
     diameter: string
-    typeSize: SerifSize
+    typeSize: SansSize
```
xtina-starr, line 11: @zephraph one thing I'm not sure about is how we should setup the `Size` mapping for `xs`, `sm`, and `md`. Do we have this documented anywhere?

xtina-starr: I'm referring to [these](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Avatar/Avatar.shared.tsx#L16-L29).

zephraph: The type sizes or the avatar image size itself? I assume the type sizes would remain the same. If not, we could probably those from figma... let me see if I can dig that up.

zephraph: The designs aren't very exhaustive at this point. Let's just leave the sizes as is, we can iterate on it if design would like different sizes.

artsyit:
:rocket: PR was released in [`v9.13.0`](https://github.com/artsy/palette/releases/tag/v9.13.0) :rocket:

jpotts244:
🚀!

Issue #704, opened 2020-05-29
Overall system mixin strategy


dzucconi:
Opening this to get consensus on how to proceed.

It feels to me like the overall design of what system props are available on which components is unintuitive — but it's also possible I don't have a good mental model of the overall strategy with regard to primitives.

Here's some of the things that bit me in the past hour:

* Box has `bg` but Flex doesn't
* Flex has `top/bottom` but not `left/right`
* Typography supports `as` aliased as `element` but nothing else does

A good first step I think would be to move the flex mixin into `Box`, and then have `Flex` just become `Box` with `display: flex` as default.

Additionally I'd be interested in what is required to support `as` universally on our primitive elements.

damassi:
> A good first step I think would be to move the flex mixin into Box, and then have Flex just become Box with display: flex as default.

I fully endorse this. 

Some history around Palette is that when @zephraph and I first got it off the ground, it was patched into a flury other first-ever tasks, such as our new ssr relay pages and architecture. It was also a gamble that Katarina helped us advocate for with considerable risk. Since then, things have been non-stop for the both of us. 

I think now with you on board we're in a good place to revisit what we've done and make it sturdy, as well as cleanup. For example: 

> Flex has top/bottom but not left/right

The history of this is someone was using flex, needed top, and opened a quick PR to add it. Same with bottom. So far noonoe has needed left or right, until you stumbled upon it. All of this is ripe for cleanup.

zephraph:
While I agree with the direction I firmly believe we shouldn’t do it on the current version of styled-systems.

We desperately need to upgrade. The latest version of styled-system has a drastically reduced set of helpers that makes concerted effort towards providing consistent props a relatively trivial matter. 

I know it’s gonna be hell to upgrade, but we _really_ should. Take a look at my WIP pr. I’ll try to get back around to that next week.

damassi:
I've also tried to upgrade but found our `space` scale stuff broke across the board in scary way. I can dig up my PR if its not listed in @zephraph's.

dzucconi:
Interesting, something to do with us using unitless values in the theme? Could convert them to px and then have the space function do the parseInt? (I'm taking a look at this now as well)

dzucconi:
https://styled-system.com/guides/migrating/

> Number values are no longer converted to strings with px units. Most CSS-in-JS libraries now handle this, but if you need to use string values with units, provide them in your theme object and in prop values.

damassi:
>  unitless values in the theme

[...]

PR #707, opened 2020-06-03
Unifies Box/Flex

dzucconi:
Picking up on the initial impetus behind https://github.com/artsy/palette/issues/704

We might have a few minor issues surrounding the flexGrow/Shrink props (which are now required to be numeric). I noticed that since `size` was being passed to the underlying Flex in Avatar, that was causing an issue because space ships with a `size` mixin which sets width/height. Maybe we just remove that, actually?

<details>
  <summary>📦 Published PR as canary version: <code>10.1.1-canary.707.11364.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@10.1.1-canary.707.11364.0
  # or 
  yarn add @artsy/palette@10.1.1-canary.707.11364.0
  ```
</details>


dzucconi:
Ah, I see that also hits the Typography component.

I'm also thinking we can *dramatically* simplify that once we fix the font declarations so, maybe that's up next.

packages/palette/src/elements/Avatar/Avatar.shared.tsx
```diff
@@ -93,11 +92,11 @@ export const BaseAvatar = ({
 }
 
 /** InitialsHolder */
-export const InitialsHolder = styledWrapper(Flex)<{ size: SizeKey }>`
+export const InitialsHolder = styledWrapper(Flex)`
   background-color: ${color("black10")};
-  border-radius: ${props => sizeValue(props.size).diameter};
   text-align: center;
   overflow: hidden;
+  border-radius: 50%;
```
zephraph, line 99: So to recap this change, `size` was being passed down to the `InitialsHolder` only to calculate the radius, but because it's just supposed to be round `border-radius: 50%` achieves the same thing thereby making the passing down of size to this component unnecessary?

packages/palette/src/elements/Box/Box.tsx
```diff
@@ -5,81 +5,45 @@ import { styled as primitives } from "../../platform/primitives"
...
-  display,
-  height,
-  left,
-  minWidth,
-  maxWidth,
-  minHeight,
-  maxHeight,
+  border,
+  color,
+  flexbox,
```
zephraph, line 42: Does this change the default `display` of this component to be `flex`?

packages/palette/src/elements/Flex/Flex.tsx
```diff
@@ -1,129 +1,18 @@
...
- * A utility component that encapsulates flexbox behavior
+ * Flex is Box with display: flex
  */
-export const Flex = primitives.View<FlexProps>`
-  display: flex;
-  ${flexMixin}
-  ${flexGrow}
-  ${flexShrink}
-`
+export const Flex = styledWrapper(Box)<FlexProps>``
```
zephraph, line 12: 👍

packages/palette/src/elements/Typography/Typography.tsx
```diff
@@ -26,9 +26,10 @@ import {
   MaxWidthProps,
   space,
   SpaceProps,
-  style,
   textAlign,
   TextAlignProps,
+  verticalAlign,
+  VerticalAlignProps,
```
zephraph, line 32: `verticalAlign` is included in the new `layout` helper which we should use instead. 

> The layout utility includes style props for `width`, `height`, `display`, `minWidth`, `minHeight`, `maxWidth`, `maxHeight`, `size`, `verticalAlign`, `overflow`, `overflowX`, and `overflowY`.

https://styled-system.com/api#layout

zephraph: (Likewise, a lot of the other props can be replace with `typography` )

https://styled-system.com/api#typography

zephraph:
I think the type failures that you're getting here is what I was discussing with Orta. 

The keys for the typesizes in the theme file are defined as "1", "2", etc but get simplified to `1`, `2`, etc automatically by TS. I'm not sure exactly what to do about that yet.

packages/palette/src/elements/Typography/Typography.tsx
dzucconi: So, I figured I'd update Typography on it's own because it can mostly go away with fixing the font-face declarations.

packages/palette/src/elements/Box/Box.tsx
dzucconi: Nah

packages/palette/src/elements/Avatar/Avatar.shared.tsx
dzucconi: Yeah, it was doing it manually, and size is included with space so it was overwriting it and setting: `width: xs; height: xs`

dzucconi:
Hm, ok yeah those errors are specifically related to TypeScript 3.9.3. Is that an actual bug?

We can maybe address this when we implement the new typography scale and work out a plan for deprecating the old styles (maybe convert them into non-numeric strings and leave it alone)

zephraph:
From my conversation w/ @orta it sounded like it was intentional (because it's fundamentally equivalent to TS under the hood). 

We could get around it by broadening the typography props to accept any string.

orta:
oh cool - Damon's back, hiiiii

damassi:
@dzucconi - so to be clear, this just collapses the shared code and you suggest we continue using `<Flex>` and `<Box>` for markup clarity, yah? Or just `<Box>`?

dzucconi:
> we continue using `<Flex>` and `<Box>` for markup clarity, yah? Or just `<Box>`

I don't think this matters much. I'd probably exclusively use `<Box>`. I suppose there's something to be said about components that are shared with iOS and having to exclusively use `<Flex>` for those?

dzucconi:
So looking at force/eigen. There's 3 places in each that use the flexGrow="string" syntax which would no longer be supported. Simple fix to upgrade for each but technically this release would be a breaking change.

artsyit:
:rocket: PR was released in [`v11.0.0`](https://github.com/artsy/palette/releases/tag/v11.0.0) :rocket:

PR #708, opened 2020-06-05
allow customizing border color for BorderBox

oxaudo:
This might be controversial since allows to expand the variation on boxes, but we do already have a way to do it by extending with Styled components and we do allow to customize spacing. So seems logical to allow to customize `borderColor` as well.

This is for the https://github.com/artsy/volt/pull/4357

### DIFF
--- a/packages/palette-docs/content/docs/elements/layout/BorderBox.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/BorderBox.mdx
@@ -13,6 +13,7 @@ name: BorderBox
 <Playground title="BorderBox">
   <>
     <BorderBox mb={2}>One box...</BorderBox>
-    <BorderBox>Pushes down another</BorderBox>
+    <BorderBox mb={2}>Pushes down another</BorderBox>
+    <BorderBox mb={2} borderColor="red100">And another with border color</BorderBox> 
   </>
 </Playground>
--- a/packages/palette/src/elements/BorderBox/BorderBoxBase.tsx
+++ b/packages/palette/src/elements/BorderBox/BorderBoxBase.tsx
@@ -8,6 +8,8 @@ import { Flex, FlexProps } from "../Flex"
 import {
   background,
   BackgroundProps,
+  border,
+  BorderProps,
   height,
   HeightProps,
   maxWidth,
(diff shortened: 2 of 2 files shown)


oxaudo:
Really not sure what's up with specs here. Don't think it is related to my change in any ways:(

dzucconi:
Re: specs — just a timeout — re-running.

I accepted the UI specs, which are fine (I need to tweak both of those to stop requiring the approval)

oxaudo:
Hm... Have rerun specs couple of time and no success:( There is some console error that talks about modal flow. But it seems to be consistently timing out after Paginations spec. Any ideas on how to debug those? Locally all pass for me even though I get the same Modal warning error. Pagination test does take `32.284s` locally though...

dzucconi:
Yeah, I don't think those warnings are related (they are just warnings, and I couldn't figure out what on Earth it was complaining about since they are wrapped in act statements already... though it deserves a look at some point).

It *looks* like the trouble starts with

```
 PASS  src/elements/Dialog/__tests__/Dialog.test.tsx (112.061s)
```

@damassi @zephraph any ideas?

damassi:
@dzucconi - not too sure without digging in. We actually have this suppressed in Force: https://github.com/artsy/force/blob/master/src/v2/jest.envSetup.ts#L72-L81

dzucconi:
I can't exactly see a reason why this would matter, but this PR is from a fork?

oxaudo:
Yea... I do have all the repositories forked originally. Did not think that was an issue just yet:)

dzucconi:
Interestingly: I rebased this PR and while although the rest of the spec timings look normal, it's just hanging after pagination: https://app.circleci.com/pipelines/github/artsy/palette/2397/workflows/3b9be46a-81c8-46aa-aa5c-f48277334ea2/jobs/11320/steps

starsirius:
Hmm interesting... Let me give that a try with `--runInBand` or `-w=2`. 🤞

starsirius:
Well.. They, including this PR, all [passed](https://circleci.com/gh/artsy/palette/tree/oxaudo-updates). 🤷‍♂️   Noticed without `--runInBand` or `-w=2`, each jest test runs a lot slower tho.

oxaudo:
🍏 ! Thanks guys:)

Merging this since did not hear any objections on change here.

artsyit:
:rocket: PR was released in [`v10.1.0`](https://github.com/artsy/palette/releases/tag/v10.1.0) :rocket:

PR #711, opened 2020-06-16
BorderBox should support custom padding

oxaudo:
I think somewhere through refactor the space prop for BorderBox was lost. And seems logical to me that it should be supported for BorderBox.

In fact arguably we should support all the properties there that Box supports + border props. But I only needed spacing for now. So adding it here to fix this view:

### DIFF
--- a/packages/palette-docs/content/docs/elements/layout/BorderBox.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/BorderBox.mdx
@@ -14,6 +14,6 @@ name: BorderBox
   <>
     <BorderBox mb={2}>One box...</BorderBox>
     <BorderBox mb={2}>Pushes down another</BorderBox>
-    <BorderBox mb={2} borderColor="red100">And another with border color</BorderBox> 
+    <BorderBox mb={2} px={3} py={1} borderColor="red100">And another with border color and irregular padding</BorderBox> 
   </>
 </Playground>
--- a/packages/palette/src/elements/BorderBox/BorderBoxBase.tsx
+++ b/packages/palette/src/elements/BorderBox/BorderBoxBase.tsx
@@ -1,11 +1,16 @@
 // @ts-ignore
 import React from "react"
-import { border, BorderProps } from "styled-system"
+import {
+  border,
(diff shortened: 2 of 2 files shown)


dzucconi:
It looks like what's happening here is the explicit padding declaration is overriding the system function. What you can do instead of mixing space in again is set the `defaultProps` property and remove the explicit padding.

```tsx
BorderBoxBase.defaultProps = {
  p: 2
}
```

Though, I suppose I don't mind this either actually — I forget how a default `p` value interacts with `px/py`

packages/palette-docs/content/docs/elements/layout/BorderBox.mdx
```diff
@@ -14,6 +14,6 @@ name: BorderBox
   <>
     <BorderBox mb={2}>One box...</BorderBox>
     <BorderBox mb={2}>Pushes down another</BorderBox>
-    <BorderBox mb={2} borderColor="red100">And another with border color</BorderBox> 
+    <BorderBox mb={2} p={1} borderColor="red100">And another with border color and tight padding</BorderBox>
```
dzucconi, line 17: Could you just see how setting a default `p: 2` interacts if you utilize px/py in a story?

oxaudo:
Off @dzucconi .... Yea, I see the default applied in some other components. Doing it here seems to overwrite the padding defined in the component. Here is how it looks with the default:

Screen Shot 2020-06-16 at 10 56 51 AM

and that is how it looks with current implementation here:

Screen Shot 2020-06-16 at 10 57 58 AM

Kinda ugly but works...

dzucconi:
I see, ok makes sense to me. Thanks for bearing with. The other way to do this would be to take the boxMixin and apply it after the padding declaration in the styles instead of wrapping Flex. But I think this works fine. Thanks!

oxaudo:
Thanks for updating the label here. Forgot about that part!

artsyit:
:rocket: PR was released in [`v11.0.2`](https://github.com/artsy/palette/releases/tag/v11.0.2) :rocket:

PR #714, opened 2020-06-17
[web] Implements new typographic styles

dzucconi:
Re: [FX-1999](https://artsyproduct.atlassian.net/browse/FX-1999) & [FX-1998](https://artsyproduct.atlassian.net/browse/FX-1998)

Tokens: [figma](https://www.figma.com/file/YJp6VUm7QktPJua6zAcduX/Palette-Design-System?node-id=821%3A332)
Treatments: [figma](https://www.figma.com/file/YJp6VUm7QktPJua6zAcduX/Palette-Design-System?node-id=821%3A55)

This introduces a component called `Text` — which implements all of the codified type treatments as styled-system variants as well as supports the typographic tokens through the normal styled-system typographic functions.

I need to double check some things in iOS, and will leave this in a draft state until we deploy the new fonts stylesheet, which this depends on.

### Custom styles

<details>
  <summary>📦 Published PR as canary version: <code>11.1.2-canary.714.11580.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@11.1.2-canary.714.11580.0
  # or 
  yarn add @artsy/palette@11.1.2-canary.714.11580.0
  ```
</details>


packages/palette/src/Theme.tsx
```diff
@@ -95,6 +101,11 @@ export const themeProps = {
 
   fontFamily,
 
+  fonts: TEXT_FONTS,
+  fontSizes: TEXT_FONT_SIZES,
+  letterSpacings: TEXT_LETTER_SPACING,
+  lineHeights: TEXT_LINE_HEIGHTS,
```
dzucconi, line 107: These follow the styled-system theme specification. There's a type they export but we can't use it at the moment.

packages/palette/src/elements/Text/Text.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+        lineHeight: "solid",
+        letterSpacing: "tightest",
+      },
+      {
+        fontFamily: "serif",
+        fontSize: "55px",
+        lineHeight: "solid",
+        letterSpacing: "tightest",
+      },
+    ]
```
dzucconi, line 151: These specs were taken from the custom examples in the figma file.

zephraph:
React Native's main component for rendering text is called `Text`. Do you think that'll be confusing or conflict?

packages/palette/src/elements/Text/Text.tsx
```diff
@@ -0,0 +1,243 @@
...
+  size9: "32px",
+  size8: "28px",
+  size7: "24px",
+  size6: "18px",
+  size5: "16px",
+  size4: "15px",
+  size3: "14px",
+  size2: "13px",
+  size1: "12px",
+}
```
dzucconi, line 39: Explicitly named these with `sizeN` to avoid the scale index confusion and to line up with the naming convention design is using.

dzucconi:
@zephraph yeah I'm aware and kind of did this intentionally, because I couldn't come up with a better name. I'm open to suggestions here.

packages/palette/src/elements/Text/Text.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+  .add("As", () => {
+    const ELEMENTS = ["h1", "h2", "h3", "p"] as Array<
+      keyof JSX.IntrinsicElements
+    >
+
+    return (
+      <>
+        {ELEMENTS.map(element => {
+          return (
+            <Text key={element} as={element} variant="text">
```
dzucconi, line 117: `as` works totally fine.

packages/palette/src/elements/Text/Text.tsx
```diff
@@ -0,0 +1,243 @@
...
+ * which will be evaluated as px based on the given font-size.
+ */
+export const calculateLetterSpacing = (
+  fontSize: TextFontSize,
+  letterSpacing: TextLetterSpacing
+): number => {
+  const tracking = parseFloat(TEXT_LETTER_SPACING[letterSpacing])
+  const size = parseInt(TEXT_FONT_SIZES[fontSize], 10)
+  return size * tracking
+}
```
dzucconi, line 84: This is really the only annoying thing we have to do for iOS — converting the em value into a plain number. On web this gets mixed down into the correct px value.

packages/palette/src/elements/Text/Text.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+  > thead > tr > th {
+    text-align: left;
+    font-weight: normal;
+  }
+
+  > thead > tr > th,
+  > tbody > tr > td {
+    border-bottom: 1px solid ${themeGet("colors.black10")};
+    border-left: 1px solid ${themeGet("colors.black10")};
+    padding: ${themeGet("space.1")};
```
dzucconi, line 23: We should just be using `themeGet` instead of our custom helpers. For instance here the spacing value gets converted into px automatically, and in the case of colors, if we ever want to support a dark mode, we'll have to convert to this usage.

packages/palette/src/elements/Text/Text.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+/**
+ * Text
+ */
+export const Text = primitives.Text<TextProps>`
+  ${variant({ variants: TEXT_VARIANTS.small })}
+
+  @media (min-width: ${themeGet("breakpoints.0")}) {
+    ${variant({ variants: TEXT_VARIANTS.large })}
+  }
```
dzucconi, line 234: So, variants supports the responsive value array syntax just fine, but since we're only swapping values at the xs breakpoint I decided to just do this explicitly.

packages/palette/src/elements/Text/Text.story.tsx
damassi: @zephraph - I'm trying to remember the thinking behind our helpers -- I think it had to do with types, as `themeGet` isnt' type-safe.

dzucconi: as in you could pass it any string? We could also just also just wrap it with types for each respective theme key. The critical aspect here is that we get the value through the theme in the props, which could change (as in the case of dark mode) rather than the static theme.

packages/palette/src/elements/Text/Text.tsx
```diff
@@ -0,0 +1,243 @@
...
+  ColorProps & {
+    variant?: TextVariant
+    textColor?: ResponsiveValue<Color>
+  }
+
+const textColor = style({
+  prop: "textColor",
+  cssProperty: "color",
+  key: "colors",
+})
```
damassi, line 215: 👍

packages/palette/src/elements/Text/Text.tsx
damassi: How did this end up looking on RN? any impact when encountering `@media`?

damassi:
@dzucconi @zephraph - RN's main button primitive is also called `Button` but that hasn't caused any real problems in Eigen.

packages/palette/src/elements/Text/Text.story.tsx
zephraph: A lot of the faffing about we've done with helpers were for types. So long as we can preserve type safety I'm not opposed to using themeGet.

zephraph: That said, I _do_ like the explicitness of the helpers and I don't think using them as any significant disadvantage over `themeGet` (given that their implementation could just as simply be altered to use `themeGet` under the hood).

dzucconi: Yeah, I think that makes sense, I guess ultimately I wonder how often if at all we use these helpers outside of a styled-components context.

dzucconi:
Going to try publishing a canary. Can't tell if eigen doesn't like yalc or eigen is just being eigen.

damassi:
@dzucconi - heads up about [this command](https://github.com/artsy/palette/blob/master/packages/palette/package.json#L30). It just copies files over and works good.

ds300:
> @dzucconi @zephraph - RN's main button primitive is also called Button but that hasn't caused any real problems in Eigen, re Palette.

We often use the basic RN `Text` component for adding emphasis, to avoid doubling up on bottom margins added by `Sans`. As long as we refrain from adding bottom margins to this new `Text` I think it'll be fine, otherwise dealing with the name clashes would be a pain.

dzucconi:
@ds300 

> for adding emphasis
What does this mean?

> to avoid doubling up on bottom margins added by Sans
There should only be a bottom margin if you're explicitly using `m/my/mb` — not sure what you mean here?

ds300:
> @ds300
> 
> > for adding emphasis
> > What does this mean?
e.g. `<Sans weight="regular">Banksy has <Sans weight="medium">{numArtworks}</Sans> artworks for sale</Sans>`

> > to avoid doubling up on bottom margins added by Sans
> > There should only be a bottom margin if you're explicitly using `m/my/mb` — not sure what you mean here?

Apparently so! I distinctly remember debugging a visual issue where having nested `Sans` components caused extra bottom margin, and I must have falsely assumed the problem was that `Sans` itself had bottom margin, when it must have been coming from somewhere else. Sorry for the confusion, please ignore 😓

packages/palette/.storybook/preview-head.html
```diff
@@ -1,5 +1,5 @@
 <link
   rel="stylesheet"
   type="text/css"
-  href="https://webfonts.artsy.net/all-webfonts.css"
+  href="https://production-webfonts.artsy.net/all-webfonts.css"
```
dzucconi, line 4: Going to wait until Monday to migrate the DNS. Will change this back then.

packages/palette/src/elements/Text/Text.ios.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * to a px string. Since unitless line-heights are valid/normal, styled-system
+ * won't convert it to px.
+ */
+export const calculateLineHeight = (
+  fontSize: TextFontSize,
+  lineHeight: TextLineHeight
+): string => {
+  const size = parseInt(TEXT_FONT_SIZES[fontSize], 10)
+  return `${size * TEXT_LINE_HEIGHTS[lineHeight]}px`
+}
```
dzucconi, line 34: So I'm going to just make and handle this as a separate ticket but basically: we'll map over the variants using these helpers, and we'll have to also intercept the letterSpacing and lineHeight props in order to convert them dynamically based on the given font size.

dzucconi:
Merging — This shouldn't have any implications for anything existing: technically I've taken over the `Text` export but AFAICT that was only being exported for the benefit of some specs and isn't actually used in any real code.

artsyit:
:rocket: PR was released in [`v11.2.0`](https://github.com/artsy/palette/releases/tag/v11.2.0) :rocket:

PR #715, opened 2020-06-24
Disables knobs to remove dependency on emotion typings

dzucconi:
@ds300 noticed that after the Storybook upgrade, [typings from `@emotion/core` were being required in the built type declarations](https://app.renovatebot.com/package-diff?name=@artsy%2fpalette&from=11.0.2&to=11.1.0)

Apparently this is [due to addon-knobs](https://github.com/storybookjs/storybook/issues/8321). See also: https://github.com/emotion-js/emotion/issues/1800

addon-knobs it turns out is dead and being replaced with [addon-controls](https://github.com/storybookjs/storybook/tree/next/addons/controls#how-will-this-replace-addon-knobs) which OK, fine.

For now I'm just removing knobs because we only use it in a single place so it's value here isn't huge. I'd like to upgrade to Storybook 6 beta, and I started to go down that road but we run into some issues with our babelrc that I don't feel like dealing with at the moment.

### DIFF
--- a/packages/palette/.storybook/addons.js
+++ b/packages/palette/.storybook/addons.js
@@ -1,4 +1,3 @@
 import "@storybook/addon-actions/register"
-import "@storybook/addon-knobs/register"
 import "@storybook/addon-links/register"
 import "@storybook/addon-viewport/register"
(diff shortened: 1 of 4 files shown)


artsyit:
:rocket: PR was released in [`v11.1.1`](https://github.com/artsy/palette/releases/tag/v11.1.1) :rocket:

PR #716, opened 2020-06-25
Upgrades styled-bootstrap-grid; removes overflow: hidden

dzucconi:
So I'd like to use a vanilla `position: sticky` on an element that happens to be contained within a grid container and you can't actually stick anything if the parent container is `overflow: hidden`.

I can see that this property was introduced here: https://github.com/artsy/reaction/pull/1031 with the foreshadowing "It'll never be scrollable because of that (for better or worse)" 😈  but I'm not entirely sure why this was necessary?

Setting the `overflow` to `auto` on the artist page seems fine? If we *do* need it to be hidden then we can either wrap or mix in a system function for it.

@zephraph I know this was a while ago so I know the memory of this change is hazy but let me know if I'm missing something obvious here.

----

Also I upgraded styled-bootstrap-grid — that said, I have zero idea what the breaking changes in those versions might actually be: there's no changelog and https://github.com/dragma/styled-bootstrap-grid/issues/85 🤷  — the grid demo within the palette docs looks good.

<details>
  <summary>📦 Published PR as canary version: <code>11.8.1-canary.716.12006.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@11.8.1-canary.716.12006.0
  # or 
  yarn add @artsy/palette@11.8.1-canary.716.12006.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Grid/Grid.tsx
+++ b/packages/palette/src/elements/Grid/Grid.tsx
@@ -8,12 +8,7 @@ import {
 
 import styled from "styled-components"
 
-/**
- * TODO: v2 of `styled-bootstrap-grid` contains TS typings, but we need to
- * upgrade to styled-components 4 before it's possible to upgrade to v2.
- */
-
-/** Outter wrapper when using a grid */
+/** Outer wrapper when using a grid */
 export const Grid: any = styled(_Container)`
   overflow: hidden;
   max-width: ${props => props.theme.grid.breakpoints.xl};
--- a/packages/palette/src/elements/Grid/Grid.tsx
+++ b/packages/palette/src/elements/Grid/Grid.tsx
@@ -1,17 +1,14 @@
-import { color, flex, maxWidth, space, textAlign, width } from "styled-system"
-
 import {
   Col as _Col,
   Container as _Container,
   Row as _Row,
 } from "styled-bootstrap-grid"
-
 import styled from "styled-components"
+import { color, flex, maxWidth, space, textAlign, width } from "styled-system"
 
 /** Outer wrapper when using a grid */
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Grid/Grid.tsx
```diff
@@ -1,22 +1,14 @@
...
- * TODO: v2 of `styled-bootstrap-grid` contains TS typings, but we need to
- * upgrade to styled-components 4 before it's possible to upgrade to v2.
- */
-
-/** Outter wrapper when using a grid */
+/** Outer wrapper when using a grid */
 export const Grid: any = styled(_Container)`
-  overflow: hidden;
-  max-width: ${props => props.theme.grid.breakpoints.xl};
+  max-width: ${props => props.theme.grid.breakpoints.xl}px;
```
dzucconi, line 11: Minor bug here. Apparently this doesn't matter though, we never use this as an outer container on it's own.

dzucconi:
@damassi do you happen to know anything about this?

damassi:
@dzucconi - nope, that PR / issue doesn't seem familiar :/

zephraph:
@dzucconi that PR was a loooooong time ago, heh. I don't exactly recall the specifics of why that was necessary, but I'm more than happy to use a different solution if it works just as well. 

Were you able to try out the canary?

dzucconi:
@zephraph thx — yeah the canary looks good on the artist tabs. Going to merge 👍 👍

artsyit:
:rocket: PR was released in [`v11.8.1`](https://github.com/artsy/palette/releases/tag/v11.8.1) :rocket:

PR #717, opened 2020-06-26
Add cards from viewing rooms

pvinis:
Adding cards from https://www.figma.com/file/GrCeNUJbKpfZtgwHga64cC/Artsy-Viewing-Room-Entry-Points-Components?node-id=78%3A1259 for iOS and adding some placeholders for web.

Small with 1 artwork:
Screenshot 2020-06-29 at 14 25 59

Small with 2 artworks:
Screenshot 2020-06-29 at 14 26 39

Medium:
Screenshot 2020-06-29 at 14 27 26

Large:
Screenshot 2020-06-29 at 14 27 39


packages/palette/src/elements/Cards/LargeCard.ios.tsx
```diff
@@ -0,0 +1,63 @@
...
+          height: "100%",
+          opacity: 0.15,
+        }}
+      />
+      <Flex
+        style={{
+          position: "absolute",
+          bottom: 0,
+          left: space(1),
+          right: space(6),
```
ds300, line 42: would it make sense to use `maxWidth` here instead? Since the card is responsive, I'd assume we want a fixed break point on iPad.

packages/palette/src/elements/Cards/LargeCard.ios.tsx
```diff
@@ -0,0 +1,63 @@
...
+      >
+        <Sans size="5t" color={color("white100")}>
+          {title}
+        </Sans>
+        {!!subtitle && (
+          <Sans size="3t" color={color("white100")}>
+            {subtitle}
+          </Sans>
+        )}
+        <Spacer mt={1} />
```
ds300, line 53: prettty sure this should be `2`, as well as the spacing for the tag below.

packages/palette/src/elements/Cards/MediumCard.ios.tsx
```diff
@@ -0,0 +1,69 @@
...
+          <Sans size="3t" color={color("white100")}>
+            {subtitle}
+          </Sans>
+        )}
+        <Spacer mt={1} />
+      </Flex>
+      {!!tag && (
+        <Tag
+          {...tag}
+          style={{ position: "absolute", top: space(1), left: space(1) }}
```
ds300, line 64: I think all the padding in this file should be `15` rather than `space(1)`

packages/palette/src/elements/Cards/SmallCard.ios.tsx
```diff
@@ -0,0 +1,69 @@
...
+          aspectRatio: 1.5 / 1.0,
+          flexDirection: "row",
+          borderRadius: 4,
+          overflow: "hidden",
+        }}
+      >
+        <Flex flex={2} background={color("black10")}>
+          <Image source={{ uri: images[0] }} style={{ flex: 1 }} />
+        </Flex>
+        <Spacer mr={0.3} />
```
ds300, line 35: looks like these separators are supposed to be 2px to me. You can get around styled-system by making it a string `"2px"`

packages/palette/src/elements/Cards/Cards.story.tsx
```diff
@@ -0,0 +1,56 @@
+import { storiesOf } from "@storybook/react"
+import React from "react"
+import { LargeCard } from "./LargeCard"
+import { MediumCard } from "./MediumCard"
+import { SmallCard } from "./SmallCard"
+
+storiesOf("Components/Avatar", module)
```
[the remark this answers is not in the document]
pvinis: ooooh that's where this comes from. 🤦 good catch.

packages/palette/src/elements/Cards/CardTag.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react"
+import { CardTagProps } from "./CardTag.shared"
+
+/**
+ * `Tag` is used for the Cards, and is controlled by their `tag` prop.
+ */
+export const Tag: React.FC<CardTagProps> = () => {
+  return null
+}
```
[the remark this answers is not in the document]
pvinis: I thought we might use it to replace the tags in `Tags`, but yea, this makes sense. I'll rename, and when we use this to replace the other ones, we can rename

ds300:
also in the future it'd be cool to get screenshots in the PR description! :)

packages/palette/src/elements/Cards/MediumCard.ios.tsx
pvinis: For these ones we talked with tin and he said instead of 15 it's ok to use the `space(1)`, since we don't have a 15 in our regular usage.

packages/palette/src/elements/Cards/SmallCard.ios.tsx
pvinis: For this too, talking with tin we decided to use the closest available `space`, so 0.3 should be alright.

packages/palette/src/elements/Cards/LargeCard.ios.tsx
pvinis: Hm, let me try it there. My idea was that there will always be a grid or a list that will allocate some width to each item, and them the card will just occupy that space. 🤔

packages/palette/src/elements/Cards/MediumCard.ios.tsx
ds300: Ok, but for future reference we use 15px in quite a few places in eigen because in those places 10px looks too small and 20px looks too big :) I don't know whether this is one of those places, but IMO the spacing system is more like guide rails than hard rules and it's fine to do custom spacing when it _looks_ more 'right' than any of the predefined options.

ds300: also `15px` is just `0.5` + `1` in our design system values :p

packages/palette/src/elements/Cards/SmallCard.ios.tsx
ds300: actually now that I see the screenshots 0.3 looks too chunky to me and I'd recommend we keep it consistent with the designs, not to mention there's already a bunch of places in the app where we use this visual pattern at 2px so this would create inconsistency.

packages/palette/src/elements/Cards/LargeCard.ios.tsx
```diff
@@ -0,0 +1,63 @@
...
+          width: "100%",
+          height: "100%",
+          opacity: 0.15,
+        }}
+      />
+      <Flex
+        style={{
+          position: "absolute",
+          bottom: 0,
+          left: space(1),
```
ds300, line 41:
```suggestion
          left: space(2),
```

packages/palette/src/elements/Cards/SmallCard.ios.tsx
pvinis: I don't like using random values, but alright, I'll use the 2px.

pvinis:
I think now this should be ready. Thanks david for the careful suggestions! 🙌

ds300:
Nice work 🙌

artsyit:
:rocket: PR was released in [`v11.3.0`](https://github.com/artsy/palette/releases/tag/v11.3.0) :rocket:

PR #719, opened 2020-06-29
[FX-2067] Implements Text.ios

dzucconi:

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -20,7 +20,7 @@
     "prepublishOnly": "yarn clean && yarn compile && yarn type-declarations",
     "prettier-project": "yarn prettier --write 'src/**/*.{ts,tsx}'",
     "release": "auto shipit && node scripts/clean-up-after-auto.js",
-    "test": "yarn type-check && yarn jest",
+    "test": "yarn type-check && yarn jest --runInBand",
     "test:watch": "jest --watch --runInBand",
     "type-check": "tsc --noEmit --pretty && yarn type-check-ios && yarn type-check-scripts",
     "type-check-ios": "tsc -p tsconfig.ios.json --noEmit --pretty",
--- a/packages/palette/src/elements/Text/Text.ios.tsx
+++ b/packages/palette/src/elements/Text/Text.ios.tsx
@@ -1,34 +1,66 @@
+import React from "react"
+import { variant as systemVariant } from "styled-system"
+import { styled as primitives } from "../../platform/primitives"
+import { textMixin, TextProps } from "./Text.shared"
 import {
-  TEXT_FONT_SIZES,
-  TEXT_LETTER_SPACING,
-  TEXT_LINE_HEIGHTS,
-  TextFontSize,
-  TextLetterSpacing,
-  TextLineHeight,
-} from "./tokens"
+  calculateLetterSpacing,
+  calculateLineHeight,
+  isControlledFontSize,
+  isControlledLetterSpacing,
+  isControlledLineHeight,
+  TEXT_VARIANTS,
+} from "./tokens.ios"
 
 /**
- * em-units don't exist on React Native so we convert it to a number
- * which will be evaluated as px based on the given font-size.
+ * InnerText
  */
-export const calculateLetterSpacing = (
-  fontSize: TextFontSize,
-  letterSpacing: TextLetterSpacing
-): number => {
-  const tracking = parseFloat(TEXT_LETTER_SPACING[letterSpacing])
-  const size = parseInt(TEXT_FONT_SIZES[fontSize], 10)
-  return size * tracking
+const InnerText = primitives.Text<TextProps>`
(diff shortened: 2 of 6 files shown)


dzucconi:
@MounirDhahri @pvinis can I interest either of you in pairing on getting font-families supported on this?

pvinis:
mm I could do some pairing, sure. I like font stuff.

dzucconi:
@pvinis — So, I had been operating under the assumption that font families didn't work on eigen because of how the code was structured in the previous typography components. But in fact they work just fine! They just didn't work on web. [So utilizing the existing font names inside of eigen](https://github.com/artsy/palette/pull/719/commits/fbd046e8c46f5e8a30066bb5ce5b7b0f49e31526), the italics/bold options are picked up and the correct font faces are applied.

packages/palette/src/elements/Text/Text.ios.tsx
```diff
@@ -1,34 +1,66 @@
...
+        }
+      : {}),
+    ...(!variant && lineHeight && fontSize
+      ? // Possibly convert the lineHeight
+        {
+          lineHeight:
+            isControlledLineHeight(lineHeight) && isControlledFontSize(fontSize)
+              ? calculateLineHeight(fontSize, lineHeight)
+              : lineHeight,
+        }
```
dzucconi, line 54: The line-heights are unitless values, and the letter-spacing is specified as em — both for web. So we need to convert it based on the given font-size. We only do so if both values are themed ('controlled').

dzucconi:
Something is definitely wrong with CI here, these are timing out when there isn't a CircleCI status degradation.

Locally: 
```
Test Suites: 30 passed, 30 total
Tests:       1 skipped, 173 passed, 174 total
Snapshots:   0 total
Time:        8.73s, estimated 35s
Ran all test suites.
✨  Done in 25.22s.
```

dzucconi:
Our shared CircleCI config makes this difficult to edit on a per-project basis? Just setting the `runInBand` flag in the package to see if that helps.

packages/palette/src/elements/Text/Text.shared.tsx
```diff
@@ -14,7 +14,7 @@ import { TextVariant } from "./tokens"
  * Text
  */
 export type TextProps = TypographyProps &
-  ColorProps & {
+  Omit<ColorProps, "color"> & {
     variant?: TextVariant
     textColor?: ResponsiveValue<Color>
```
pvinis, line 19: Can we change this to just `color` actually, now that we have removed the one from `ColorProps`?

packages/palette/package.json
```diff
@@ -20,7 +20,7 @@
     "prepublishOnly": "yarn clean && yarn compile && yarn type-declarations",
     "prettier-project": "yarn prettier --write 'src/**/*.{ts,tsx}'",
     "release": "auto shipit && node scripts/clean-up-after-auto.js",
-    "test": "yarn type-check && yarn jest",
+    "test": "yarn type-check && yarn jest --runInBand",
```
pvinis, line 23: Should this stay here or is it a temporary fix? I'm not too confident yet with the inner workings of palette.

packages/palette/src/elements/Text/Text.shared.tsx
dzucconi: So actually `color` works fine as well. There's a long standing issue with styled-system's `ColorProps` that interferes with the native color prop. We omit it here but it still works as you'd expect.

packages/palette/package.json
dzucconi: I think this should stay unless we can pass args thru the CircleCI config. Either way it's fine to merge. Coming from https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server

dzucconi:
> I see that we can do fontSize="size12" to get the controlled one and we can also do fontSize={20} to get a non-controlled one. Is this what we want? If yes (I'm guessing the answer is yes here), I understand it, but I don't feel this is a great "API" for this. To be honest I would prefer something like size is for controlled like size={12} and fontSize is for regular font size like fontSize={30}, or something along those lines. What do you think?

Right — We're simply playing within the lines [drawn by styled-system's API](https://styled-system.com/api#typography). These conventions are pretty well battled-tested at this point, and where I agree there are places where one could argue for a more elegant API given the specific domain, where styled-system really shines is the consistent conventions stretched over the design system at large.

dzucconi:
Also it's worth mentioning that 99% of the time you're going to simply be using the variants and should not stray outside of them. The specs simply call for the possibility of fully custom typography for special cases — I suspect this will be even less common inside of the iOS app vs web.

Going to merge.

artsyit:
:rocket: PR was released in [`v11.4.0`](https://github.com/artsy/palette/releases/tag/v11.4.0) :rocket:

PR #720, opened 2020-06-30
Makes alt property required on web images

dzucconi:
Publishing a canary and if it's simple enough to knock down the errors I think we should merge — this is a bit of a no-brainer to enforce.

<details>
  <summary>📦 Published PR as canary version: <code>11.4.1-canary.720.11774.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@11.4.1-canary.720.11774.0
  # or 
  yarn add @artsy/palette@11.4.1-canary.720.11774.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Image/Image.tsx
+++ b/packages/palette/src/elements/Image/Image.tsx
@@ -12,7 +12,7 @@ export interface WebImageProps extends ImageProps {
   /** Flag for if image should be lazy loaded */
   lazyLoad?: boolean
   /** Alternate text for image */
-  alt?: string
+  alt: string
   /** A11y text label */
   ["aria-label"]?: string
   /** The title of the image */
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/Image/Image.tsx
```diff
@@ -39,6 +39,7 @@ export const Image = ({
 
 /** Props for a web-only ResponsiveImage component. */
 export interface WebResponsiveImageProps extends ResponsiveImageProps {
+  alt: string
```
dzucconi, line 42: Realizing this component isn't actually an `img` tag (obviously should be). So will probably have to roll off some of the alts in force once I'm done populating them. But will also have to re-implement this component as an actual image.

dzucconi:
This served it's purpose for now (driving out places to add alt tags) — but once we convert the ResponsiveImage to an actual image I think we should consider making it required, really wouldn't be much work.

PR #722, opened 2020-07-02
Supports a small entity header design via "smallVariant" prop

ashleyjelks:
https://artsyproduct.atlassian.net/browse/FX-2070

The Entity Header Palette component should support receiving a `smallVariant` prop that changes the layout and design of the component. When this prop is passed the component only renders the Entity name and Followed/Unfollowed State.

With the `smallVariant` prop on iOS:
Screen Shot 2020-07-02 at 3 35 45 PM
Screen Shot 2020-07-02 at 3 35 51 PM

Without the `smallVariant` prop on iOS:
Screen Shot 2020-07-02 at 1 15 55 PM

With the `smallVariant` prop on web:
Screen Shot 2020-07-02 at 11 23 31 AM

Without the `smallVariant` prop on web:
Screen Shot 2020-07-02 at 1 24 53 PM

In the Palette Playground:
Screen Shot 2020-07-02 at 1 30 38 PM

### DIFF
--- a/packages/palette/src/elements/EntityHeader/EntityHeader.tsx
+++ b/packages/palette/src/elements/EntityHeader/EntityHeader.tsx
@@ -16,6 +16,7 @@ interface EntityHeaderProps extends SpacerProps {
   name: string
   FollowButton?: JSX.Element
   onClick?: (event: React.MouseEvent<HTMLElement>) => void
+  small?: boolean
 }
 
 interface ContainerComponentProps {
@@ -35,6 +36,7 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
   name,
   meta,
   FollowButton,
+  small,
   ...remainderProps
 }) => {
   const ContainerComponent = href ? FlexLink : Flex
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/EntityHeader/EntityHeader.tsx
```diff
@@ -52,35 +54,68 @@ export const EntityHeader: SFC<EntityHeaderProps> = ({
...
+                  >
+                    •
+                  </Sans>
+                )}
+                <Box
+                  display="inline-block"
+                  onClick={event => {
+                    // Capture click event so that interacting with Follow doesn't
+                    // trigger Container's link.
+                    event.stopPropagation()
```
dzucconi, line 109: I can see this was here all ready but this just suggests to me that the structure of this component is odd — like is it appropriate UX to extend the hit area to the right of a borderless follow button? 🤔

ashleyjelks: But the Sans component has a div wrapper, so there are bound to the follow button even when its text. Is the suggestion here to add an invisible border around the FollowButton when its text and not an actual Button component?

dzucconi:
Looks good to me — feel free to merge.

`non-blocking comment`:

I do think `small` as an API doesn't quite fit. The *size* is pretty much the sam on both iOS and web. I kind of wonder if `details={false}` or `tombstone={false}` with a default set to `true` is more on the nose.

The design differences between iOS `small` and `not small` also don't quite fit with either naming. Is that older design with the larger follow CTAs going to continue to be used?

packages/palette/src/elements/EntityHeader/EntityHeader.tsx
dzucconi: Ah, I had assumed that this was an anchor/a tag with a nested button. Tapping the button would propagate the parent link click. Rather I suppose would be to link everything separately and the follow button as an actual button would be fine and wouldn't have to have an outer wrapper prevent the event propagation. I'd say just leave this alone, I guess just making a mental note.

ashleyjelks:
> Looks good to me — feel free to merge.
> 
> `non-blocking comment`:
> 
> I do think `small` as an API doesn't quite fit. The _size_ is pretty much the sam on both iOS and web. I kind of wonder if `details={false}` or `tombstone={false}` with a default set to `true` is more on the nose.
> 
> The design differences between iOS `small` and `not small` also don't quite fit with either naming. Is that older design with the larger follow CTAs going to continue to be used?

I agree that small isn't the right prop name, good point. And technically the text is larger than the when this prop isn't used so it's also inaccurate to call it small.

artsyit:
:rocket: PR was released in [`v11.6.0`](https://github.com/artsy/palette/releases/tag/v11.6.0) :rocket:

PR #723, opened 2020-07-03
[Part of GALL-2970] - adding a small card to web view

oxaudo:
To be able to work on GALL-2971 I need a small version of the card. For now only implementing that part so that I can use it in ViewingRoomsApp. (Small version)

I think design for all of those cards is here: https://www.figma.com/file/GrCeNUJbKpfZtgwHga64cC/Artsy-Viewing-Room-Entry-Points-Components?node-id=136%3A2022

### DIFF
--- a/packages/palette-docs/content/docs/elements/layout/Cards.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/Cards.mdx
@@ -3,7 +3,7 @@ name: Cards
 ---
 
 <Playground title="Small Card with one image">
-  <div style={{ width: '50%' }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <SmallCard
       images={[
         "https://www.fillmurray.com/400/400",
@@ -11,11 +11,11 @@ name: Cards
       title="This one has one image"
       subtitle="It's good"
     />
-  </div>
+  </Box>
 </Playground>
 
 <Playground title="Small Card with two images">
-  <div style={{ width: '50%' }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <SmallCard
       images={[
         "https://www.fillmurray.com/400/400",
@@ -24,11 +24,11 @@ name: Cards
       title="This one has 2 images"
       subtitle="It's good"
     />
-  </div>
+  </Box>
 </Playground>
 
 <Playground title="Small Card with three images">
-  <div style={{ width: '50%' }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <SmallCard
       images={[
         "https://www.fillmurray.com/400/400",
@@ -38,11 +38,11 @@ name: Cards
       title="This one has 3 images"
       subtitle="It's good"
     />
-  </div>
+  </Box>
 </Playground>
 
 <Playground title="Small Card with three images and tag">
-  <div style={{ width: '50%' }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <SmallCard
       images={[
         "https://www.fillmurray.com/400/400",
@@ -58,11 +58,11 @@ name: Cards
         textColor: "black100",
       }}
     />
-  </div>
+  </Box>
 </Playground>
 
 <Playground title="Medium Card with tag">
-  <div style={{ width: 200 }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <MediumCard
       image="https://www.fillmurray.com/400/300"
       title="Check this out"
@@ -74,11 +74,11 @@ name: Cards
         textColor: "black100",
       }}
     />
-  </div>
+  </Box>
 </Playground>
 
 <Playground title="Large Card with tag">
-  <div style={{ width: 200 }}>
+  <Box width="50%" p="2px" backgroundColor="yellow30">
     <LargeCard
       image="https://www.fillmurray.com/400/300"
       title="Check this out"
@@ -90,5 +90,5 @@ name: Cards
         textColor: "black100",
       }}
     />
-  </div>
+  </Box>
 </Playground>
--- a/packages/palette-docs/content/docs/elements/layout/Cards.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/Cards.mdx
@@ -2,42 +2,54 @@
 name: Cards
 ---
 
-<Playground title="Small Card with two images">
-  <div style={{ width: 200 }}>
+<Playground title="Small Card with one image">
+  <div style={{ width: '50%' }}>
     <SmallCard
       images={[
         "https://www.fillmurray.com/400/400",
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/Cards/CardTag.tsx
```diff
@@ -1,9 +1,33 @@
...
+}) => {
+  return (
+    <BorderBox
+      borderRadius="2"
+      overflow="hidden"
+      backgroundColor={color(bgColor) || bgColor}
+      borderColor={color(borderColor) || borderColor}
+      m={0}
+      p={0}
+      position="absolute"
```
oxaudo, line 24: I think for now we only want to display this absolutely positioned over the card. So was not sure why we needed the complications with passing `style` there. For now hardcoding this position here.

packages/palette/src/elements/Cards/SmallCard.ios.tsx
```diff
@@ -36,12 +36,9 @@ export const SmallCard: React.FC<SmallCardProps> = ({
         <Flex flex={1}>
           {images.length < 2 && <Flex flex={1} background={color("black10")} />}
           {!!images[1] && (
-            <Image
-              source={{ uri: images.length === 2 ? images[1] : images[1] }}
-              style={{ flex: 1 }}
-            />
+            <Image source={{ uri: images[1] }} style={{ flex: 1 }} />
```
oxaudo, line 39: This is just a small refactor. I think the check for the images length was redundant there.

packages/palette/src/elements/Cards/SmallCard.ios.tsx
```diff
@@ -36,12 +36,9 @@ export const SmallCard: React.FC<SmallCardProps> = ({
...
           )}
-          {images.length > 2 && (
+          {!!images[2] && (
```
oxaudo, line 41: Following the pattern from above.

packages/palette/src/elements/Cards/SmallCard.tsx
```diff
@@ -1,11 +1,83 @@
...
-  return null
+export const SmallCard: React.FC<SmallCardProps> = ({
+  images,
+  title,
+  subtitle,
+  tag,
+}) => {
+  return (
+    <Box width="100%" position="relative">
+      <Box width="100%" height="0" pt="67%" position="relative">
```
oxaudo, line 24: Was not sure if there is a better was to implement `ratio` in web views. But that seems to work - so rolling with it here. Suggestions are welcome though:)

packages/palette/src/elements/Cards/SmallCard.tsx
```diff
@@ -1,11 +1,83 @@
...
+        <Flex
+          flexDirection="row"
+          width="100%"
+          height="100%"
+          position="absolute"
+          top="0"
+          borderRadius="4px"
+          overflow="hidden"
+        >
+          <Box backgroundColor="black10" width="67%" overflow="hidden" mr="1px">
```
oxaudo, line 34: Copied this layout idea from iOs version. The case that the VR has no valid images does not seem to be handled. So not sure what should happen there. But I guess the header image is required from CMS for publishing VRs. So hopefully that will not be the case often if ever.

packages/palette/src/elements/Cards/SmallCard.tsx
pvinis: I did try the same thing for eigen here https://github.com/artsy/eigen/pull/3457/commits/8fdd1b2e6cdb87cdd0d47f2a7a3518710d435154#diff-89e2903a44c036391e4b31d92dba65faR8 bt thankfully RN has `aspectRatio`. I think for web this is a common way. If we end up needing this more, we could add an extra Layout component in the future.

packages/palette/src/elements/Cards/CardTag.tsx
pvinis: So the reason for this is because the Small card needs the left and top margin to be 10px, the Medium one needs 15px, and the Large needs 20px. In the ios versions of the cards you can see how `space(x)` differs on `<CardTag ..`, so we should support this additional styling here.

packages/palette/src/elements/Cards/SmallCard.tsx
dzucconi: This works. I'm going to add an [aspect ratio box component](https://github.com/auspices/eos/blob/master/src/AspectRatioBox/AspectRatioBox.tsx) in here soon. 🤷

packages/palette-docs/content/docs/elements/layout/Cards.mdx
```diff
@@ -2,42 +2,54 @@
...
+      ]}
+      title="This one has one image"
+      subtitle="It's good"
+    />
+  </div>
+</Playground>
+
 <Playground title="Small Card with two images">
-  <div style={{ width: 200 }}>
+  <div style={{ width: '50%' }}>
```
pvinis, line 18: I tried making the changes myself but you use your fork, and I have no idea how to add to that for a PR on the main repo.

Could we replace all the `<div style={{ width: '50%' }}>` with `<Box p="2" backgroundColor="orange" width="50%">`?
This way we can see the card inside that box a bit clearer? 🤔

packages/palette/src/elements/Cards/SmallCard.tsx
oxaudo: Refactor scheduled for the future (TM) :)

packages/palette-docs/content/docs/elements/layout/Cards.mdx
oxaudo: Ha:) I actually specifically changed the enclosed container to % so we can resize the page and see how the Card adapts. I think that will mimic more how it will be used in the wild. But I like your idea of coloring the boundary box. Will update in a bit!

pvinis: Yea, I noticed the `%`, and I thought it wasa  very nice touch :D

packages/palette/src/elements/Cards/CardTag.tsx
oxaudo: I've updated the display to give a background. Not sure if that is better since it's a bit confusing if it's intentional or a bug, but definitely gives a better idea of boundary and changes when resizing.

Not sure why I don't see an update with storybook here, but the look:
Screen Shot 2020-07-07 at 12 20 37 PM

oxaudo:
@pvinis - I think I've incorporated your suggestions here. Let me know if looks on now:)

pvinis:
let's goooooo 🙏 #mergeongreen

artsyit:
:rocket: PR was released in [`v11.7.0`](https://github.com/artsy/palette/releases/tag/v11.7.0) :rocket:

PR #724, opened 2020-07-07
Removes blanket focus disable

dzucconi:
This is a pretty presumptuous rule to have in here and we shouldn't be doing this anyway.

I ran into this because in trying to clean up how we integrate `:focus-visible` — this is sometimes winning in specificity depending on where I insert the CSS.


artsyit:
:rocket: PR was released in [`v11.6.1`](https://github.com/artsy/palette/releases/tag/v11.6.1) :rocket:

PR #725, opened 2020-07-07
Removes theme from config

dzucconi:
So this *also* seems to be creating issues.

https://github.com/storybookjs/storybook/issues/7379

Eigen appears to work once this is removed.

This isn't exactly a critical feature and this package wasn't even a specified dependency so I'm just removing.


artsyit:
:rocket: PR was released in [`v11.6.2`](https://github.com/artsy/palette/releases/tag/v11.6.2) :rocket:

PR #726, opened 2020-07-10
Implements AspectRatioBox

dzucconi:
Going to make some improvements in `<ResponsiveImage />` and would like to have this underlying it, which is a useful layout primitive to have around, in general.

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -124,6 +124,7 @@
     "d3-interpolate": "^1.3.2",
     "d3-shape": "^1.3.5",
     "luxon": "^1.15",
+    "proportional-scale": "3.0.1",
     "rc-slider": "^8.6.2",
     "react-lazy-load-image-component": "1.5.0",
     "react-live": "^1.12.0",
--- a/packages/palette/src/elements/AspectRatioBox/AspectRatioBox.tsx
+++ b/packages/palette/src/elements/AspectRatioBox/AspectRatioBox.tsx
@@ -1,0 +1,51 @@
+import { scale } from "proportional-scale"
+import React from "react"
+import { Box, BoxProps } from "../Box"
+
+/** AspectRatioBox */
+export type AspectRatioBoxProps = BoxProps & {
+  aspectWidth: number
+  aspectHeight: number
+  maxWidth: number
+  maxHeight: number
+}
+
+/** AspectRatioBox */
+export const AspectRatioBox: React.FC<AspectRatioBoxProps> = ({
+  aspectWidth,
+  aspectHeight,
+  maxWidth,
+  maxHeight,
+  children,
+  ...rest
+}) => {
+  const { width, paddingBottom } = scale({
+    width: aspectWidth,
+    height: aspectHeight,
+    maxWidth,
+    maxHeight,
+  })
+
+  return (
+    <Box
+      position="relative"
+      width="100%"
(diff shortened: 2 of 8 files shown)


packages/palette/src/elements/Cards/Cards.story.tsx
```diff
@@ -42,7 +42,7 @@ storiesOf("Components/Cards", module)
   .add("Large", () => {
     return (
       <LargeCard
-        image="https://www.fillmurray.com/400/300"
+        image="https://picsum.photos/seed/example/400/300"
```
dzucconi, line 45: lol sorry @pvinis — we need something with a seedable random image for the stories so that Chromatic doesn't diff it in every pull request

packages/palette/package.json
```diff
@@ -124,6 +124,7 @@
     "d3-interpolate": "^1.3.2",
     "d3-shape": "^1.3.5",
     "luxon": "^1.15",
+    "proportional-scale": "3.0.1",
```
dzucconi, line 127: One of mine, simple enough: https://github.com/dzucconi/proportional-scale/blob/master/src/scale.ts

packages/palette/src/elements/AspectRatioBox/AspectRatioBox.tsx
```diff
@@ -0,0 +1,51 @@
...
+      width="100%"
+      style={{ maxWidth: `${width}px` }}
+      {...rest}
+    >
+      <Box
+        position="relative"
+        width="100%"
+        height={0}
+        overflow="hidden"
+        paddingBottom={paddingBottom}
```
damassi, line 41: 😄

packages/palette/src/elements/index.tsx
```diff
@@ -1,4 +1,5 @@
 export * from "./Avatar"
+export * from "./AspectRatioBox"
```
damassi, line 2: soooooo minor -- this should go before avatar

dzucconi:
(just now improved the story a bit to demonstrate maxWidth — resizes to 100% if it's smaller than that)

packages/palette/src/elements/index.tsx
dzucconi: omg

dzucconi:
> only question is whether this is truly needed as a primitive, which increases palate's API surface area

Sure I'm fine with that. Lemme cherry pick off these story updates, regardless

oxaudo:
This looks closed. Does that mean that `AspectRatioBox` is not a palette concept? Are we adding it in Force? I think there are couple of places there that we can use it.

dzucconi:
@oxaudo I'm happy to re-open and expose this if you have an immediate use for it!

I think @damassi's concern about the API surface is a valid one, and I was just going to expose it when needed

oxaudo:
@damassi - this PR https://palette.artsy.net/elements/layout/cards/ basically did the same thing inline. So could have used that component there. The large card there is also specked with ratio: https://www.figma.com/file/GrCeNUJbKpfZtgwHga64cC/Artsy-Viewing-Room-Entry-Points-Components?node-id=78%3A1259 - So maybe good to have it as a component just for those cases? Not sure:)

damassi:
If we've got a few legit cases that we can refactor to use this component we can do that! Ideally we can do that all at once though so there's not multiple implementations floating around.

PR #730, opened 2020-07-14
Implements ResponsiveImage (and friends)

dzucconi:
The ultimate goal here is to just implement a new `ResponsiveImage` component that is a real `` tag, can be lazy, and supports all native `` props as well as any box props.

Nothing is exposed yet, hence the draft. I'd need to clean up the exports, basically.

I went with a parallel implementation where each piece is broken up into it's own useful primitive. We don't have to expose all of these for now but there is:
* `<ResponsiveBox />` (formerly AspectRatioBox)
* `<Skeleton />` (pulsing animation Box)
* new `<LazyImage />` (which is largely the same as the old, just avoids having to manually implement types and retains all the native callbacks)
* `<ResponsiveImage />` which is the composition of all of the above

<details>
  <summary>📦 Published PR as canary version: <code>11.8.2-canary.730.12074.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@11.8.2-canary.730.12074.0
  # or 
  yarn add @artsy/palette@11.8.2-canary.730.12074.0
  ```
</details>


packages/palette/src/elements/Box/Box.tsx
```diff
@@ -55,3 +53,28 @@ export const Box = primitives.View<BoxProps>`
...
+    if (BOX_PROP_REGEX.test(key)) {
+      boxProps[key] = props[key]
+      continue
+    }
+
+    notBoxProps[key] = props[key]
+  }
+
+  return [boxProps, notBoxProps]
+}
```
dzucconi, line 80: So yeah, the usecase here is that frequently one might want to implement a component in such a way as to split layout concerns with an outer box. This gives you a typesafe way to split off the box related props without having to manually pull them off.

packages/palette/src/elements/LazyImage/LazyImage.tsx
```diff
@@ -0,0 +1,126 @@
...
+    setMode(Mode.Error)
+    onError && onError(event)
+  }
+
+  const handleContextMenu = (
+    event: React.MouseEvent<HTMLImageElement, MouseEvent>
+  ) => {
+    preventRightClick && event.preventDefault()
+    onContextMenu && onContextMenu(event)
+  }
```
dzucconi, line 92: If any of these design system components utilize callbacks we need to be sure we aren't clobbering the native version and are passing through the real event.

packages/palette/src/elements/Skeleton/Skeleton.story.tsx
```diff
@@ -0,0 +1,13 @@
+import { storiesOf } from "@storybook/react"
+import React from "react"
+import { Skeleton } from "./Skeleton"
+
+storiesOf("Components/Skeleton", module).add("Basic", () => {
+  return (
+    <>
+      <Skeleton m={2} width={400} height={300} />
+      <Skeleton m={2} width={400} height={300} done />
```
damassi, line 9: should `done` be the primary axis of behavior or would `loading` be more appropriate?

dzucconi: The way I was thinking was that the skeleton is always loading by default because that's reason for existence and eventually receives a done which can reflect ready/error (either way stops the animation) — so I guess trying to avoid a default as false state.

damassi: Cool, makes sense 👍

dzucconi:
So I think what I'll have to do is mark this as a breaking change and add the old "ResponsiveImage" as `BackgroundedImage` (naming?)? I could add that in Force rather than here if we figure that's not useful. It's used in a few of places as a square background image crop in Force. @damassi any thoughts?

damassi:
How about `ResponsiveImageDeprecated`? 

The main spots it's used is on artist/shows and on the /artwork/id page.

dzucconi:
So one of the conundrums I'm dealing with with this is that the these components behave *very* differently depending on the options, which is partially also why I found myself unable to even use it in a new app.

* By default it doesn't lazyload and is the codepath for that outputs an `` tag, meaning it is `display: inline`, by default
* Pass it `lazyLoad` it has an outer wrapper that is `display: block`, by default
* Use `<ResponsiveImage />` without `lazyLoad` and you get a div with a square aspect ratio
* Pass that `lazyLoad` and you actually don't even get a `<ResponsiveImage />` anymore, you just get a `<LazyImage />` with a kind of broken wrapper

---

So we fix the `<ResponsiveImage />` issues by just ignoring them completely: re-create the aspect ratio background image thing inside of force, deprecate it and then replace all the old usage.

The more subtle problem is what to do about the alternate codepath for non-lazy load image giving it a `display: inline` by default. Right now in this PR, images are always in the container, lazy or not. And so they always have a `<Skeleton />` behind them, which is fine, it just means that if there was any usage that happened to rely on `display: inline` — which is probably unlikely (?) it'd break.

----

So the two options here are, default that skeleton to `display: inline-block` which is a semi-reasonable default for images to have and for instance if you rendered out an `<Image />` without specifying its dimensions, it'd render at the native img dimensions.

Alternatively leave the skeleton at `display: block`. If you rendered out an `<Image />` without specifying its dimensions it would scale to width: 100%.

dzucconi:
I think I'd like to approach this in a more incremental fashion. There's a lot of smaller pieces to this that can be implemented and exposed rather than just tearing the entire thing down. Let me make a series of smaller PRs to try to get a grip on it.

PR #731, opened 2020-07-15
GALL-2970 Add web version of medium card

ansor4:
Adds a medium size card to palette. See spec for medium card here: https://www.figma.com/file/GrCeNUJbKpfZtgwHga64cC/Artsy-Viewing-Room-Entry-Points-Components?node-id=78%3A1368

2020-07-15 at 3 26 35 PM

### DIFF
--- a/packages/palette/src/elements/Cards/MediumCard.tsx
+++ b/packages/palette/src/elements/Cards/MediumCard.tsx
@@ -1,10 +1,49 @@
 import React from "react"
+import { Box } from "../Box"
+import { Flex } from "../Flex"
+import { Image } from "../Image"
+import { Sans } from "../Typography"
+import { CardTag } from "./CardTag"
 import { MediumCardProps } from "./MediumCard.shared"
 
 /**
  * `MediumCard` is a card with one image one tall image, and text for title and subtitle
  * at the bottom.
  */
-export const MediumCard: React.FC<MediumCardProps> = ({}) => {
-  return null
+export const MediumCard: React.FC<MediumCardProps> = ({
+  image,
+  title,
+  subtitle,
+  tag
+}) => {
+  return (
+    <Box width="280px" height="370px" position="relative">
+      <Flex
+        width="100%"
+        height="100%"
+        position="absolute"
+        top="0"
+        borderRadius="4px"
+        overflow="hidden"
+      >
+        <Image src={image} alt={title} height="100%" width="auto" />
+        <Box
+          style={{
+            background: "linear-gradient(rgba(0, 0, 0, 0), rgba(0,0,0,1))",
+            opacity: 0.15
+          }}
+          position="absolute" width="100%" height="100%"
+        />
+        <Box position="absolute" bottom="15px" left="15px">
+          <Sans size="5t" weight="regular" color="white100">
+            {title}
+          </Sans>
+          <Sans size="3t" weight="regular" color="white100">
+            {subtitle}
+          </Sans>
+        </Box>
+        {!!tag && <CardTag {...tag} position="absolute" top="15px" left="15px" />}
+      </Flex>
+    </Box >
+  )
 }
--- a/packages/palette-docs/content/docs/elements/layout/Cards.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/Cards.mdx
@@ -64,7 +64,7 @@ name: Cards
 <Playground title="Medium Card with tag">
   <Box width="50%" p="2px" backgroundColor="yellow30">
     <MediumCard
-      image="https://www.fillmurray.com/400/300"
+      image="https://www.fillmurray.com/280/370"
       title="Check this out"
       subtitle="It's good"
       tag={{


packages/palette/src/elements/Cards/MediumCard.tsx
```diff
@@ -1,10 +1,50 @@
...
+        top="0"
+        borderRadius="4px"
+        overflow="hidden"
+      >
+        <Image src={image} alt={title} height="100%" width="auto" />
+        <Box
+          style={{
+            background: "linear-gradient(rgba(0, 0, 0, 0), rgba(0,0,0,1))",
+            opacity: 0.15
+          }}
```
dzucconi, line 35: We do this gradient overlay thing in quite a few places, seems like we might want to abstract that?

packages/palette/src/elements/Cards/MediumCard.tsx
```diff
@@ -1,10 +1,50 @@
...
-export const MediumCard: React.FC<MediumCardProps> = ({}) => {
-  return null
+export const MediumCard: React.FC<MediumCardProps> = ({
+  image,
+  title,
+  subtitle,
+  tag
+}) => {
+  return (
+    <Box width="280px" height="370px" position="relative">
```
dzucconi, line 21: I know the specs say this is a fixed size but I don't see how that'd work in practice (vs say a scalable aspect ratio box) — why the fixed px size?

packages/palette/src/elements/Cards/MediumCard.tsx
```diff
@@ -1,10 +1,50 @@
...
+      >
+        <Image src={image} alt={title} height="100%" width="auto" />
+        <Box
+          style={{
+            background: "linear-gradient(rgba(0, 0, 0, 0), rgba(0,0,0,1))",
+            opacity: 0.15
+          }}
+          position="absolute" width="100%" height="100%"
+        />
+        <Box position="absolute" bottom="15px" left="15px">
```
dzucconi, line 38: Seems odd to me to step outside the spacing scale — would ask design why?

packages/palette/src/elements/Cards/MediumCard.tsx
```diff
@@ -1,10 +1,50 @@
...
+    <Box width="280px" height="370px" position="relative">
+      <Flex
+        width="100%"
+        height="100%"
+        position="absolute"
+        top="0"
+        borderRadius="4px"
+        overflow="hidden"
+      >
+        <Image src={image} alt={title} height="100%" width="auto" />
```
dzucconi, line 30: It's not immediately apparent to me what this would do with a non perfectly sized image? Do we just expect we're always going to pass in a cropped URL?

oxaudo: Totally correct @dzucconi - it's currently an issue that we have on a Small card as well. We need to either use `ResponsiveImage` I guess or pass width/height for figure out if width or height should be 100% to cover the space.

The issue is we are in the process of moving from simple URL on the VR to storing and Image model - so it's not exactly clear to me which approach will work better - maybe we can just perfectly size the image with `resize` on graphQl layer.

dzucconi: `object-fit: cover` would probably work well here.

packages/palette/src/elements/Cards/MediumCard.tsx
oxaudo: @dzucconi - I believe the idea here is that this card should be used in the top carousel of featured works on this page: https://www.figma.com/file/8k9EDGkQKclWjFwoyP8tuJ/Artsy-Viewing-Room-Landing-Page-%5BWeb%5D?node-id=82%3A0 and this carousel is fixed height.

packages/palette/src/elements/Cards/MediumCard.tsx
dzucconi: But yeah a cropped URL as well. I just assume with any lower level component: if it's possible to pass it non-ideal data, then it'll eventually happen. 🤷

packages/palette/src/elements/Cards/MediumCard.tsx
ansor4: I believe the small card has 10px and the large card has 20px, and so medium falls in the middle which is outside the spacing scale, so it kinda makes sense.

damassi: We shouldn't ever really step outside of the spacing scale as one-offs like this add up. Rounding down or up would probably be fine?

ds300: we rounded down to 10px on the native version, for what it's worth https://github.com/artsy/palette/pull/717/files/e44bd4c4bfee5c2a0720fc5c54c8934b9831d8f9#diff-8af20cc56832db8da1d1fc1c57c71860

ansor4: Conferred with @tinn on this, seems like we're already using 15px in other places, it just isn't in palette yet. Not sure how to resolve or what that means for this PR though, when design is saying 10px is too tight for this card. Would setting 15px be ok for now and we go back and clean it up when that spacing makes its way into palette?

pvinis: We changed it to 15 for ios too. Tin said small cards get 10 (so 1), medium get 15, large get 20 (so 2).
https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Cards/MediumCard.ios.tsx#L62

packages/palette/src/elements/Cards/MediumCard.tsx
ansor4: That sounds good, but maybe out of scope for this PR? Happy to follow up this PR after VR launch

packages/palette/src/elements/Cards/MediumCard.tsx
ansor4: hmm, `object-fit: cover` definitely fixes the issue, but it's unclear to me whether I should pass that property in through `Image` or if we should switch to a `ResponsiveImage`. I'd advocate merging in this PR as is for now to unblock development and following up with the right solution when we decide on that.

packages/palette/src/elements/Cards/MediumCard.tsx
pvinis: Can one of you add this in an issue or front end practice or wherever the right place would be? 🙏

packages/palette/src/elements/Cards/MediumCard.tsx
pvinis: After talking with anson, I think we can merge and we can change it later if it's needed, but it seems fixed is what we want here.

artsyit:
:rocket: PR was released in [`v11.9.0`](https://github.com/artsy/palette/releases/tag/v11.9.0) :rocket:

Issue #733, opened 2020-07-20
Consider unifying Storybook and docs


dzucconi:
The experience of trying to update the palette-docs is frustrating. No types, every change seems to trigger errors on compilation, changes randomly pop up the "Gatsby.js development 404 page" etc. All its dependencies are out of date and it's difficult to manage installs in palette from the root because installation of docs breaks. Most of the information surrounding process is no longer relevant, etc, etc.

One option we have here is to just upgrade to the new Story format and either write a custom wrapper that utilizes the exported stories, or just use Storybook + mdx docs to improve documentation from within storybook.

damassi:
> The experience of trying to update the palette-docs is frustrating. No types, every change seems to trigger errors on compilation, changes randomly pop up the "Gatsby.js development 404 page"

Maybe we should pair? The team has been working in these docs since 2019 with none of the problems that you've mentioned here. It's quite possible you're boothing things incorrectly, or misunderstand.

damassi:
Pulling palette repo down, i noticed that common fsevents issue. Regenerated the lock file and `yarn install --check-files` and things installled as expected. `yarn start` then booted the docs as expected. Changes to palette are then picked up by gatsby.

PR #734, opened 2020-07-20
Removes serif from Text

dzucconi:
Also adds the `overflowEllipsis` option, and updates the typography docs. I would have added a bit more to those but [I can't get the docs to actually compile correctly](https://github.com/artsy/palette/issues/733).

### DIFF
--- a/packages/palette/src/elements/Text/Text.tsx
+++ b/packages/palette/src/elements/Text/Text.tsx
@@ -1,13 +1,18 @@
 import { themeGet } from "@styled-system/theme-get"
+import styled, { css } from "styled-components"
 import { variant } from "styled-system"
-import { styled as primitives } from "../../platform/primitives"
 import { textMixin, TextProps } from "./Text.shared"
 import { TEXT_VARIANTS } from "./tokens"
 
-/**
- * Text
- */
-export const Text = primitives.Text<TextProps>`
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/Text/Text.tsx
```diff
@@ -1,20 +1,27 @@
...
+`
+
+/** Text */
+export const Text = styled.div<TextProps & { overflowEllipsis?: boolean }>`
   ${variant({ variants: TEXT_VARIANTS.small })}
+  ${textMixin}
 
   @media (min-width: ${themeGet("breakpoints.0")}) {
     ${variant({ variants: TEXT_VARIANTS.large })}
+    ${textMixin}
```
dzucconi, line 21: Just PR-creeped this bug fix in. Due to how styled-components emits the media query CSS, we need to double include the textMixin in both contexts so when you specify say a `solid` line-height for Text (which is what I need in the `NavBar`, for instance — a valid use-case, the media query doesn't override it.

damassi:
LGTM 👍 

#mergeongreen

artsyit:
:rocket: PR was released in [`v11.10.0`](https://github.com/artsy/palette/releases/tag/v11.10.0) :rocket:

PR #735, opened 2020-07-20
Fixes Icon-related typings

dzucconi:
Not sure how these were breaking in master, but that lockfile regeneration apparently did it.

### DIFF
--- a/packages/palette/src/svgs/Icon.tsx
+++ b/packages/palette/src/svgs/Icon.tsx
@@ -1,43 +1,35 @@
-// @ts-ignore
 import React, { FC } from "react"
 import styled from "styled-components"
 import {
-  left,
-  LeftProps,
+  compose,
+  layout,
+  LayoutProps,
   position,
   PositionProps,
-  right,
-  RightProps,
   space,
   SpaceProps,
-  top,
-  TopProps,
 } from "styled-system"
 import { Color } from "../Theme"
 
-// : React.SVGProps<SVGSVGElement>
-
-// tslint:disable-next-line:no-empty-interface
 export interface IconProps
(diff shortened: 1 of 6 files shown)


packages/palette/src/svgs/Icon.tsx
```diff
@@ -1,42 +1,20 @@
...
-} from "styled-system"
+import { position, PositionProps, space, SpaceProps } from "styled-system"
 import { Color } from "../Theme"
 
-// : React.SVGProps<SVGSVGElement>
-
-// tslint:disable-next-line:no-empty-interface
 export interface IconProps
-  extends React.SVGProps<any>,
+  extends React.HTMLAttributes<SVGElement>,
```
dzucconi, line 7: This one is what we wanted. The rest is just styled-system clean up.

packages/palette/src/svgs/Icon.tsx
```diff
@@ -1,42 +1,20 @@
...
-    RightProps,
-    LeftProps {
+    PositionProps {
   fill?: Color
   title?: string
 }
 
 /** Wrapper for icons to include space */
-export const Icon = styled.svg.attrs<IconProps>({})`
+export const Icon = styled.svg<IconProps>`
```
dzucconi, line 15: I should bulk fix all these `attrs<Type>({})` things

damassi:
LGTM 👍

packages/palette/src/svgs/Icon.tsx
dzucconi: (`React.SVGAttributes<SVGSVGElement>` rather... but the issue lies elsewhere)

PR #736, opened 2020-07-20
Text extends Box

dzucconi:
Am updating NavBar to use new Text components and noticed that we supported space, etc on text. Which is great and we should. I originally didn't add any of those to the new Text components to keep it consistent with iOS, but this is no longer an issue.

Includes some related clean up from today and reverting the lock file change so that master passes.


packages/palette/src/svgs/Icon.tsx
```diff
@@ -1,43 +1,31 @@
...
+export const Icon = styled.svg<IconProps>`
   position: relative;
-
-  ${space};
-  ${top};
-  ${right};
-  ${left};
-  ${position};
+  ${iconMixin}
 `
```
dzucconi, line 29: position now includes top/right/left

damassi:
👍

artsyit:
:rocket: PR was released in [`v11.10.0`](https://github.com/artsy/palette/releases/tag/v11.10.0) :rocket:

PR #737, opened 2020-07-21
Implements utility for breaking apart layout props

dzucconi:
OK, so I'd like to break apart https://github.com/artsy/palette/pull/730 into some more manageable pieces.

This gives us a typesafe utility that we can use when you have compound components where you've got an outer Box container that is being used to manage layout but would like to spread native rest props on the inner component (think img, input, etc).


zephraph:
👍

artsyit:
:rocket: PR was released in [`v11.11.0`](https://github.com/artsy/palette/releases/tag/v11.11.0) :rocket:

PR #738, opened 2020-07-21
Breakpoints aliasing

dzucconi:
@williardx has an interesting use case where he wants to override the Grid's `maxWidth` and set it to 100% when at or below the small breakpoint.

Initially I had thought you should be able to use the responsive props for this but it turns out that's not useful because in terms of the directionality of the media query, to do this you'd want a max-width, and responsive props create styles using min-width.

Nonetheless, I thought I'd PR support for object style syntax here. Which at least makes this a lot more clear and less mysterious.

### DIFF
--- a/packages/palette/src/Theme.tsx
+++ b/packages/palette/src/Theme.tsx
@@ -49,6 +49,15 @@ export const unitlessBreakpoints = {
 }
 
 /**
+ * We alias breakpoints onto the scale so that styled-system has access
+ * to the named breakpoints as well as the scale
+ */
+const BREAKPOINTS_SCALE = Object.assign(
+  [breakpoints.sm, breakpoints.md, breakpoints.lg, breakpoints.xl],
+  breakpoints
+)
+
+/**
  * All of the config for the Artsy theming system, based on the
  * design system from our design team:
  * https://www.notion.so/artsy/Master-Library-810612339f474d0997fe359af4285c56
@@ -60,7 +69,7 @@ export const themeProps = {
   /**
    *  This allows styled-system to hook into our breakpoints
    */
-  breakpoints: [breakpoints.sm, breakpoints.md, breakpoints.lg, breakpoints.xl],
+  breakpoints: BREAKPOINTS_SCALE,
 
   /**
    * Artsy's color schemes:
--- a/packages/palette/src/elements/Grid/Grid.story.tsx
+++ b/packages/palette/src/elements/Grid/Grid.story.tsx
@@ -1,0 +1,29 @@
+import { storiesOf } from "@storybook/react"
+import React from "react"
+import { BorderBox } from "../BorderBox"
+import { Col, Grid, Row } from "./Grid"
+
+storiesOf("Components/Grid", module).add("Basic", () => {
+  return (
+    <Grid
+      border={{
+        default: "4px solid pink",
+        xs: "4px solid red",
+        sm: "4px solid green",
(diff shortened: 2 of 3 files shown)


packages/palette/src/Theme.tsx
```diff
@@ -48,6 +48,25 @@ export const unitlessBreakpoints = {
...
+// @ts-ignore
+BREAKPOINTS_SCALE.xs = breakpoints.xs
+// @ts-ignore
+BREAKPOINTS_SCALE.sm = breakpoints.sm
+// @ts-ignore
+BREAKPOINTS_SCALE.md = breakpoints.md
+// @ts-ignore
+BREAKPOINTS_SCALE.lg = breakpoints.lg
+// @ts-ignore
+BREAKPOINTS_SCALE.xl = breakpoints.xl
```
dzucconi, line 68: styled-system requires that you stick these aliases directly on the scale object.

packages/palette/src/elements/Grid/Grid.story.tsx
```diff
@@ -0,0 +1,29 @@
+import { storiesOf } from "@storybook/react"
+import React from "react"
+import { BorderBox } from "../BorderBox"
+import { Col, Grid, Row } from "./Grid"
+
+storiesOf("Components/Grid", module).add("Basic", () => {
+  return (
+    <Grid
+      border={{
+        default: "4px solid pink",
```
dzucconi, line 10: So `default` can actually be anything and the behavior is a bit unintuitive. Basically if it doesn't find a key that's a valid alias it'll use that value as the default. And if you have multiple missing keys, whichever is last in the object wins.

packages/palette/src/Theme.tsx
damassi: It's a bit of a head trip to see an array tacking on object props -- can we use an intermediate variable so that `BREAKPOINTS_SCALE` can remain consistent type-wise at runtime?

packages/palette/src/elements/Grid/Grid.tsx
```diff
@@ -4,28 +4,55 @@ import {
...
+export const Col = styled(_Col)<
+  ColorProps & FlexProps & SpaceProps & TextAlignProps & WidthProps
+>`
+  ${compose(
+    color,
+    flex,
+    space,
+    textAlign,
+    width
+  )}
```
damassi, line 55: I like, just a _tiny_ bit tidier

packages/palette/src/Theme.tsx
dzucconi: This was a good callout. For whatever reason I didn't think this was possible but simple `Object.assign` here works perfectly

packages/palette/src/Theme.tsx
```diff
@@ -48,6 +48,15 @@ export const unitlessBreakpoints = {
...
 }
 
+/**
+ * We alias breakpoints onto the scale so that styled-system has access
+ * to the named breakpoints as well as the scale
+ */
+const BREAKPOINTS_SCALE = Object.assign(
+  [breakpoints.sm, breakpoints.md, breakpoints.lg, breakpoints.xl],
+  breakpoints
+)
```
damassi, line 58: Still so strange, but better 😄

artsyit:
:rocket: PR was released in [`v11.11.1`](https://github.com/artsy/palette/releases/tag/v11.11.1) :rocket:

PR #745, opened 2020-07-28
[Deps] Update packages

damassi:
This runs `yarn upgrade-interactive` over all of our deps, and downgrades `jest-styled-components` to fix a bunch of failed tests. 

@dzucconi - Next step would be to turn on dependabot / renovate auto-updates, unless there's other issues?


dzucconi:
🙏

packages/palette-docs/package.json
```diff
@@ -79,6 +79,7 @@
     "docz-utils": "^0.13.6",
     "gatsby": "2.3.34",
     "gatsby-image": "^2.0.20",
+    "gatsby-link": "^2.4.13",
```
damassi, line 82: See this [recent Gatsby issue](https://github.com/gatsbyjs/gatsby/issues/25708#issuecomment-658110313) /shrug

PR #747, opened 2020-07-30
Implements ResponsiveBox

dzucconi:
Continuing to pull pieces off the [`ResponsiveImage` PR](https://github.com/artsy/palette/pull/730) — the time has come that I need an aspect ratio constrained box in force 🔲


packages/palette/package.json
```diff
@@ -124,6 +124,7 @@
     "d3-interpolate": "^1.3.2",
     "d3-shape": "^1.3.5",
     "luxon": "^1.15",
+    "proportional-scale": "^3.0.1",
```
zephraph, line 127: Ha, nice. I gave it a ⭐

packages/palette/src/elements/ResponsiveBox/ResponsiveBox.tsx
```diff
@@ -0,0 +1,51 @@
...
+    height: aspectHeight,
+    maxWidth,
+    maxHeight,
+  })
+
+  return (
+    <Box
+      position="relative"
+      width="100%"
+      style={{ maxWidth: `${width}px` }}
```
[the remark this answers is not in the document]
dzucconi: Yeah, potentially you might see `100%`, in which case the `maxHeight` would be not necessary.

I think it might make sense to alter the type signature here to be either `maxWidth: number, maxHeight: number | maxWidth: '100%'` (and ignore any other potential values)

My main use-case here is for creating a bounding box for images — so generally you've got some upper bound to the image dimensions (the native dimensions at 1x (or have at 2x, etc)). But 100% is definitely a valid use as well.

packages/palette/package.json
dzucconi: This is one of those things I always felt like a fraud looking up on StackOverflow 😆

packages/palette/src/elements/Text/Text.ios.tsx
```diff
@@ -12,9 +13,9 @@ import {
 } from "./tokens.ios"
 
 /** TextProps */
-export type TextProps = BaseTextProps
+export type TextProps = BaseTextProps & NativeTextProps
 
-const InnerText = primitives.Text<TextProps>`
+const InnerText = styled(NativeText)<TextProps>`
```
dzucconi, line 18: Woops, somehow this snuck in.

artsyit:
:rocket: PR was released in [`v11.15.0`](https://github.com/artsy/palette/releases/tag/v11.15.0) :rocket:

PR #751, opened 2020-08-03
Reverts serif removal, migrates HTML to new typography

dzucconi:
So:

> Garamond will continue to be used in Editorial, and in all B2B web pages and outreach.

So I've reverted its removal here.

Also updating the HTML component to extend Text instead of Sans/Serif. This will constitute a breaking change, though the only call sites currently are on the Feature pages. I'll handle the integration of this as soon as it's released.

One other nice thing about this is now we get proper bold and italic fonts within rendered HTML 🥳

### DIFF
--- a/packages/palette/src/elements/HTML/HTML.story.tsx
+++ b/packages/palette/src/elements/HTML/HTML.story.tsx
@@ -11,7 +11,7 @@ const HTML_EXAMPLE = `
 
 <h4>H4 Headline</h4>
 
-<p>I’m of the opinion that they use no inert material. All their equipment and instruments are alive, in some form or other.</p>
+<p>I’m <em>of the opinion</em> that they use <strong>no <em>inert</em> material.</strong> All their equipment and instruments are alive, in some form or other.</p>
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/HTML/HTML.story.tsx
```diff
@@ -11,7 +11,7 @@ const HTML_EXAMPLE = `
 
 <h4>H4 Headline</h4>
 
-<p>I’m of the opinion that they use no inert material. All their equipment and instruments are alive, in some form or other.</p>
+<p>I’m <em>of the opinion</em> that they use <strong>no <em>inert</em> material.</strong> All their equipment and instruments are alive, in some form or other.</p>
```
damassi, line 14: ++

damassi:
👍 

#mergeOnGreen

artsyit:
:rocket: PR was released in [`v12.0.0`](https://github.com/artsy/palette/releases/tag/v12.0.0) :rocket:

PR #752, opened 2020-08-05
Adds remaining block-level elements that can come from rendered Markdown

dzucconi:
Minor thing, normalizing the remaining block-level elements that might come out of Markdown.

Not trying to be super comprehensive with how these are styled (for instance we're targeting `blockquote` specifically on Features) — but rather just give everything a normalized reasonable default.


artsyit:
:rocket: PR was released in [`v12.0.1`](https://github.com/artsy/palette/releases/tag/v12.0.1) :rocket:

PR #753, opened 2020-08-06
Add ArrowDownCircleFillIcon and ArrowUpCircleFillIcon

yuki24:
This PR two icons [we are going to use in the new My Bids screen](https://www.figma.com/file/QGuaH3b77El9tQwSMueEHo/Bidding-Management?node-id=1088%3A1521):

Screen Shot 2020-08-06 at 3 11 13 PM

Here is how they look like on the Palette doc:

AUCT-1172


packages/palette/src/svgs/ArrowDownCircleFillIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+  ...props
+}) => {
+  return (
+    <Icon {...props} viewBox="0 0 18 18">
+      <Title>{title}</Title>
+      <Path
+        d="M16 9.5C16 13.366 12.866 16.5 9 16.5C5.13401 16.5 2 13.366 2 9.5C2 5.63401 5.13401 2.5 9 2.5C12.866 2.5 16 5.63401 16 9.5Z"
+        fill={color(props.fill)}
+      />
+      <Path d="M9 12.7L6 7.5H12L9 12.7Z" fill="white" />
```
dzucconi, line 17: These fills should actually be holes. It'd be better if we could "flatten" the SVG to boolean out the shape from the other paths? (Consider if we were to support a dark mode)

yuki24: @willanderson Could you make an SVG file that has a hole, rather than a path that overrides the other `<path>`?

zephraph: If you pass the raw svg through [svgo](https://jakearchibald.github.io/svgomg/) it might do that for you.

yuki24: Sadly svgo couldn't merge paths properly (it still left the two paths that overrides the circle with a white background).

PR #755, opened 2020-08-13
Removes lodash

dzucconi:
Simple enough to remove lodash here, if we want to. Personally, I think the standard lib is good enough to not need it anymore anyway. Just remembered I had this branch from [this chat](https://artsy.slack.com/archives/CA8SANW3W/p1597260579293800)


packages/palette/package.json
```diff
@@ -123,6 +123,7 @@
     "babel-plugin-styled-components": "^1.10.0",
     "d3-interpolate": "^1.3.2",
     "d3-shape": "^1.3.5",
+    "debounce": "^1.2.0",
```
dzucconi, line 126: It looks like debounce is just hacking around a bug in the RadioButton but I haven't really looked at it. This package is tiny enough as a replacement for now.

packages/palette/src/elements/ProgressBarTimer/ProgressBarTimer.tsx
```diff
@@ -1,7 +1,7 @@
 import { DateTime } from "luxon"
 import React from "react"
-import { ProgressBar } from ".."
 import { color } from "../../helpers"
+import { ProgressBar } from "../ProgressBar"
```
dzucconi, line 4: No clue why removing lodash surfaced this breaking Storybook

damassi:
Really nice 👍

artsyit:
:rocket: PR was released in [`v12.0.2`](https://github.com/artsy/palette/releases/tag/v12.0.2) :rocket:

PR #759, opened 2020-08-24
Chore - Cleanup Dependency Cycles

icirellik:
There are several dependency cycles in Palette that under certain
circumstances such as loading a component using its canonical path
which will create partially loaded modules that lead to undefined objects at
runtime.

The solution is to import the files directly rather than relying on the global
index.


artsyit:
:rocket: PR was released in [`v12.2.0`](https://github.com/artsy/palette/releases/tag/v12.2.0) :rocket:

PR #760, opened 2020-08-24
[FX-2175] Carousel

dzucconi:
So the idea is to replace our dependence on Flickity with a few very specific carousel implementations. The vast majority of them are simple rails that have (or should have) a consistent behavior [as outlined in this spec](https://www.figma.com/file/I5sz0m16SiW7F3tsSoLRxK/Fair-Page?node-id=0%3A1).

What we'll do here is on mobile (touch, specifically) simply render an overflowing and horizontally scrollable rail, so we don't have to do any funny business trying to create a synthetic touch surface: this carousel is designed and meant to be operated with a pointing device.

Some affordances here are made for keyboard users. **Though this work isn't complete.** I'll follow up with some better focus management (TODO list below). Currently you can skip the rail entirely, skip to the end of the set of elements, and tabbing through from left-to-right will shift over the rail to make the current item visible. [The W3C guidelines for carousels can be found here](https://www.w3.org/WAI/tutorials/carousels/).

## a11y TODO
- [x] keyboard left-arrow/right-arrow navigates pages only when elements are in focus
- [ ] interaction between pressing next and being able to tab into the left-most visible item

----

I'm opening this PR to publish a canary and see how integration feels. I'll add on the touch version of this rail, but I'm not 100% sure how I want to handle that yet (since we don't have access to fresnel inside of palette (?).

<details>
  <summary>📦 Published PR as canary version: <code>12.2.1-canary.760.12660.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@12.2.1-canary.760.12660.0
  # or 
  yarn add @artsy/palette@12.2.1-canary.760.12660.0
  ```
</details>


packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+const Previous = styled(Arrow)`
+  left: 0;
+  transform: translateX(-100%);
+`
+
+const Cell = styled(Box)``
+
+export interface CarouselProps extends BoxProps {
+  children: JSX.Element | JSX.Element[]
```
dzucconi, line 83: There's quite a few options we *could* expose here, but we won't until they are actually needed.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const startRef = useRef<HTMLButtonElement | null>(null)
+  const endRef = useRef<HTMLButtonElement | null>(null)
+
+  const cells = useMemo(
+    () =>
+      Children.toArray(children)
+        .filter(isValidElement)
+        .map(child => ({ child, ref: createRef<HTMLDivElement>() })),
+    []
+  )
```
dzucconi, line 98: Wrap up children into a pageable array and create refs to each item which are threaded in when rendered. `useMemo` so that the refs are stable (`useRef` is frowned upon in `map` re: the rules of hooks).

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const { index, handleNext, handlePrev, setCursor } = useCursor({
+    max: pages.length,
+  })
+
+  const init = () => {
+    const { current: viewport } = viewportRef
+    const els = cells.map(({ ref }) => ref.current)
+    const values = els.map(node => node.clientWidth)
+    setCursor(0)
+    setPages(paginate(viewport.clientWidth, values))
```
dzucconi, line 111: All of the logic is a simple set of pure functions in the `utils` file. This paginate function takes the viewport width and the widths of every cell and returns back an array of page stops to transform with.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+          <ChevronIcon direction="left" width={30} height={30} />
+        </Previous>
+
+        <Next onClick={handleNext} disabled={index === pages.length - 1}>
+          <ChevronIcon direction="right" width={30} height={30} />
+        </Next>
+      </nav>
+
+      <Viewport ref={viewportRef as any}>
+        <Rail as="ul" style={{ transform: `translateX(-${pages[index]}px)` }}>
```
dzucconi, line 151: The offset which is animated ☝️

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+          {cells.map(({ child, ref }, i) => (
+            <Cell
+              as="li"
+              key={i}
+              ref={ref as any}
+              pr={i !== cells.length - 1 && 2}
```
dzucconi, line 157: Cell gap ☝️

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </Rail>
+      </Viewport>
+
+      <Skip ref={endRef} onClick={skipToStart} mt={1}>
+        Skip to beginning of content
+      </Skip>
+
+      <VisuallyHidden aria-live="polite" aria-atomic="true">
+        Page {index + 1} of {pages.length}
+      </VisuallyHidden>
```
dzucconi, line 171: This comes from the W3C guidelines: as you navigate, the aria-live content gets announced. Here, it just tells you what page you're on. The `polite` setting means that it waits for everything else to finish announcing before announcing the changed live content.

packages/palette/src/elements/Skip/Skip.tsx
```diff
@@ -0,0 +1,40 @@
...
+        bg="black10"
+        {...rest}
+      >
+        <Text p={1} variant="text">
+          {children}
+        </Text>
+      </Container>
+    )
+  }
+)
```
dzucconi, line 40: This exports our skip link that is also used in the NavBar.

packages/palette/src/elements/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export interface VisuallyHiddenProps extends BoxProps {
+  show?: boolean
+}
+
+/** Visually hide element, but present in DOM */
+export const VisuallyHidden = styled(Box)<VisuallyHiddenProps>`
+  ${visuallyHiddenMixin}
+  ${({ show }) => show && visuallyUnhiddenMixin}
+`
```
dzucconi, line 34: Just some helpers here for showing and hiding. Useful for the aria-live stuff, for instance.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  cursor: pointer;
+  user-select: none;
+  color: ${color("black30")};
+  transition: opacity 250ms, color 250ms;
+
+  > svg {
+    fill: currentColor;
```
dzucconi, line 55: Of the opinion all our SVGs should default to `currentColor`

> currentColor acts like a variable for the current value of the color property on the element. And the Cascading part of CSS is still effective, so if there’s no defined color property on an element, the cascade will determine the value of currentColor

Just let's you set color on the outer context without having to target the SVG fill explicitly.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+const Rail = styled(Box)`
+  display: flex;
+  height: 100%;
+  transition: transform 500ms;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+`
+
+const ARROW_WIDTH = space(4)
```
dzucconi, line 39: The reason this is a hardcoded value is because this is the available margin in our page wrapper. The arrows sit outside of the width of the carousel, which is a little bit weird but makes sense in our universe.

dzucconi:
Going to close/open this to see if I can trigger the canary.

packages/palette/src/elements/VisuallyHidden/VisuallyHidden.tsx
anandaroop: **question**: Is this particular cocktail of css attributes based on your past experience, or is it recommended as a best practice somewhere for this kind of show/hide-y stuff?  (I.e. using a clip path wouldn't have occurred to me 🙈 )

packages/palette/src/elements/Carousel/__tests__/utils.test.ts
```diff
@@ -0,0 +1,135 @@
+import { chunk, compound, paginate } from "../utils"
```
damassi, line 1: Is there possibly some less technical terms for these utilities?

packages/palette/src/elements/Carousel/Carousel.tsx
anandaroop: **suggestion (non-blocking)**: Is that animation duration governed by [L33](https://github.com/artsy/palette/pull/760/commits/8c1a7151dd2ab40213c85409b63612e344441039?file-filters%5B%5D=#diff-307ddca242ea3ceaabcc04d4d2c47b10R33) above? If so maybe pull that into a named constant?

packages/palette/src/elements/Carousel/__tests__/utils.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    expect(chunk(333, [100, 200, 300, 100, 200, 300])).toStrictEqual([
+      300,
+      300,
+      300,
+      300,
+    ])
+  })
+
+  it("chunks correctly (6)", () => {
+    expect(chunk(100, [100, 200, 300, 100, 200, 300])).toStrictEqual([
```
damassi, line 44: Can we use named arguments for these functions? At a glance, its not clear what 100 is, or what the array is for where if this accepted an object things would be a) no longer order dependent; and b) more readable friendly through time.

packages/palette/src/elements/Carousel/__tests__/utils.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      100,
+      200,
+      300,
+    ])
+  })
+})
+
+describe("paginate", () => {
+  it("returns a single page of 0 when the widths fit within the viewport", () => {
+    expect(paginate(1000, [100, 100, 100, 100, 100])).toStrictEqual([0])
```
damassi, line 57: Same here

packages/palette/src/elements/Carousel/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+export const compound = (xs: number[]): number[] => {
+  return xs.reduce((p, n) => p.concat((+p.slice(-1) || 0) + n), [])
+}
+
+/**
+ * Build an array of offsets
+ * @param viewport window width
+ * @param values array of cell widths
+ */
+export const chunk = (viewport: number, values: number[]) => {
```
damassi, line 14: I see - these are more lodash-like utils. Could we move them up to the top level utils folder?

packages/palette/src/elements/Carousel/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+/** Align last page to the left or right */
+export type Alignment = "left" | "right"
+
+/**
+ * Creates an array of offsets based on the given viewport
+ * @param viewport window width
+ * @param values array of cell widths
+ * @param alignment align last page to the left or right
+ */
+export const paginate = (
```
damassi, line 37: This makes sense to keep in the carousel folder, but split into its own file called `paginate`.

packages/palette/src/elements/Skip/Skip.tsx
```diff
@@ -0,0 +1,40 @@
...
+  ${visuallyHiddenMixin}
+
+  &:focus {
+    outline: 0;
+    ${visuallyUnhiddenMixin}
+  }
+`
+
+/** SkipProps */
+export type SkipProps = ClickableProps
```
damassi, line 19: It's unclear what `Skip` pertains to, and seems more specifically to the a subcomponent in the `Carousel` folder.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  Children,
+  createRef,
+  isValidElement,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react"
+import styled from "styled-components"
+import { useCursor } from "use-cursor"
```
anandaroop, line 11: **lulz**: Promise you won't unpublish this? 😛

packages/palette/src/elements/Skip/Skip.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/** Skip */
```
damassi, line 21: Real comments please 😄

packages/palette/src/elements/Skip/Skip.tsx
damassi: Ah, I see -- mind moving this github comment into an actual docblock comment?

packages/palette/src/elements/Carousel/Carousel.tsx
damassi: This would be better as a comment in code, or this snippet assigned to a self-describing variable.

packages/palette/src/elements/Carousel/Carousel.tsx
damassi: Comment in code

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,200 @@
...
+  height: 100%;
+  transition: transform 500ms;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+`
+
+const ARROW_WIDTH = space(4)
+
+const Arrow = styled(Clickable)`
```
damassi, line 41: There are instances of the carousel (ViewingRooms, for example) that take custom left and right arrows; mind matching that bit of the existing API as well?

packages/palette/src/elements/Carousel/__tests__/utils.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  })
+
+  it("handles the last page remainder (2)", () => {
+    expect(paginate(880, [333, 333, 333, 333, 333])).toStrictEqual([
+      0,
+      666,
+      785,
+    ])
+  })
+})
```
damassi, line 135: Can we add some additional UI specs that check that the expected components render, etc? These cover most of the lower-level functionality but we'll want to guard against another developer coming along and messing with say the layout.

packages/palette/src/elements/VisuallyHidden/VisuallyHidden.tsx
dzucconi: This is from [The A11Y Project](https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/) but you're right there should be a comment explaining that.

packages/palette/src/elements/Carousel/Carousel.tsx
dzucconi: 🙈

dzucconi: At the very least I can't unpublish it :) — the whole problem was using Github links directly in packages/Gemfiles.

packages/palette/src/elements/Carousel/Carousel.tsx
dzucconi: I'm sort of split about this. Right now there are only two carousels that fall outside of a the consistent pattern that spec refers to: Viewing rooms and the main image view of the artwork page.

The artwork page one is an easy call: that has a very different UX and would benefit from a snowflake pulling from a few primitives that we might expose in Palette.

The viewing rooms one is similar here but still different enough: it anchors to center, pages a single cell at a time (vs an entire page), has overlaid arrows, a progress bar, and the mobile implementation varies (has the progress bar and arrows still, for instance). I suspect that one is also a snowflake.

My thinking is to implement it as a snowflake in force, but use a composition of pieces from palette rather than exposing a bunch of props, the existence of which might (will) encourage their usage.

So I can:
* expose the `paginate` function and add support a `center` alignment
* implement progress dots and a progress bar for the artwork images and viewing rooms, respectively
* implement a snapping/progress aware scrollable rail

Those pieces would let us build out both of those snowflakes in force without having to support a complicated API that would possibly lead to design inconsistencies.

In the meantime just to get off Flickity in most of the cases, we'll move Force's `Carousel` => `LegacyCarousel` and utilize it in both the viewing rooms and artwork pages.

We'll then build an adapter component in force and just ignore the following props: `width`, `height`, `options`, `arrowPosition`, `showArrows`, `renderLeftArrow`, `renderRightArrow`. Or just update the call sites to the new API, there's only 14 excluding the ❄️ .

zephraph: I like that strategy. It would be nice generally to push for more consistency and ensuring we have as simple of an api surface area as we can.

packages/palette/src/elements/VisuallyHidden/VisuallyHidden.tsx
zephraph: Yeahh, adding the link in code would be good. Also I wonder if this module's name could be tweaked in a way to communicate it's utility. `HideAccessibly` doesn't roll off the tongue quite as well, but _something_ in that vein that helps describe the context. 

Last thing... that article is from 2013. Are we sure this is still the best way to achieve this?

dzucconi: 👍 

2020 :p https://css-tricks.com/inclusively-hidden/

`InclusivelyHidden` is a much more specific/better name too

dzucconi: Another good article from WebAIM: https://webaim.org/techniques/css/invisiblecontent/

dzucconi:
I think I've addressed all the feedback here. Thanks all! Will merge after lunch but if there's any additional feedback I will address.

artsyit:
:rocket: PR was released in [`v12.3.0`](https://github.com/artsy/palette/releases/tag/v12.3.0) :rocket:

packages/palette/src/elements/Carousel/Carousel.tsx
damassi: This makes sense! Thanks for the thorough comment 👍

PR #761, opened 2020-08-25
[FX-2175] Swiper, ProgressDots, custom navigation Carousel buttons

dzucconi:
This is the <del>mobile</del> touch side of the carousel coin. In most cases we're going to want a simple overflow scrolling rail with consistent cell gaps, which is simple enough.

I've added some additional functionality that also will let us support the outside cases: scroll snapping and tracking progress.

----

<del>I'll follow up here with a progress dots example and that'll be all the pieces to this puzzle.</del>

<details>
  <summary>📦 Published PR as canary version: <code>12.3.1-canary.761.12791.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@12.3.1-canary.761.12791.0
  # or 
  yarn add @artsy/palette@12.3.1-canary.761.12791.0
  ```
</details>


packages/palette/src/elements/Swiper/__tests__/Swiper.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <Box>2</Box>
+        <Box>3</Box>
+      </Swiper>
+    )
+
+    // Renders all cells
+    expect(wrapper.find(Box).length).toBe(3)
+    expect(wrapper.html()).toContain("scroll-snap-align: center")
+  })
+})
```
damassi, line 20: 👍

packages/palette/src/elements/Swiper/__tests__/progress.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it("calculates the amount of progress made (1)", () => {
+    expect(percentage({ viewport: 1000, total: 10000, left: 0 })).toBe(0)
+  })
+
+  it("calculates the amount of progress made (2)", () => {
+    expect(
+      percentage({
+        viewport: 1000,
+        total: 10000,
+        left: 10000 - 1000,
```
damassi, line 13: ty for the named args

packages/palette/src/elements/Swiper/progress.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Calculate the percentage of the rail that has been scrolled.
+ * If the left edge is at the total minus the viewport we are done scrolling.
+ * @param viewport window width
+ * @param total total width of the rail
+ * @param left left edge of scroll
+ */
+export const percentage = ({
```
damassi, line 8: Maybe rename the file `percentage` to match the main export?

packages/palette/src/elements/Swiper/progress.ts
```diff
@@ -0,0 +1,33 @@
...
+}) => {
+  return Math.floor((left / (total - viewport)) * 100)
+}
+
+/**
+ * Convert the progress into the active index, given an array of `length`.
+ * @param progress number from 1 to 100
+ * @param length length of an array
+ */
+export const index = ({
```
damassi, line 25: (nonblocking) Maybe `index` could be named something a bit more descriptive? Its easy to see `index` and get it confused with old web baggage / conventions.

packages/palette/src/elements/Swiper/Swiper.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const CELL_GAP_PADDING_AMOUNT: SpacingUnit[] = [1, 2]
+
+const Container = styled(Box)`
+  position: relative;
+`
+
+const visuallyDisableScrollbar = css`
+  /* IE 10+ */
+  -ms-overflow-style: none;
```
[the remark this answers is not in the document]
dzucconi: As far as I know, no — probably because it's a bit of a hack to get this working, targeting various different vendor specific properties like this. [Here's an amusingly long Stack Overflow thread](https://stackoverflow.com/questions/16670931/hide-scroll-bar-but-while-still-being-able-to-scroll) filled with various solutions.

I like the idea of just styling the scrollbar to be transparent, which feels a bit more future proof than this, but off-hand I don't know how well supported that is.

packages/palette/src/elements/Swiper/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./activeIndex"
+export * from "./percentage"
```
zephraph, line 2: Should these be exported from the root of Swiper here?

These are all re-exported in `elements/index` meaning you could do `import { percentage } from '@artsy/palette'` which feels like it isn't the right api.

I could see maybe something like 

```
import { Swiper } from '@artsy/palette'
Swiper.activeIndex
```
That would make a little more sense as it's constrained to the element it's used within instead of polluting the global package scope.

You don't have to follow that particular solution, but I _would_ like to see these not exposed at this level. Of course, open to a counter if I'm missing something.

packages/palette/src/elements/ProgressDots/ProgressDots.story.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const interval = setInterval(handleNext, 1000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [])
+
+  return (
+    <>
+      <ProgressDots activeIndex={index} amount={5} my={2} />
+    </>
```
zephraph, line 19: I don't feel strongly about this so it's not required. Feel free to resolve if you'd like to keep it as is.

```suggestion
    <ProgressDots activeIndex={index} amount={5} my={2} />
```

packages/palette/src/elements/Swiper/index.ts
dzucconi: Yeah, I concur. These aren't needed yet so I can cross that bridge when I get to it.

packages/palette/src/elements/ProgressDots/ProgressDots.tsx
```diff
@@ -0,0 +1,58 @@
...
+  transition: background-color ${DOT_COLOR_MS}ms;
+
+  ${({ active }) =>
+    active &&
+    css`
+      background-color: ${color("black100")};
+    `}
+`
+
+Dot.displayName = "Dot"
```
zephraph, line 23: ❤️ 

We should check to make sure we're doing this everywhere else... really does help debugging.

packages/palette/src/elements/Swiper/activeIndex.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+export const activeIndex = ({
+  progress,
+  length,
+}: {
+  progress: number
+  length: number
+}) => {
+  return Math.floor((progress * (length - 1)) / 100)
+}
```
zephraph, line 14: Should this be it's own file? I don't feel strongly about it either way, but I'd like to hear your perspective.

packages/palette/src/elements/ProgressDots/ProgressDots.story.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Demo = () => {
+  const { index, handleNext } = useCursor({ max: 5 })
+
+  useEffect(() => {
+    const interval = setInterval(handleNext, 1000)
+    return () => {
+      clearInterval(interval)
+    }
+  }, [])
```
zephraph, line 14: Shouldn't `handleNext` be a dependency here?

packages/palette/src/elements/Swiper/activeIndex.ts
dzucconi: Honestly, I just wanted a `utils.ts` or `helpers.ts` file in both Carousel and Swiper. 😄

packages/palette/src/elements/ProgressDots/ProgressDots.story.tsx
dzucconi: We really need to switch off TSLint in this project and add in the hooks linter.

packages/palette/src/elements/ProgressDots/ProgressDots.tsx
dzucconi: Isn't there a babel plugin for this?

damassi: https://github.com/artsy/palette/blob/master/packages/palette/.babelrc#L16 should do it

dzucconi: I guess it's not being applied within Jest — the reason I added it here is so I could target it in Jest without exporting.

artsyit:
:rocket: PR was released in [`v12.4.0`](https://github.com/artsy/palette/releases/tag/v12.4.0) :rocket:

PR #763, opened 2020-08-31
Pass through done prop on SkeletonText

dzucconi:
Prop was missing — small fix to pass it through.


artsyit:
:rocket: PR was released in [`v12.4.2`](https://github.com/artsy/palette/releases/tag/v12.4.2) :rocket:

PR #764, opened 2020-08-31
Adds support for `srcSet`

dzucconi:
So I had initially intended to refactor this entire component but that got away with me. My intention with simply adding the prop here is that currently because we swap `src` props for `FillwidthItem` on page load, we actually double load images on all 1x devices 😈


artsyit:
:rocket: PR was released in [`v12.4.3`](https://github.com/artsy/palette/releases/tag/v12.4.3) :rocket:

PR #765, opened 2020-09-01
Move paginateCarousel Rename to Original File

icirellik:
Moves the renaming of the paginateCarousel function into the file where
it is create rather then renaming it in the `index.ts`.


artsyit:
:rocket: PR was released in [`v12.5.0`](https://github.com/artsy/palette/releases/tag/v12.5.0) :rocket:

PR #776, opened 2020-09-14
Cleans up Cards; support all image attributes

dzucconi:
Regarding: https://github.com/artsy/force/pull/6264 — would like to support retina images here.

This also cleans up a bit; removes the `LargeCard` which wasn't a real web component; and updates the typography to the new variants.


artsyit:
:rocket: PR was released in [`v13.3.0`](https://github.com/artsy/palette/releases/tag/v13.3.0) :rocket:

PR #777, opened 2020-09-14
Add 1.5 and 5 to our spacing units

zephraph:
Catch up with the discussion here: https://artsy.slack.com/archives/C9XJKPY9W/p1600109997018200. 

TL;DR some new designs were starting to use 15px more which aren't represented by our spacing system. Generally we _don't_ want to have 1 offs, especially for margins and paddings. Design has been using 15px more for the app and xs experiences, so we've decided to add it to account for that. 50px has similarly been used multiple times and was suggested to be added by @nicoleyeo.

### DIFF
--- a/packages/palette/src/Theme.tsx
+++ b/packages/palette/src/Theme.tsx
@@ -160,6 +160,8 @@ export const themeProps = {
     0.5: "5px",
     /** Equivalent to 10px  */
     1: "10px",
+    /** Equivalent to 15px  */
+    1.5: "15px",
     /** Equivalent to 20px  */
     2: "20px",
     /** Equivalent to 30px  */
@@ -167,6 +169,8 @@ export const themeProps = {
     /** Equivalent to 40px  */
     4: "40px",
     /** Equivalent to 50px  */
+    5: "50px",
+    /** Equivalent to 60px  */
     6: "60px",
     /** Equivalent to 90px  */
     9: "90px",
--- a/packages/palette/src/elements/Cards/MediumCard.tsx
+++ b/packages/palette/src/elements/Cards/MediumCard.tsx
@@ -43,7 +43,7 @@ export const MediumCard: React.FC<MediumCardProps> = ({
           width="100%"
           height="100%"
         />
-        <Box position="absolute" bottom="15px" left="15px">
+        <Box position="absolute" bottom={1.5} left={1.5}>
           <Sans size="5t" weight="regular" color="white100">
             {title}
           </Sans>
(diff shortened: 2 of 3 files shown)


packages/palette/src/Theme.tsx
```diff
@@ -160,13 +160,17 @@ export const themeProps = {
...
+    /** Equivalent to 15px  */
+    1.5: "15px",
     /** Equivalent to 20px  */
     2: "20px",
     /** Equivalent to 30px  */
     3: "30px",
     /** Equivalent to 40px  */
     4: "40px",
     /** Equivalent to 50px  */
+    5: "10px",
```
dzucconi, line 172: `5: "50px"`

zephraph: haaa. yes.

packages/palette/src/elements/Cards/MediumCard.tsx
```diff
@@ -43,17 +43,15 @@ export const MediumCard: React.FC<MediumCardProps> = ({
...
             {title}
           </Sans>
           <Sans size="3t" weight="regular" color="white100">
             {subtitle}
           </Sans>
         </Box>
-        {!!tag && (
-          <CardTag {...tag} position="absolute" top="15px" left="15px" />
-        )}
+        {!!tag && <CardTag {...tag} position="absolute" top={1.5} left={1.5} />}
```
dzucconi, line 54: Ah I just PR'd some changes here. I can deal with the conflicts.

packages/palette/src/Theme.tsx
dzucconi: 😛

packages/palette/src/elements/Cards/MediumCard.tsx
zephraph: Sounds good. Relatively minor in either case.

packages/palette/src/Theme.tsx
```diff
@@ -160,13 +160,17 @@ export const themeProps = {
...
+    /** Equivalent to 15px  */
+    1.5: "15px",
     /** Equivalent to 20px  */
     2: "20px",
     /** Equivalent to 30px  */
     3: "30px",
     /** Equivalent to 40px  */
     4: "40px",
     /** Equivalent to 50px  */
+    5: "50px",
```
damassi, line 172: 🥳

artsyit:
:rocket: PR was released in [`v13.2.0`](https://github.com/artsy/palette/releases/tag/v13.2.0) :rocket:

PR #779, opened 2020-09-15
[FX-2245] Read more styling

dzucconi:
Re: [FX-2245](https://artsyproduct.atlassian.net/browse/FX-2245)

In order to handle this — the simplest way to go about it is to style  the content from the outside. This PR exposes `HTMLStyles` which is the container from the `HTML` component in order to handle this consistently. Some related updates have also been made.

We can't style from inside the component because the read more link leans on `display: inline` which will cause changes to the line-height. This is also a bit more general as well, avoiding making assumptions about the HTML we are truncating.

Some minor updates have also been made like supporting cursor/textDecoration props in `Clickable` and some light updates in the `ReadMore` component itself.

### DIFF
--- a/packages/palette/src/elements/Clickable/Clickable.tsx
+++ b/packages/palette/src/elements/Clickable/Clickable.tsx
@@ -1,12 +1,19 @@
 import React from "react"
 import styled from "styled-components"
+import { compose, ResponsiveValue, system } from "styled-system"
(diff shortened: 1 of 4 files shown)


dzucconi:
Mmm — lemme re-open this tomorrow actually, after investigating.

PR #780, opened 2020-09-16
Update to use new type styles across all palette components

zephraph:
There's still more to do here, but this should cover a pretty large swath.

There's a lot to check over too, so, yeah. I paired with @willanderson to make these changes, so any differences in typography sizes was made intentionally. Some things were skipped due to complexity and some open questions remain.

### DIFF
--- a/packages/palette/src/elements/Message/Message.tsx
+++ b/packages/palette/src/elements/Message/Message.tsx
@@ -3,7 +3,7 @@ import { color } from "../../helpers"
 import { styledWrapper } from "../../platform/primitives"
 import { SansSize } from "../../Theme"
 import { Flex, FlexProps } from "../Flex"
-import { Sans } from "../Typography"
+import { Text } from "../Text"
 
 /**
  * Spec: zpl.io/2Zg4Rdq
@@ -29,14 +29,15 @@ const StyledFlex = styledWrapper(Flex)`
  */
 export const Message: SFC<MessageProps> = ({
   children,
-  textSize = "3t",
+  textSize,
   ...others
 }) => {
   return (
     <StyledFlex p={2} {...others}>
-      <Sans size={textSize} color={color("black60")} weight="regular">
+      {/* TODO: Can the size be left here? */}
+      <Text size={textSize} color={color("black60")} variant="text">
         {children}
-      </Sans>
+      </Text>
     </StyledFlex>
   )
 }
--- a/packages/palette/src/elements/Message/Message.tsx
+++ b/packages/palette/src/elements/Message/Message.tsx
@@ -1,9 +1,8 @@
 import React, { SFC } from "react"
 import { color } from "../../helpers"
 import { styledWrapper } from "../../platform/primitives"
-import { SansSize } from "../../Theme"
 import { Flex, FlexProps } from "../Flex"
-import { Text } from "../Text"
+import { Text, TextVariant } from "../Text"
 
 /**
  * Spec: zpl.io/2Zg4Rdq
@@ -14,7 +13,7 @@ interface MessageProps extends FlexProps {
   /**
    * Size of text to display in message window
    */
-  textSize?: SansSize
+  variant?: TextVariant
 }
 
 const StyledFlex = styledWrapper(Flex)`
@@ -29,13 +28,12 @@ const StyledFlex = styledWrapper(Flex)`
  */
 export const Message: SFC<MessageProps> = ({
   children,
-  textSize,
+  variant = "text",
   ...others
 }) => {
   return (
     <StyledFlex p={2} {...others}>
-      {/* TODO: Can the size be left here? */}
-      <Text size={textSize} color={color("black60")} variant="text">
+      <Text color={color("black60")} variant="text">
         {children}
       </Text>
     </StyledFlex>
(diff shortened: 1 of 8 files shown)


dzucconi:
Was wondering if this was gonna have `Button` in it. Was just looking at that one a bit scared. 😄

zephraph:
LOL. Nah, too afraid.

In all serious though, it seemed hard to rectify without breaking changes.

packages/palette/src/elements/Message/Message.tsx
```diff
@@ -29,14 +29,15 @@ const StyledFlex = styledWrapper(Flex)`
...
   children,
-  textSize = "3t",
+  textSize,
   ...others
 }) => {
   return (
     <StyledFlex p={2} {...others}>
-      <Sans size={textSize} color={color("black60")} weight="regular">
+      {/* TODO: Can the size be left here? */}
+      <Text size={textSize} color={color("black60")} variant="text">
```
dzucconi, line 38: `size` is a valid prop from layout:



It's probably worth just denoting this a breaking change and removing `textSize` or accept `variant` or just ignore it completely?

packages/palette/src/helpers/injectGlobalStyles.tsx
```diff
@@ -36,9 +38,9 @@ export function injectGlobalStyles<P>(
     }
 
     html, body {
-      font-family: 'AGaramondPro-Regular';
-      font-size: 16px;
-      line-height: 24px;
+      font-family: ${themeProps.fontFamily.sans.regular as string};
+      font-size: ${themeProps.fontSizes.size3};
```
dzucconi, line 42: This should be `size5` => `16px`?

Changing this will change the value of `1rem`

packages/palette/src/helpers/injectGlobalStyles.tsx
```diff
@@ -36,9 +38,9 @@ export function injectGlobalStyles<P>(
...
+      line-height: ${TEXT_LINE_HEIGHTS.body};
```
zephraph, line 43: cc @dzucconi this is the scariest part IMO

packages/palette/src/helpers/injectGlobalStyles.tsx
```diff
@@ -36,9 +38,9 @@ export function injectGlobalStyles<P>(
     }
 
     html, body {
-      font-family: 'AGaramondPro-Regular';
-      font-size: 16px;
-      line-height: 24px;
+      font-family: ${themeProps.fontFamily.sans.regular as string};
```
dzucconi, line 41: This actually won't be correct I believe. The new tokens are under `fonts`.

packages/palette/src/helpers/injectGlobalStyles.tsx
dzucconi: We probably want to like alias these as ROOT_FONT_SIZE, ROOT_LINE_HEIGHT, etc, in that tokens file just so it's clear if any changes are made in that file.

zephraph: Ah, yeah, that's a good point... should add it to the themeProps instead though? That tokens file is a bit buried.

packages/palette/src/elements/Message/Message.tsx
zephraph: Ahh, crap, I see. Yeah, will need to figure out how to properly update this and find where it's used.

packages/palette/src/helpers/injectGlobalStyles.tsx
dzucconi: https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Text/tokens.ts is pretty clear still. My hope is that once we retire `Sans/Serif` we can clean up the theme a bit.

dzucconi: Oh, I see what you're saying — yah, makes sense (I only carved out the Text/tokens file because of the confusion with the older tokens.)

zephraph: Need to start adding some deprecation notices soon.

packages/palette/src/elements/Message/Message.tsx
zephraph: Resolved this but it's a breaking change.

packages/palette/src/helpers/injectGlobalStyles.tsx
```diff
@@ -36,9 +37,9 @@ export function injectGlobalStyles<P>(
     }
 
     html, body {
-      font-family: 'AGaramondPro-Regular';
-      font-size: 16px;
-      line-height: 24px;
+      font-family: ${themeProps.fonts.sans};
+      font-size: ${themeProps.fontSizes.size5};
+      line-height: ${themeProps.lineHeights.body};
```
anandaroop, line 42: Exciting!

anandaroop (approved):
Thanks for doing this!

packages/palette/src/elements/Message/Message.tsx
```diff
@@ -14,7 +13,7 @@ interface MessageProps extends FlexProps {
   /**
    * Size of text to display in message window
    */
-  textSize?: SansSize
+  variant?: TextVariant
```
zephraph, line 16: Breaking

dzucconi:
We should pick this back up — in particular I'm interested in dropping Garamond as the base font on React pages so that we don't have to preload Garamond on them anymore.

PR #781, opened 2020-09-18
Update gatsby plugins to be node v12 compatible

zephraph:
So I've been unable to successfully get `yarn install` to work in palette for a while. I went so far as to completely remove every node and python installation on my machine and start from scratch. Still, no dice.

It's always a build error w/ sharp. I started doing more digging and it's a common issue w/ gatsby and sharp for those upgrading to node 12. There's an [issue in sharp](https://github.com/lovell/sharp/issues/1683) that points to a [gatsby pr](https://github.com/gatsbyjs/gatsby/pull/13646) that helps describe what needs to be done. Essentially we just have to upgrade the sharp related gatsby plugins. That's what I did for this PR.


packages/palette/src/elements/Banner/Banner.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react"
 import styled from "styled-components"
-import { CloseIcon } from "../../svgs"
+import { CloseIcon } from "../../svgs/CloseIcon"
```
zephraph, line 3: This is a side issue that was causing the docs build to fail. I'm not certain exactly why (because the close icon is imported to the index), but pointing it to the exact file fixed the error. Probably a best practice to do that anyway since those index files don't allow for tree shaking.

dzucconi:
🙏

packages/palette/src/elements/Banner/Banner.tsx
damassi: yah, when we try to import from a bulk export list we get into circular dep problems

artsyit:
:rocket: PR was released in [`v13.4.1`](https://github.com/artsy/palette/releases/tag/v13.4.1) :rocket:

PR #783, opened 2020-09-21
Fixes horizontal margin overrides

dzucconi:
Some `Swiper` updates included `margin: 0` in the container which was overriding any passed in `mx` — here we set a default and set up a story to use as a visual spec.


artsyit:
:rocket: PR was released in [`v13.4.1`](https://github.com/artsy/palette/releases/tag/v13.4.1) :rocket:

PR #785, opened 2020-09-22
Sets default Text variant to `"text"`

dzucconi:
This sets the default `Text` variant as `text`, which is the obvious thing to do here. Thanks for raising @anandaroop.


anandaroop:
🙏

artsyit:
:rocket: PR was released in [`v13.6.0`](https://github.com/artsy/palette/releases/tag/v13.6.0) :rocket:

PR #788, opened 2020-09-23
Ensures consistent typography in EntityHeaders

dzucconi:
Re: [FX-2251](https://artsyproduct.atlassian.net/browse/FX-2251)

The use of caption was causing some line-height differences as well due to the way this component nests a (faux-)button (necessitating the use of stopPropagation).

-----

Added stories for use as visual specs and simplified the examples a bit.


artsyit:
:rocket: PR was released in [`v13.7.2`](https://github.com/artsy/palette/releases/tag/v13.7.2) :rocket:

PR #789, opened 2020-09-24
Only calls onChange when the active index actually changes

dzucconi:
`onChange` in `Swiper` was being called on scroll. This PR keeps track with an internal state and only calls `onChange` when the active index actually changes.


artsyit:
:rocket: PR was released in [`v13.7.3`](https://github.com/artsy/palette/releases/tag/v13.7.3) :rocket:

PR #791, opened 2020-10-01
[FX-2317] Fixes pagination arrow alignment

dzucconi:
The bug here surfaced when we added `5` to the spacing scale: `top` accepts themed values and since `space(0.5)` yields `5` this no longer falls back to `5px` but rather maps to the themed value of `50`.

I've added some stories to serve as visual specs since we'll have to edit this component a bit in the future.


artsyit:
:rocket: PR was released in [`v13.7.5`](https://github.com/artsy/palette/releases/tag/v13.7.5) :rocket:

PR #794, opened 2020-10-07
Implements GridColumns

dzucconi:
The idea here is to implement a new grid component built around CSS grid instead of Bootstrap's grid. The main motivation is to support (themed space scale) gutters.

This exposes two main components: `GridColumns` and `Column`

By default it collapses to a single column at the `xs` viewport width. If an array of responsive values is passed, those are respected.

-------

`GridColumns` is simply a display: grid with 12 evenly spaced columns and a default row and column gap of `space(2)`.

`Column` is a `Box` that has a simplified interface for building up a `grid-column` value. It accepts `span`, `start` and `wrap` in addition to normal `BoxProps`.

* `span` is a number between 1 and 12 of columns to span across
* `start` is a number between 1 and 12 of which column to start at
* `wrap` is a boolean which, if true, will force a break to the next row in the grid

-------

<details>
  <summary>📦 Published PR as canary version: <code>13.8.1-canary.794.13492.0</code></summary>
  <br />
  
  ```sh
  npm install @artsy/palette@13.8.1-canary.794.13492.0
  # or 
  yarn add @artsy/palette@13.8.1-canary.794.13492.0
  ```
</details>


packages/palette/src/elements/ColumnGrid/ColumnGrid.tsx
```diff
@@ -0,0 +1,123 @@
...
+        ? starts[i]
+        : starts[starts.length - 1]
+
+    if (x && x + n > 13) {
+      throw new Error("`span` and `start` must fit within the grid")
+    }
+
+    return x ? `${x} / span ${n}` : `span ${n}`
+  })
+}
```
dzucconi, line 84: This function looks complicated but all it's doing is merging `start` and `span` values into a responsive `gridColumn` value. All the codepaths have test coverage.

damassi (commented):
Maybe my lack of CSSGrid knowledge is hampering me, but `span` and `start` doesn't feel intuitive in terms of API, where the bootstrap grid does. 

`span` could be `columns` -- the number of columns, which it represents literally. (This also aligns with other component APIs we use througout force.)

`start` could be `columnStart` -- what column to start at. 

Then things are 1:1 with their actual function, and a user who only has a rough understanding of things would be able to follow. 

`breakAfter` could also be expanded to be a bit more literal, in this case matching the common flexbox idiom: `wrap` (as thats what its doing if i'm understanding correctly).

By simplifying and aligning language across various tools there's less conceptual information in the system.

damassi:
> It collapses to a single column at the xs viewport width. This is important to note as array values for span / start begin at the sm breakpoint, as a result.

This is pretty confusing to me and feels a bit too clever. Can we keep things more aligned to how users have used things in the past: 

```tsx
<Column columns={[1, 2, 4, 8, 16]}>`
```
Which spans all of our breakpoints without special rules

dzucconi:
> By simplifying and aligning language across various tools there's less conceptual information in the system.

Good notes and I agree with `start` => `columnStart` and `breakAfter` => `wrap`. Though I think span would be better named `columnSpan`.

> This is pretty confusing to me and feels a bit too clever

It's sort of the opposite of what you're imagining I think? The grid remains 12 columns throughout and so the first item of the array would have to be `12` so as to span across all 12 columns at the `xs` breakpoint. (We could change the number of columns at the grid level, though I think that's probably even more confusing).

I imagine that 99% of the time you're going to want a single column at `xs` — according to design, artwork columns are the only 2-column layout at mobile and those have their own special grid — and so doing this automatically will avoid *every single* `Column` instance from having to do `columnSpan={[12, 6]}` or what have you.

I think the vast majority of layouts will simply use single values for `columnSpan` and if you knew what you were doing and wanted to control that `xs` breakpoint you could just set a value for `gridColumn` (the actual CSS prop) directly.

dzucconi:
Also: `ColumnGrid` => `Columns`? 🤔

packages/palette/src/elements/ColumnGrid/ColumnGrid.story.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+        <Column span={6}>
+          <Text as="h2" variant="subtitle" mb={1}>
+            Page subtitle
+          </Text>
+
+          <Text>{IPSUM}</Text>
+        </Column>
+
+        <Column span={5} start={8}>
```
pepopowitz, line 45: For what it's worth I like the names `span` and `start`, this example reads very clearly to me about what it's doing.

packages/palette/src/elements/ColumnGrid/ColumnGrid.tsx
```diff
@@ -0,0 +1,123 @@
...
+const __Column__ = styled(Box)<__ColumnProps__>`
+  ${gridColumn}
+`
+
+/**
+ * Builds a value for `gridColumn` based on the assumptions laid out
+ * by our 12-column grid.
+ */
+export const calculateGridColumn = ({ span, start }: ColSpanProps) => {
+  const spans = (() => {
```
pepopowitz, line 40: non-blocking: I'm curious about the use of IIFE's here. Is there a need for them, or is it a stylistic choice to avoid creating a function elsewhere?

packages/palette/src/elements/ColumnGrid/ColumnGrid.tsx
```diff
@@ -0,0 +1,123 @@
...
+import styled from "styled-components"
+import { gridColumn, GridColumnProps } from "styled-system"
+import { Box, BoxProps } from "../Box"
+import { CSSGrid, CSSGridProps } from "../CSSGrid"
+
+/** ColumnGrid implements `Box` and the common grid properties */
+export type ColumnGridProps = CSSGridProps
+
+/**
+ * A 12-column fluid grid. Collapses to a single column at mobile breakpoints.
```
pepopowitz, line 11: In the event that I want to some day override `gridColumn` to break out of the single mobile column, it might be hard for me to figure out how to do that. I'm not sure if this is the right place to put it, but is there a docs comment we can put somewhere that would tell me that setting `gridColumn` explicitly will let me break out of the single mobile column?

pepopowitz (approved):
This is great! I think the API to the grid looks really nice. I'm going to assign to @damassi since he has requested changes.

damassi (approved):
> I imagine that 99% of the time you're going to want a single column at xs — according to design, artwork columns are the only 2-column layout at mobile and those have their own special grid — and so doing this automatically will avoid every single Column instance from having to do columnSpan={[12, 6]} or what have you.

This is true, but there are implicit rules to understand here which must be known, and which break some conventions. My only concern is that in existing grid code (which this component could then be used to refactor out) as well as other common 3rd party libraries its assumed that that the first item in the array is always `xs` -- _if it's defined_. If its omitted, it's presumed to be 12 columns, which is sort of a win win.

From a dev's perspective reading over the code later, seeing this:

```tsx
<Column columns={[12, 6]}> 
```

Makes it totally clear that this is a responsive layout, and that at `xs` its 12 columns and at `sm` its `6`. It might be slightly more verbose but i think there's value in that for responsiveness and for team-wide readability through time. 

Also, it matches expectations in all of our styled-system responsive props. So again patterns align app-wide and library-convention-wise. 

☝️ is just feedback, but isn't a hard blocker if you feel strongly about the current design. Though I do think that renaming the props would be beneficial.

damassi:
Regarding naming:

Why not 
```
ColumnGrid => Grid
```
If we're creating a successor we can use aliasing as things are being swapped in

dzucconi:
@damassi your comment makes me think that a more precise terminology here might be like:

```tsx
<Columns>
  <Cell columnSpan={6}>A cell spanning 6 columns</Cell>
</Columns>
```

Meaning the underlying grid is the thing that *has* the columns, and that the cells span across them.

🤔

dzucconi:
> if it's defined

Oh, yeah that does make a fair amount of sense to me.

dzucconi:
> If we're creating a successor we can use aliasing as things are being swapped in.

I think my main reason for the naming here is laziness on my part in making a breaking change and having to alias it at the all the call sites in force. Also it seems like we don't use the fluid grid in places which would make compatibility a bit of a problem.

damassi:
Maybe? Something of a compromise 😄 Also stands out bit more in terms of the contrasting order of words.

```
<GridColumns>
  <Column>
</GridColumns>
```

packages/palette/src/elements/ColumnGrid/ColumnGrid.tsx
dzucconi: Yeah, initially. Kind of the equivalent of doing this in Ruby:

```ruby
foo = begin
  ...
end
```

Though with the changes I just made, this got complex enough for it to actually be extracted.

packages/palette/src/elements/ColumnGrid/ColumnGrid.tsx
dzucconi: Taking @damassi's suggestion simplified this a lot actually, so now there really isn't a reason to use gridColumn directly.

Included an example in the kitchen sink story.

packages/palette/src/elements/GridColumns/GridColumns.story.tsx
```diff
@@ -0,0 +1,176 @@
...
+          columnSpan={[12, 6]}
+          columnStart={[1, 4]}
+          wrap
+        />
+
+        <Column bg="black100" columnStart={4}>
+          <Text color="white100">{IPSUM}</Text>
+        </Column>
+
+        <Column border="1px solid red" columnSpan={4} columnStart={1}>
```
damassi, line 127: Talked offline, but woke up thinking about this props name thing and, i think brevity is probably better, since `Column` is in the component name. Sorry for the back and forth!

dzucconi: I was also flipping back and forth depending on what ultimately this was named. like `Cell columnSpan` makes sense and `Column span` makes sense. Since we're going with `Column` for now, we'll go with that. Always can revisit if it proves confusing.

artsyit:
:rocket: PR was released in [`v13.9.0`](https://github.com/artsy/palette/releases/tag/v13.9.0) :rocket:

PR #799, opened 2020-10-16
[PCI-3] Supports props based navigation in Swiper/Carousel

dzucconi:
Closes: [PCI-3](https://artsyproduct.atlassian.net/browse/PCI-3)

Will follow up with a spec here, just drafting for now.

The use-case here is being able to reset the artwork image browser to the beginning/default image if you click "View in Room".

This is a bit more straight-forward in `<Carousel />` vs `<Swiper />` since you're updating the index as you scroll already.

I think this API makes sense (re: ***initial***Index) as it works as you'd expect by default and then if you want to gain full control over the state via props, you need to lift it up a level by passing a setter into `onChange`.

-----
### Swiper

-----

### DIFF
--- a/packages/palette/src/elements/Carousel/Carousel.story.tsx
+++ b/packages/palette/src/elements/Carousel/Carousel.story.tsx
@@ -57,6 +57,22 @@ const Dynamic = () => {
   return <Demo widths={widths} />
 }
 
+const Navigation = () => {
+  const [initialIndex, resetIndex] = useState(0)
+  return (
+    <Box>
+      <Demo initialIndex={initialIndex} onChange={resetIndex} />
+      <Box display="flex" justifyContent="space-around">
+        <Clickable onClick={() => resetIndex(0)}>Navigate to page 1</Clickable>
+        <Clickable onClick={() => resetIndex(1)}>Navigate to page 2</Clickable>
+        <Clickable onClick={() => resetIndex(2)}>Navigate to page 3</Clickable>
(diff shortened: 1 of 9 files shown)


packages/palette/src/elements/Carousel/Carousel.story.tsx
```diff
@@ -57,6 +57,22 @@ const Dynamic = () => {
   return <Demo widths={widths} />
 }
 
+const Navigation = () => {
+  const [initialCursor, resetCursor] = useState(0)
+  return (
+    <Box>
+      <Demo initialCursor={initialCursor} onChange={resetCursor} />
+      <Box display="flex" justifyContent="space-around">
+        <Clickable onClick={() => resetCursor(0)}>Navigate to page 1</Clickable>
```
damassi, line 66: (nonblocking) - "Cursor" seems like somewhat technical niche vocabulary vs something like `initialPage` and `setPage` or `goToPage`

artsyit:
:rocket: PR was released in [`v13.11.0`](https://github.com/artsy/palette/releases/tag/v13.11.0) :rocket:

packages/palette/src/elements/Swiper/Swiper.tsx
```diff
@@ -123,7 +154,9 @@ export const Swiper: React.FC<SwiperProps> = ({
         left: rail.scrollLeft,
       })
 
-      onChange && setIndex(activeIndex({ progress, length: cells.length }))
+      const page = activeIndex({ progress, length: cells.length })
```
damassi, line 157: Index is clearer, ty 👍

PR #803, opened 2020-10-23
fix: All icon to be filled

damassi:
Fix icon fill


artsyit:
:rocket: PR was released in [`v13.13.0`](https://github.com/artsy/palette/releases/tag/v13.13.0) :rocket:

PR #810, opened 2020-11-09
feat: support boolean errors for Input

mdole:
In Volt, we sometimes use an `<Input>` element from Palette without wanting to use the built-in text error rendering (we instead display custom formatted error message to match older components). This means we sometimes just pass a boolean error value to `<Input>`, which then renders the nice red border error state - but it also renders an empty `<Sans>` component.

I know this non-standard error handling is sorta an anti-pattern, _but_ I feel that it's helpful to support boolean errors since we can already do this in practice and it causes an unexpected render. 

Here's a gif that shows the issue with rendering an empty component (stretches out other adjacent components): https://share.getcloudapp.com/8LurzAWq

This PR adds official support for a boolean error being passed to the `<Input>`, which would cause it to render a red border but without rendering an empty text component.

### DIFF
--- a/packages/palette/src/elements/Input/Input.test.tsx
+++ b/packages/palette/src/elements/Input/Input.test.tsx
@@ -35,12 +35,22 @@ describe("Input", () => {
     expect(wrapper.text()).toEqual(props.description)
   })
 
-  it("returns an input with error when provided", () => {
-    const props = {
-      error: "This is the title",
-    }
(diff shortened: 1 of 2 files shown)


zephraph (approved):
Seems reasonable to me. 👍

artsyit:
:rocket: PR was released in [`v13.17.0`](https://github.com/artsy/palette/releases/tag/v13.17.0) :rocket:

PR #814, opened 2020-11-20
fix: [Text] update variant type to support array syntax

damassi:
For most stuff we can fall back to the automatic responsive sizes, but in some instances its nice to be able to use responsive array syntax when the differences are greater. (Versus `<Media>` component)


artsyit:
:rocket: PR was released in [`v13.20.0`](https://github.com/artsy/palette/releases/tag/v13.20.0) :rocket:

dzucconi:
Thanks!

PR #815, opened 2020-11-20
fix: Fix type def

damassi:
(Oops)


artsyit:
:rocket: PR was released in [`v13.21.0`](https://github.com/artsy/palette/releases/tag/v13.21.0) :rocket:

PR #817, opened 2020-11-22
Extract palette tokens into a separate package

zephraph:
It was decided in https://github.com/artsy/README/issues/318 to pull all react native specific components out of palette and into eigen. There was a note in the RFC that sharing tokens would be good, but that ended up not getting implemented. There was recent confusion about some specific spacing sizes missing in eigen that had been recently added to palette. That re-raised the need to share tokens. 

Check out the [slack thread](https://artsy.slack.com/archives/C02BAQ5K7/p1605638079020600) for more details. 

Essentially, this pulls out our design tokens and theme config into a separate package. It'll be followed up w/ a PR to migrate palette to begin consuming this. The last step will be to upgrade eigen to also consume it. 

I've switched lerna over to indpendent versioning mode so that the token package isn't updated every time we make a change to palette web. I also made an active effort to minimized/eliminate any dependencies in the tokens package so some things (like the `Theme` wrapper component) will need to be implemented upstream via palette web or palette native (or however we refer to it).

### DIFF
--- a/packages/palette-tokens/README.md
+++ b/packages/palette-tokens/README.md
@@ -2,18 +2,31 @@
 
 The [design tokens](https://www.lightningdesignsystem.com/design-tokens/) that power [Artsy](https://www.artsy.net)'s [Palette](https://palette.artsy.net).
 
-This package isn't intended to be consumed directly, but rather included as a transitive dependency of palette.
+:warning: This package isn't intended to be consumed directly, but rather included as a transitive dependency of palette. :warning:
 
 ## Usage
 
-In the main theme file...
+In the main theme file for a Palette implementation:
 
 ```
-import buildTheme from "@artsy/palette-tokens"
-export * from "@artsy/palette-tokens";
+import tokens from "@artsy/palette-tokens"
 import { fontFamily } from "./platform/fonts"
 
-export const themeConfig = buildTheme(fontFamily);
+/**
+ * This is required only in consuming versions of palette to keep
+ * the existing api the same. If you're using the tokens as a standalone
+ * you may not necessarily need to include this export statement.
+ */
+export * from "@artsy/palette-tokens";
+
+/**
+ * Any platform specific tokens can be included in the final `themeConfig` as
+ * shown.
+ */
+export const themeConfig = {
+  ...tokens,
+  fontFamily
+}
 ```
 
 ## License
--- a/packages/palette-tokens/README.md
+++ b/packages/palette-tokens/README.md
@@ -1,0 +1,46 @@
+# `palette-tokens`
+
+The [design tokens](https://www.lightningdesignsystem.com/design-tokens/) that power [Artsy](https://www.artsy.net)'s [Palette](https://palette.artsy.net).
+
+This package isn't intended to be consumed directly, but rather included as a transitive dependency of palette.
+
+## Usage
+
+In the main theme file...
+
+```
+import buildTheme from "@artsy/palette-tokens"
+export * from "@artsy/palette-tokens";
+import { fontFamily } from "./platform/fonts"
+
+export const themeConfig = buildTheme(fontFamily);
+```
+
+## License
+
+MIT License. See [LICENSE](../../LICENSE).
+
+## About Artsy
+
+<a href="https://www.artsy.net/">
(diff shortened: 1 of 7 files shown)


packages/palette-tokens/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Usage
+
+In the main theme file...
+
+```
+import buildTheme from "@artsy/palette-tokens"
+export * from "@artsy/palette-tokens";
+import { fontFamily } from "./platform/fonts"
+
+export const themeConfig = buildTheme(fontFamily);
```
damassi, line 16: The undocumented requirement (`import { fontFamily } from "./platform/fonts"`) is a bit confusing. Can we elaborate on what this is, and out of curiosity what would this look like in eigen?

packages/palette-tokens/README.md
```diff
@@ -0,0 +1,46 @@
...
+The [design tokens](https://www.lightningdesignsystem.com/design-tokens/) that power [Artsy](https://www.artsy.net)'s [Palette](https://palette.artsy.net).
+
+This package isn't intended to be consumed directly, but rather included as a transitive dependency of palette.
+
+## Usage
+
+In the main theme file...
+
+```
+import buildTheme from "@artsy/palette-tokens"
```
damassi, line 12: can we avoid the default export?

packages/palette-tokens/README.md
```diff
@@ -0,0 +1,46 @@
...
+export * from "@artsy/palette-tokens";
```
damassi, line 13: It might be nice if rather than needing to scoop this stuff up in a wildcard kind of way we spread it into an object and made things explicit:

```tsx
import { themeValues, buildTheme} from "@artsy/palette-tokens"

const themeConfig = buildTheme(fontFamily);
```

ds300 (approved):
Looks fine to me, thanks for taking the initiative to set this up! Excited to see the auto + lerna-indpendent-mode deploys in action!

MounirDhahri:
This is great 🎉

packages/palette-tokens/README.md
zephraph: The `fontFamily` object in [eigen](https://github.com/artsy/eigen/blob/master/src/palette/platform/fonts/fontFamily.ts) and [palette](https://github.com/artsy/palette/blob/master/packages/palette/src/platform/fonts/fontFamily.ts) vary in definition and shape. Instead of trying to account for the shape differences, I opted to let the platforms deal w/ it. 

That said, API your proposing is fine too 👍

packages/palette-tokens/README.md
zephraph: I'll speak to this below

packages/palette-tokens/README.md
zephraph: There are two things that this exports. The "tokens" which is essentially the themeConfig minus any platform specific things and the few global objects (unitlessBreakpoints, breakpoints) and a bunch of types specified at the bottom of the file. The intent is that the consuming `Theme.tsx` file either in the primary palette package or in eigen should have the same exports as it has today. 

While we want to avoid default exports generally, the default export in this package is important. There's no way to export all named exports _except_ a particular one. We essentially _don't_ want to import the incomplete tokens object in the consuming version of `Theme` because... well... it's incomplete. `themeConfig` is the true representation of that. So, this implementation does that essentially:

```
// only imports the default export
import buildTheme from "@artsy/palette-tokens"
// Exports all _named_ exports from the underlying package
export * from "@artsy/palette-tokens";
```

This is, afaik, the only way to accomplish this goal. Given that we'll only have two files that'll consume this and it _shouldn't_ change, I feel having a default export here is acceptable. 

This documentation is written from the perspective of someone integrating this package into a palette implementation. The export shape is important to maintain compatibility w/ existing palette implementations. I'll update the documentation to reflect this.

damassi (approved):
LGTM, but i'd like @dzucconi to chime in before merging 👍

dzucconi (approved):
LGTM — thanks for doing this 👍

artsyit:
:rocket: PR was released in [`@artsy/palette-tokens@1.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-tokens%401.0.0) :rocket:

PR #818, opened 2020-11-23
Refactor: Integrate tokens package to palette

zephraph:
This is a follow-up to #817. This adds the `@artsy/palette-tokens` package as a dependency of `@artsy/palette` and removes all the duplicative declarations. 

This is relatively close, but I don't have the bandwidth/mental space to get it through right now.

The largest remaining piece is that the tokens package is being transpiled to commonjs and the palette build is choking on that. So we may need to do something like not transpile the modules and add some es modules interop or somehow handle the dependency. I _think_ that's the biggest thing right now.


damassi (approved):

zephraph:
There's a few small followups I have to make before I can merge this. Will try to get to those today.

PR #821, opened 2020-12-03
feat: implements base tabs

dzucconi:
This PR implements a new `BaseTabs` component which adheres to the most recent Figma specs — it's essentially just a display component for a tab list and set of tabs (not the tab panels). Which allows us to `as={RouterLink}` them as needed for the `RouterTabs` in Force, which are more common lately. To that end it also overflows and gives a little fade-out affordance for use in the search page so that we can get rid of the `TabCarousel` and that instance of Flickity.

tabs

I'd like to utilize these `BaseTabs` to build out the client-side `Tabs` for consistency and a11y. Which should be straightforward enough. But before I do that I'd like to simplify the API a bit to drop functionality we're not using.

* `autoScroll` appears to be broken? I can try to glean the intention here and support this if desired. `OrderStepper` in Force looks to be the only component that is using this functionality.
  * Native `el.scrollIntoView` would be much simpler and should probably just be default on rather than a prop
* `transformTabBtn` appears to not be used anywhere, and I don't really understand why it would be used and what for.
* `Sup` appears to be not used and seems extraneous.

So I'd propose dropping those props and exports, cutting a major release and then updating the call-sites in Force and Volt.

I've also styled the existing tabs to update the typography, colors, and spacing to bring it inline with the spec.

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@13.24.3-canary.821.14150.0
  # or 
  yarn add @artsy/palette@13.24.3-canary.821.14150.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
+++ b/packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
@@ -1,0 +1,182 @@
+import { action } from "@storybook/addon-actions"
+import { storiesOf } from "@storybook/react"
+import React from "react"
+import { BaseTab, BaseTabs } from "."
+import { ChevronIcon } from "../../svgs"
+import { Box } from "../Box"
+import { Clickable } from "../Clickable"
+
+// Fake `RouterLink`
+const RouterLink: React.FC<{ to: string }> = ({ to, children, ...rest }) => {
+  return (
+    <a href={to} {...rest}>
+      {children}
+    </a>
+  )
+}
+
+storiesOf("Components/BaseTabs", module)
+  .add("Default", () => {
+    return (
+      <BaseTabs m={3}>
+        <BaseTab>First</BaseTab>
+        <BaseTab>Second</BaseTab>
+        <BaseTab active>Active</BaseTab>
+        <BaseTab>Last</BaseTab>
+      </BaseTabs>
+    )
+  })
+  .add('as="a"', () => {
+    return (
+      <BaseTabs m={3}>
+        <BaseTab as="a" href="#overview">
+          Overview
+        </BaseTab>
+
+        <BaseTab as="a" href="#cv">
+          CV
+        </BaseTab>
+
+        <BaseTab as="a" href="#shows" active>
+          Shows
+        </BaseTab>
+
+        <BaseTab as="a" href="#auction-results">
+          Auction Results
+        </BaseTab>
+
+        <BaseTab as="a" href="#articles">
+          Articles
+        </BaseTab>
+
+        <BaseTab as="a" href="#related-artists">
+          Related Artists
+        </BaseTab>
+      </BaseTabs>
+    )
+  })
+  .add("as={Clickable}", () => {
+    return (
+      <BaseTabs m={3}>
+        <BaseTab as={Clickable} onClick={action("onClick")}>
+          Overview
+        </BaseTab>
+
+        <BaseTab as={Clickable} onClick={action("onClick")}>
+          CV
+        </BaseTab>
+
+        <BaseTab as={Clickable} onClick={action("onClick")} active>
+          Shows
+        </BaseTab>
+
+        <BaseTab as={Clickable} onClick={action("onClick")}>
+          Auction Results
+        </BaseTab>
+
+        <BaseTab as={Clickable} onClick={action("onClick")}>
+          Articles
+        </BaseTab>
+
+        <BaseTab as={Clickable} onClick={action("onClick")}>
+          Related Artists
+        </BaseTab>
+      </BaseTabs>
+    )
+  })
+  .add("as={RouterLink}", () => {
+    return (
+      <BaseTabs m={3}>
+        <BaseTab as={RouterLink} to="#example" activeClassName="active" garbage>
+          Example
+        </BaseTab>
+      </BaseTabs>
+    )
+  })
+  .add("Overflowing", () => {
+    return (
+      <BaseTabs m={3}>
+        {[
+          "Artworks (580656)",
+          "Artists (8)",
+          "Collections (7)",
+          "Artist Series (13)",
+          "Galleries (4)",
+          "Shows (3667)",
+          "Categories (21)",
+          "Articles (656)",
+          "Auctions (23)",
+          "More (46)",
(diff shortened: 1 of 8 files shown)


packages/palette/src/elements/BaseTabs/BaseTab.tsx
```diff
@@ -0,0 +1,60 @@
...
+ */
+export type BaseTabProps = TextProps & {
+  active?: boolean
+} & (// FIXME: Should be able to remove this typing once styled-components is updated
+  // Default:
+  | {}
+    // When displaying as an anchor tag, accept anchor props:
+    | ({ as: "a" } & (React.AnchorHTMLAttributes<HTMLAnchorElement>))
+    // Cop-out: if you pass as with a complex component, loosen up the types:
+    | ({ as: React.ReactElement<unknown> } & Record<string, unknown>))
```
dzucconi, line 17: `as` seems straight up broken in our version of styled components. Would really like to push forward with the upgrade. Is that a ticket somewhere?

dzucconi:
Also: it occurs to me that if we adopted [a context-oriented API for tabs](https://chakra-ui.com/docs/components/tabs) we could remove this split between BaseTabs and Tabs. If I'm releasing a breaking change then it might be worth it to clean up the interface like this. 🤷

damassi:
Would be worth adding @jonallured as a reviewer, as he might have some insights around the `sup` usage, or how this might impact volt

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+storiesOf("Components/BaseTabs", module)
+  .add("Default", () => {
+    return (
+      <BaseTabs m={3}>
+        <BaseTab>First</BaseTab>
+        <BaseTab>Second</BaseTab>
+        <BaseTab active>Active</BaseTab>
+        <BaseTab>Last</BaseTab>
+      </BaseTabs>
```
damassi, line 26: I think the answer is `yes` but is the choice to use names like `BaseTabs` and `BaseTab` because, at some higher and more product-dev centric everyday level, we'll see these things implemented and used like 

```tsx
<Tabs>
  <Tab>...</Tab>
</Tabs>
```
?

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+          return (
+            <BaseTab key={label} active={i === 1}>
+              {label}
+            </BaseTab>
+          )
+        })}
+      </BaseTabs>
+    )
+  })
+  .add("Fill", () => {
```
damassi, line 120: What does `fill` mean?

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <BaseTabs m={3} justifyContent="center">
+        <BaseTab active>Alone</BaseTab>
+      </BaseTabs>
+    )
+  })
+  .add("with a separator", () => {
+    return (
+      <BaseTabs
+        m={3}
+        separator={<ChevronIcon mx={2} fill="black30" width="12px" />}
```
damassi, line 175: 👌

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
dzucconi: Extended yes; both in Force and used to build out the client-side tabs here. But also pragmatically. I'd like them to be called `<Tablist>` `<Tab>` but wouldn't be able to do that without drastically altering the API (perhaps worth doing if I'm going to cut a major release anyway).

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
dzucconi: This pattern was mentioned in the Figma specs:

packages/palette/src/elements/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,137 @@
...
+/** Extends `Box`, provides configurable gutter */
+export const BaseTabs: React.FC<BaseTabsProps> = ({
+  fill,
+  separator = DEFAULT_SEPARATOR,
+  children,
+  ...rest
+}) => {
+  const ref = useRef<HTMLDivElement | null>()
+
+  const [paddingProps, boxProps] = splitRailProps(rest)
```
dzucconi, line 85: Woops, need to rename `paddingProps` => `railProps`

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
damassi: > Perhaps worth doing if I'm going to cut a major release anyway

Yup, why not? That naming seems a little better, and if things are gonna be updated then might as well just rip it out and then delete old tab component from palette. 

(Either way, this is non-blocking)

packages/palette/src/elements/BaseTabs/BaseTab.story.tsx
damassi: _finally_

damassi (approved):
This looks great, but would be good to add a couple light behavior tests around 
- `as` prop
- interactions such as `onClick` 
(if they don't already exist)

jonallured:
Sorry took me a while to get this screenshot, but this is how we use `sup` in Volt:

Screen Shot 2020-12-10 at 3 30 11 PM

lmk if I can help!

dzucconi:
@jonallured I get that it just seems weird that it's part of the tab, particularly when there's a native `<sup>` in HTML. <sup>wassup</sup>

dzucconi:
Regardless: I'd PR the migrations in both Force and Volt to handle any breakage.

jonallured (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette@13.25.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.25.0) :rocket:

PR #822, opened 2020-12-04
refactor: Reorganize pagination and improve tests

jonallured:
This PR aims to refactor the pagination code somewhat in order to make changing it easier. It's one of those "make the change easy (this will be hard), make the easy change"-type of PRs that I like to do and you've all grown to "love". 🤣 YMMV, but it might be handy to go commit-by-commit to make the changes easier to see. Totally good to squash though.

My goals:

* separate the `LargePagination` and `SmallPagination` components
* improve test coverage by outlining more of the cases
* remove levels of indirection when possible
* make no visual changes

I think I did pretty alright! Totally open to thoughts though!!

Next step: with this in place, I feel like I can see how I'd go from buttons to links and (unless I'm missing something!) I think there's a pretty clear path to get there.


packages/palette/src/elements/Pagination/LargePagination.tsx
```diff
@@ -0,0 +1,155 @@
+import React from "react"
+import styled, { css } from "styled-components"
+import { color } from "../../helpers/color"
+import { ChevronIcon } from "../../svgs/ChevronIcon"
+import { Box } from "../Box"
+import { Flex } from "../Flex"
+import { Sans } from "../Typography"
+
+export interface Props {
+  onClick?: (cursor: string, page: number) => void
```
dzucconi, line 10: I'm surprised we're encoding knowledge about cursors, etc into what should be a display component (I realize this was already here)

packages/palette/src/elements/Pagination/SmallPagination.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <Flex flexDirection="row" width="100%">
+      <PrevButton enabled={previous} onClick={handlePrevClick} />
+      <NextButton enabled={hasNextPage} onClick={handleNextClick} />
+    </Flex>
+  )
+}
+
+const PrevButton = ({ enabled, onClick }) => {
```
damassi, line 37: Can we add types here?

packages/palette/src/elements/Pagination/LargePagination.tsx
```diff
@@ -0,0 +1,155 @@
...
+const Page = ({ onClick, pageCursor }) => {
+  const { cursor, isCurrent, page } = pageCursor
+
+  const handleClick = () => {
+    onClick(cursor, page)
+  }
+
+  return (
+    <Button active={isCurrent} onClick={handleClick}>
+      <Sans size="3" weight="medium" display="inline">
```
dzucconi, line 70: It's probably worth updating these to use `Text` component (`<Text variant="mediumText" ...`) which correspond to the updated typographic styles.

damassi (approved):
Looks great 👍

packages/palette/src/elements/Pagination/LargePagination.tsx
dzucconi: > make no visual changes

Yep, makes sense!

dzucconi (approved):
The goals of this PR look achieved to me 🎩  ✅

packages/palette/src/elements/Pagination/LargePagination.tsx
damassi: There's been some back and forth about this (and I agree). This pattern is used across a couple codebases due to how we've designed our backend pagination cursor implentation, with `cursor` being used as part of the language. We could say "page" or whatever. 

Once we have a design for handling actual links, this can be refactored out, but it will also impact volt.

packages/palette/src/elements/Pagination/SmallPagination.tsx
jonallured: Yeah good call. I didn't initially deal with this because what I found wasn't well typed, but no reason to not do it! Took a stab at this with b4a54b8.

packages/palette/src/elements/Pagination/LargePagination.tsx
jonallured: I'll come back for this in a follow up 👍

artsyit:
:rocket: PR was released in [`@artsy/palette@13.24.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.24.1) :rocket:

PR #823, opened 2020-12-07
chore(webfonts): update webfonts domain references

dzucconi:
Re: [FX-2540](https://artsyproduct.atlassian.net/browse/FX-2540)

These were changed to `production-webfonts` during the transition. The bare `webfonts` subdomain now points to the same Cloudfront distro as `production-webfonts` and so this can be restored.


artsyit:
:rocket: PR was released in [`@artsy/palette@13.24.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.24.1) :rocket:

PR #824, opened 2020-12-08
refactor: Use links for pagination

jonallured:
Sibling here: https://github.com/artsy/force/pull/6832

This PR aims to mostly match our existing pagination component with some subtle visual differences BUT use A tags rather than BUTTON tags so that we can provide URLs for the Google bots. Note that this is a draft until I add another commit that actually sets the HREF attr. At this point I'm just setting it to `""` but wanted to get this in front of people to see if I'm on the right track.

So yeah, what I did was wrap a `Link` around the various pagination parts and then migrate the `onClick` over to it and then refactor/simplify what I found so that I could match the existing style.

Then I hopped on a call with @deliciousnachos to show her the places where I diverged:

* Individual page numbers are links so they underline on hover rather than have a background color change
* I'm using `Text` rather than `Sans` so there are subtle font changes when you zoom

That's it! Here are some pretty pictures:

<details>
<summary>before</summary>
Screen Shot 2020-12-08 at 3 55 18 PM
</details>
<details>
<summary>after</summary>
Screen Shot 2020-12-08 at 3 55 22 PM
</details>

## Next step

The next step is to correctly compute the href for these links. I'm a little unsure how to proceed here because I don't always have access to the page number. I'm hoping to hop on a call with @dzucconi or @damassi and get their feedback.

I also need to link this locally to force or whatever and ensure that I haven't actually broken anything, haha.


dzucconi:
Looks like you're on the right track to me.

> because I don't always have access to the page number

You should though?

Pass the whole component a base `href` and then you just stringify the page params for the query string. 

The complication I think you'll run into is wanting to preserve existing page query params (for filters etc) — since this has to be rendered on the server side, you can't just check `location.search`. Injecting a custom page component from the outer context is probably going to be the move here.

packages/palette/src/elements/Pagination/LargePagination.tsx
```diff
@@ -84,20 +83,25 @@ const Page: React.FC<PageProps> = props => {
...
+  const href = ""
+
   return (
-    <Button active={isCurrent} onClick={handleClick}>
-      <Sans size="3" weight="medium" display="inline">
-        {page}
-      </Sans>
-    </Button>
+    <Box bg={highlight} borderRadius={2}>
+      <Link href={href} onClick={handleClick} underlineBehavior="hover">
```
damassi, line 91: We've talked about it, but it will be interesting to see how this behaves within the Router. I think, for the first pass, we'll have to pass in an `onClick` handler in force that promptly does 

```
event.preventDefault() 
```

which will cancel the `href` behavior. 

(We can talk more if this doesn't make sense @jonallured)

jonallured:
Sibling here: https://github.com/artsy/force/pull/6832

damassi:
@jonallured - this is ready for review

jonallured:
I think this is ready to go, huge thank you to @damassi for getting it over the finish line! 🤗 

Going to merge and then see about the sibling PR over in Force.

artsyit:
:rocket: PR was released in [`@artsy/palette@13.27.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.27.0) :rocket:

PR #829, opened 2020-12-14
fix(tabs): converts back to anys

dzucconi:
I had meant to switch these to generics but will do that when I actually cut the breaking release. Moving these back to anys for now.


artsyit:
:rocket: PR was released in [`@artsy/palette@13.25.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.25.1) :rocket:

PR #830, opened 2020-12-15
fix(basetabs): support nested fragments

dzucconi:
Previously if we were using a fragment within the tabs we'd get:

This PR introduces a simple helper to flatten out the fragments:


artsyit:
:rocket: PR was released in [`@artsy/palette@13.26.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.26.0) :rocket:

PR #831, opened 2020-12-15
fix(swiper): inlines default margins to simplify setting alternatives

dzucconi:
This change allows us to, for instance, use an `mt` value which will override the `my` default.

Re: https://github.com/styled-system/styled-system/issues/599


artsyit:
:rocket: PR was released in [`@artsy/palette@13.26.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.26.1) :rocket:

PR #832, opened 2020-12-17
fix(skeleton): implements appropriate scaling for multiline texts

dzucconi:
Previously, multi-line texts would have the underlaid text peek out, because the overlay was scaled using a percentage.

Now it just tucks in the overlay a small amount, scaled to the current font size. The text is also transparent incase an edge peeks out around an aggressive border radius.


artsyit:
:rocket: PR was released in [`@artsy/palette@13.26.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.26.2) :rocket:

PR #839, opened 2021-01-07
fix: Populate pagination next/prev hrefs correctly

jonallured:
After doing some QA on staging I noticed a bug I made when I wrote an untested bit of logic about determining the next page number. I got it right when the previous page existed but on the first page, this bit of logic would return `1` and cause the Next button's HREF to be incorrect. SIGH.

How many times must I re-learn that I need tests!!

So, this PR introduces a set of tests that verify both the Next and Prev button HREF values. It ensures that the first and last pages have the correct empty hrefs and that each step along the way has the right page param being set.

https://artsyproduct.atlassian.net/browse/GRO-146

/cc @artsy/grow-devs

### DIFF
--- a/packages/palette/src/elements/Pagination/LargePagination.tsx
+++ b/packages/palette/src/elements/Pagination/LargePagination.tsx
@@ -72,7 +72,7 @@ export const LargePagination = (props: PaginationProps) => {
     )
   })
 
-  const nextPage = (previous?.page || -1) + 2
+  const nextPage = (previous?.page || 0) + 2
 
   const DotDotDot = () => {
     return (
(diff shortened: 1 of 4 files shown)


damassi:
hah, there we go 👌

artsyit:
:rocket: PR was released in [`@artsy/palette-tokens@1.2.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-tokens%401.2.2), [`@artsy/palette@13.27.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.27.2) :rocket:

PR #841, opened 2021-01-14
fix(button): button updates

dzucconi:
Re: [FX-2601](https://artsyproduct.atlassian.net/browse/FX-2601)

I'm going to publish canaries and see how far I can get without a breaking change.

- [x] Refactored the class components into functional components so they are a bit easier to read
- [x] Extends native button types so we have correct typing for attrs like (`tabIndex` — I thought about getting rid of the double focus problem by just -1-ing the tab index, but I couldn't do that until I had the real typing)
- [x] kill the `longestText` API by just handling that one-off in the children in force
- [x] pass display props correctly
- [x] break tokens out of functions and simplify using styled-system variants

Punting on these for now
- [ ] support `as="a"` (works but getting the typing correct without updating deps is gonna be trouble)
- [ ] update to new typography `Text` component

At the very least we can actually wrap buttons in links then use tabIndex to disable the extraneous focus

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@13.28.1-canary.841.14653.0
  # or 
  yarn add @artsy/palette@13.28.1-canary.841.14653.0
  ```
</details>


packages/palette/src/elements/Button/Button.tsx
```diff
@@ -172,214 +168,219 @@ export const getStylesForVariant = (variant: ButtonVariant) => {
...
+  color,
+  size,
+  longestText,
+  weight,
+  className,
+  onClick,
+  ...rest
+}) => {
+  const loadingClass = loading ? "loading" : ""
+  const disabledClass = disabled ? "disabled" : ""
```
dzucconi, line 330: I don't suppose there's any real reason these are classes? Do we do some kind of weird DOM manipulation somewhere where we have to add/remove classes?

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -172,214 +168,219 @@ export const getStylesForVariant = (variant: ButtonVariant) => {
...
-        className={[loadingClass, disabledClass, className]
-          .filter(Boolean)
-          .join(" ")}
-        onClick={this.onClick}
-        disabled={disabled}
-        {...rest}
-      >
-        {loading && <Spinner size={this.props.buttonSize} />}
-
-        {longestText ? (
```
dzucconi, line 356: I remember reading the PR for this one 😈

packages/palette/src/elements/Button/Button.tsx
damassi: I'm a little worried to remove as I'm sure this was added for a reason -- _but_ I'm also thinking that whatever relied on adding / removing classes in this way is super old and thus fine to break. Better to just clean up confusing primitive code and wait for an issue to arise.

packages/palette/src/elements/Button/Button.tsx
damassi: Kieren driving his BMW around lol

damassi (approved):
Thanks for converting to an FC, much nicer. LGTM 👍

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -61,111 +61,6 @@ export interface ButtonBaseProps
...
-              background-color: ${hover.backgroundColor};
-              border-color: ${hover.borderColor};
-              color: ${hover.color};
-            }
-          }
-        `
-    }};
-  `
-}
-
```
dzucconi, line 174: This is interesting: I can't find anything that uses these functions in palette/force/volt. Seems safe to remove @damassi?

damassi: I would double check the PR that this was introduced in to see if there's a paper trail, but yeah, I don't recall and this seems a bit unorthodox for whatever it may do 👍

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -91,123 +77,32 @@ const getSize = ({
...
-              background-color: transparent;
-              border-color: transparent;
-              color: ${colors.black100};
-            `
-        }};
-      `
-    default:
-  }
-}
-
```
dzucconi, line 302: 👋

packages/palette/src/elements/Button/Button.tsx
dzucconi: It *looks* like it was just put here by mistake when the React Native components were moved out.

packages/palette/src/elements/Button/Button.tsx
damassi: Awesome update with the `variant` -- I've never noticed that fn before. Also didn't know that would could handle responsivity using object notation.

packages/palette/src/elements/Button/Button.tsx
```diff
@@ -24,8 +24,6 @@ export interface ButtonProps
   loading?: boolean
   /** Uses inline style for button */
   inline?: boolean
-  /** This does nothing at all: TODO: Remove */
-  block?: boolean
```
dzucconi, line 61: lol I kept setting this thinking it was going to do what it said it would (previously: make button full-width)

dzucconi:
* Force looks good with the canary once https://github.com/artsy/force/pull/6904 is merged
* Taking a look at Volt now

damassi (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette@13.29.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.29.0) :rocket:

PR #842, opened 2021-01-15
fix(button): ensure loading styles take precedence over hover

dzucconi:
Minor thing here where the loading state wasn't taking precedence over hover; so you'd see the two states conflated.


artsyit:
:rocket: PR was released in [`@artsy/palette@13.29.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.29.1) :rocket:

PR #847, opened 2021-02-04
Use full 6 digits for hex colors

pvinis:
Makes things easier. We use this to pass colors in the objC part of the app too, and we could convert them to 6 digits but feels nicer to have them 6 digits directly.


damassi:
@zephraph is OOO so gonna merge

artsyit:
:rocket: PR was released in [`@artsy/palette-tokens@1.4.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-tokens%401.4.0) :rocket:

PR #852, opened 2021-02-11
feat: Add ability to pass in a component into Toggle label prop

annacarey:
Added option to pass a component instead of a string to the Toggle component.

What is the best way to test this out in the UI w/o bringing it into another app? Seems like the Playground is a bit limiting to make sure this looks how I want it to.

Storybook:
Screen Shot 2021-02-11 at 2 35 12 PM

Update to docs:
Screen Shot 2021-02-11 at 11 26 29 AM

### DIFF
--- a/packages/palette/src/elements/Toggle/Toggle.story.tsx
+++ b/packages/palette/src/elements/Toggle/Toggle.story.tsx
@@ -18,13 +18,6 @@ const SecondaryAction = () => {
   )
 }
 
-const labelComponent = (
-  <>
-    <Text variant="mediumText">Heading</Text>
-    <Text variant="text">Subheading</Text>
-  </>
-)
-
 storiesOf("Components/Toggle", module)
   .add("Toggle", () => {
     return (
@@ -71,7 +64,12 @@ storiesOf("Components/Toggle", module)
     return (
       <Box width="350px">
         <Toggle
-          label={labelComponent}
+          label={
+            <>
+              <Text variant="mediumText">Heading</Text>
+              <Text variant="text">Subheading</Text>
+            </>
+          }
           expanded
         >
           <h1>Hello world</h1>
--- a/packages/palette/src/elements/Toggle/Toggle.story.tsx
+++ b/packages/palette/src/elements/Toggle/Toggle.story.tsx
@@ -2,12 +2,13 @@ import { storiesOf } from "@storybook/react"
 import React from "react"
 import { Box } from "../Box"
 import { Link } from "../Link"
+import { Text } from "../Text"
 import { Toggle } from "./Toggle"
 
 const SecondaryAction = () => {
   return (
     <Link
-      onClick={e => {
+      onClick={(e) => {
         alert("hello world!")
         e.stopPropagation()
       }}
@@ -16,6 +17,13 @@ const SecondaryAction = () => {
     </Link>
   )
 }
+
+const labelComponent = (
+  <>
+    <Text variant="mediumText">Heading</Text>
+    <Text variant="text">Subheading</Text>
+  </>
+)
 
 storiesOf("Components/Toggle", module)
   .add("Toggle", () => {
@@ -53,6 +61,18 @@ storiesOf("Components/Toggle", module)
           expanded
           disabled
           renderSecondaryAction={SecondaryAction}
+        >
+          <h1>Hello world</h1>
+        </Toggle>
+      </Box>
+    )
+  })
+  .add("Toggle with a component as the label", () => {
+    return (
+      <Box width="350px">
+        <Toggle
+          label={labelComponent}
+          expanded
         >
           <h1>Hello world</h1>
         </Toggle>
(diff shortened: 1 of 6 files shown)


packages/palette/src/elements/Toggle/__tests__/Toggle.test.tsx
```diff
@@ -38,13 +39,22 @@ describe("Toggle", () => {
...
     expect(chevron.props().height).toEqual(12)
   })
+
+  it("renders proper component when passed component as label", () => {
+    const component = <Text>Hello</Text>
+    const wrapper = mount(<Toggle label={component}>tab content</Toggle>)
+    expect(wrapper.find("Text")).toHaveLength(1)
+  })
+
+  it.only("renders Sans component when passed string as label", () => {
```
dzucconi, line 54: You'll want to remove this `.only` (We need to install a linting rule that catches these pre-commit)

dzucconi:
Re: testing this; I'd write a storybook story; which I find much more useful than the playground — it'll be included in the visual diffing specs which have been great at catching visual regressions

damassi (approved):
Looks great, and second Damon on the storybook entry. (Lots of examples throughout palette.)

damassi:
>  w/o bringing it into another app

@annacarey are you familiar with `yarn link` ?

packages/palette/src/elements/Toggle/__tests__/Toggle.test.tsx
annacarey: Oh yes thanks for catching!

annacarey:
> > w/o bringing it into another app
> 
> @annacarey are you familiar with `yarn link` ?

I wasn't but just looked it up. Trying to use it to link palette into volt, but I don't think it's finding the correct package.

packages/palette/src/elements/Toggle/Toggle.story.tsx
```diff
@@ -17,6 +18,13 @@ const SecondaryAction = () => {
   )
 }
 
+const labelComponent = (
```
damassi, line 21: #superminor - Since this is a react component we should always use `TitleCase`. (The real-world requirement here is due to the fact that react hooks don't work  inside of lowercase-named components, amongst other things.) Also no need to include "component" in var name, can keep it simple like: `const Label = ...`

dzucconi: This is the kind of thing I'd just inline. Typically when I see this I think someone just wants to extract a component; in this case nothing complex is happening to warrant the assignment.

damassi: (We can inline it, but the comment above is more of a rule of thumb when writing in-general react code)

annacarey: Moved it in-line! But also thanks for calling out b/c I often forget when I'm defining a small React component within a file vs. just a regular variable.

packages/palette/src/elements/Toggle/Toggle.tsx
```diff
@@ -57,21 +57,34 @@ export class Toggle extends React.Component<ToggleProps> {
...
-    const { children, chevronSize, label, textSize, renderSecondaryAction } = this.props
+    const {
+      children,
+      chevronSize,
+      label,
+      textSize,
+      renderSecondaryAction,
+    } = this.props
+
+    const labelComponent =
```
annacarey, line 68: @damassi and @dzucconi should this also be moved inline/defined with title case?

damassi: Ah, so, the rules are funny...

This is slightly different in that this is either a string (`label`) or straight up JSX (the `<Sans>` component); what makes a component a formal react component (per the convo in the other thread) is that its a function:

```tsx
const Label = props => {
  return <Sans />
}
```
To use the above, you have to invoke it:

```tsx
<Label />
```

(By 'invoke', it's helpful to think of <Label /> as equaling `Label()` -- invoking a function.) 

Wheres something like 
```tsx
const Label = isString ? label : <Sans>Hi</Sans> 
```

can't actually be invoked like 

```
<Label />
```
Because a) it could be a plain string; or b) if its not a string, then its just a reference to _already invoked_ JSX.

So in this case i think its fine to leave as is, and since it can be a string or component `labelComponent` is basically "label (string) or labelComponent (component)"; the naming emcompasses both possible forms.

annacarey: Ahhh yes okay really appreciate this explanation, _so_ helpful for solidifying my understanding of how React works under the hood.

Is this ready to merge? Does that storybook test need to be reviewed manually?

artsyit:
:rocket: PR was released in [`@artsy/palette@13.32.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.32.0) :rocket:

packages/palette/src/elements/Toggle/Toggle.tsx
damassi: Nope all good!

PR #855, opened 2021-02-17
fix(checkbox): makes checkbox accessible; supports box props

dzucconi:
This PR fixes two issues with the checkbox:

1. It makes it accessible: can now be focused, selected with keyboard, and has proper aria tags
2. The label container was collapsing meaning you couldn't have checkbox labels that occupy the full-width of their container, this is the only noticeable change (flex: 1 on the label wrapper).

The component also now supports the full set of props as it's parent (clickable) — so it's a proper `Box` now.

### DIFF
--- a/packages/palette/src/elements/Checkbox/Check.tsx
+++ b/packages/palette/src/elements/Checkbox/Check.tsx
@@ -1,0 +1,74 @@
+import React from "react"
+import styled, { css } from "styled-components"
+import { color, space } from "../../helpers"
+import { CheckIcon } from "../../svgs"
+import { Box } from "../Box"
+import { CheckboxProps } from "./Checkbox"
+
+const SIZE = 2 // 20px
+const BORDER_WIDTH = 2 // 2px
+
+const Container = styled(Box)<CheckProps>`
+  width: ${space(SIZE)}px;
+  height: ${space(SIZE)}px;
+  transition: background-color 0.25s, border-color 0.25s;
+
+  ${({ disabled, selected, error }) => {
+    switch (true) {
+      case disabled:
+        return css`
+          background-color: ${color("black5")};
+          border-color: ${color("black10")};
+        `
+      case selected:
+        return css`
+          background-color: ${color("black100")};
+          border-color: ${color("black100")};
+        `
+      case error:
+        return css`
+          background-color: ${color("white100")};
+          border-color: ${color("red100")};
+        `
+      default:
+        return css`
+          background-color: ${color("white100")};
+          border-color: ${color("black10")};
+        `
+    }
+  }}
+
+  svg {
+    position: relative;
+    top: -${BORDER_WIDTH}px;
+    left: -${BORDER_WIDTH}px;
+  }
+`
+
+export interface CheckProps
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/Checkbox/Checkbox.story.tsx
```diff
@@ -8,7 +10,16 @@ export default {
...
 export const States = () => {
-  const states = [{}, { selected: true }, { disabled: true }, { error: true }]
+  const states = [
+    {},
+    { selected: true },
+    { disabled: true },
+    { error: true },
+    { selected: true, disabled: true },
+    { error: true, disabled: true },
+    { selected: true, error: true, disabled: true },
```
dzucconi, line 20: This elusive state is handled now too; for what it's worth.

packages/palette/src/elements/Checkbox/Checkbox.story.tsx
```diff
@@ -32,7 +43,22 @@ export const Demo = () => {
         action("onClick")(selected)
       }}
     >
-      <Text>click me</Text>
+      <Text lineHeight="solid">
+        use a `solid` line-height to ensure vertical centering
```
dzucconi, line 47: I see in force some usage of `pt="1px"` in the label because the default line-height/font-size combo is taller than the checkbox label; using a `solid` line-height is the appropriate fix there; so just pointing it out since it's probably common.

packages/palette/src/elements/Checkbox/Check.tsx
```diff
@@ -0,0 +1,74 @@
...
+const SIZE = 2 // 20px
+const BORDER_WIDTH = 2 // 2px
+
+const Container = styled(Box)<CheckProps>`
+  width: ${space(SIZE)}px;
+  height: ${space(SIZE)}px;
+  transition: background-color 0.25s, border-color 0.25s;
+
+  ${({ disabled, selected, error }) => {
+    switch (true) {
```
damassi, line 17: I like this `switch(true)` pattern 👍

packages/palette/src/elements/Checkbox/Checkbox.tsx
```diff
@@ -33,144 +18,78 @@ export interface CheckboxProps {
...
-
-  svg {
-    position: relative;
-    top: -${BORDER_WIDTH}px;
-    left: -${BORDER_WIDTH}px;
+  const handleClick = (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>
+  ) => {
+    !disabled && onSelect && onSelect(!selected)
+    onClick && onClick(event)
```
damassi, line 42: Totally non blocking... long term across a wide dev team, conventional `if`s are more readable, less bug prone, less prone to needing to be unwound (like implicityly returned react components), more "boring" (better), etc.

damassi (approved):
LGTM 👍

artsyit:
:rocket: PR was released in [`@artsy/palette@13.33.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.33.0) :rocket:

PR #857, opened 2021-02-18
fix(checkbox): makes a11y updates without using button

dzucconi:
Thanks for catching this. This PR reverts my revert. Then 5a4946ea8b6a34061c0fd32bad215dd5b2880425 changes it from a button back to a div, but handles the a11y changes manually.

### DIFF
--- a/packages/palette/src/elements/Checkbox/Check.tsx
+++ b/packages/palette/src/elements/Checkbox/Check.tsx
@@ -1,0 +1,74 @@
+import React from "react"
+import styled, { css } from "styled-components"
+import { color, space } from "../../helpers"
+import { CheckIcon } from "../../svgs"
+import { Box } from "../Box"
+import { CheckboxProps } from "./Checkbox"
+
+const SIZE = 2 // 20px
+const BORDER_WIDTH = 2 // 2px
+
+const Container = styled(Box)<CheckProps>`
+  width: ${space(SIZE)}px;
+  height: ${space(SIZE)}px;
+  transition: background-color 0.25s, border-color 0.25s;
+
+  ${({ disabled, selected, error }) => {
+    switch (true) {
+      case disabled:
+        return css`
+          background-color: ${color("black5")};
+          border-color: ${color("black10")};
+        `
+      case selected:
+        return css`
+          background-color: ${color("black100")};
+          border-color: ${color("black100")};
+        `
+      case error:
+        return css`
+          background-color: ${color("white100")};
+          border-color: ${color("red100")};
+        `
+      default:
+        return css`
+          background-color: ${color("white100")};
+          border-color: ${color("black10")};
+        `
+    }
+  }}
+
+  svg {
+    position: relative;
+    top: -${BORDER_WIDTH}px;
+    left: -${BORDER_WIDTH}px;
+  }
+`
+
+export interface CheckProps
+  extends Pick<CheckboxProps, "disabled" | "selected" | "error"> {}
+
+/** Toggeable check mark */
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/Checkbox/Check.tsx
```diff
@@ -68,7 +68,7 @@ export const Check: React.FC<CheckProps> = ({
       selected={selected}
       {...rest}
     >
-      <CheckIcon fill={iconColor()} />
+      <CheckIcon aria-hidden fill={iconColor()} />
```
dzucconi, line 71: Caught this when doing a voiceover pass. All our SVGs have names so this was calling out "Check" here.

packages/palette/src/elements/Checkbox/Checkbox.tsx
```diff
@@ -1,11 +1,13 @@
 import React from "react"
 import styled, { css } from "styled-components"
 import { color } from "../../helpers"
-import { Clickable, ClickableProps } from "../Clickable"
```
pepopowitz: impertinent to this PR & non-blocking -- out of curiosity should [the `Clickable` component](https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Clickable/Clickable.tsx) specify `type="button"` on its underlying `button` component? I'm not sure how much we're using it, and where, but I think I'd be surprised if I wrapped something in `clickable` and it acted as a `type="submit"` button instead of `type="button"`.

packages/palette/src/elements/Checkbox/Checkbox.tsx
```diff
@@ -58,6 +64,8 @@ export const Checkbox: React.FC<CheckboxProps> = ({
       disabled={disabled}
       error={error}
       onClick={handleClick}
+      tabIndex={0}
```
pepopowitz, line 67: I'm not sure offhand what we do about tabIndexes - will this have any weird tab order effects on pages in force? Is it something that needs to be overrideable?

pepopowitz (approved):
I think it looks great! I have some questions but I don't think any of them are blockers.

artsyit:
:rocket: PR was released in [`@artsy/palette@13.34.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.34.0) :rocket:

packages/palette/src/elements/Checkbox/Checkbox.tsx
dzucconi: A tabIndex of 0 just means that it's in the natural tab order. -1 takes it out of the tab order (but can be manually focused). And just never use tabIndexes over 0 :) — which jumps the element to the front of the tab order

packages/palette/src/elements/Checkbox/Checkbox.tsx
dzucconi: Yeah this is a good point/good idea. Will take a look at this separately.

packages/palette/src/elements/Checkbox/Checkbox.tsx
dzucconi: Basically to get something properly accessible when it's not a button: Give it a tabIndex of 0 to bring it into tab order, give it a role (button/checkbox whatever), then add a keypress handler. Here I'm only triggering on spacebar which is what [wai says to do](https://www.w3.org/TR/wai-aria-practices/examples/checkbox/checkbox-1/checkbox-1.html); interestingly I thought Enter would have been supported as well but nope; enter doesn't check native checkboxes either.

PR #858, opened 2021-02-18
Change default Clickable type; minor checkbox fix

dzucconi:
@pepopowitz good idea re: Clickable; for some reason I thought the default button type was `button` but it's actually `submit`

We've never used a Clickable as a submit button before in Force so this shouldn't have any effect/I don't feel like we need to cut a breaking change... but let me know if you think otherwise.

I can't capture this in a test because Enzyme doesn't support event propagation; so just a story for now.

Also: I forgot to prevent the default of spacebar which is to scroll the page. Publishing a canary to verify in force.

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@13.33.2-canary.858.15065.0
  # or 
  yarn add @artsy/palette@13.33.2-canary.858.15065.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Checkbox/Checkbox.tsx
+++ b/packages/palette/src/elements/Checkbox/Checkbox.tsx
@@ -49,7 +49,8 @@ export const Checkbox: React.FC<CheckboxProps> = ({
   }
 
   const handleKeyPress = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    if (event.code === "Space" && isSelectable) {
+    if (event.key === " " && isSelectable) {
+      event.preventDefault()
       onSelect(!selected)
     }
   }
--- a/packages/palette/src/elements/Checkbox/Checkbox.story.tsx
+++ b/packages/palette/src/elements/Checkbox/Checkbox.story.tsx
@@ -36,17 +36,19 @@ export const States = () => {
 export const Demo = () => {
   const [isSelected, setSelected] = useState(false)
   return (
-    <Checkbox
-      selected={isSelected}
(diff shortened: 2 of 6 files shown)


dzucconi:
I guess I could also just switch back to using Clickable in the checkbox 😅

dzucconi:
Not sure why a canary wasn't published. I'm having issues with yalc/yarn link sooooo 🤔

pepopowitz (approved):
Nice! 

I agree about a patch version being adequate.

damassi:
> I'm having issues with yalc/yarn link

This force workflow has never failed me:

```bash
cd palette/packages/palette
yarn link
yarn watch

# new pane 
cd force 
yarn link @artsy/palette  
yarn start
```

packages/palette/src/elements/Checkbox/Checkbox.tsx
```diff
@@ -49,7 +49,8 @@ export const Checkbox: React.FC<CheckboxProps> = ({
   }
 
   const handleKeyPress = (event: React.KeyboardEvent<HTMLDivElement>) => {
-    if (event.code === "Space" && isSelectable) {
+    if (event.key === " " && isSelectable) {
```
dzucconi, line 52: Weird thing: in Force event.code is undefined; event.key is fine in this case/more widely supported anyway; but odd.

artsyit:
:rocket: PR was released in [`@artsy/palette@13.34.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.34.1) :rocket:

PR #860, opened 2021-02-22
fix(toggle): supports keyboard accessibility

dzucconi:
Trying to make filters fully keyboard accessible.

Not much to this:

* refactored the component from class => functional
* made the header a `Clickable`
* added aria-expanded tag

One thing I didn't handle with this PR is doing the appropriate labelledby/region markup. We need to be able to generate unique IDs that are consistent across SSR/client-side render. Loathe to add a dependency; but looking at what [react-uid](https://github.com/thearnica/react-uid) is doing.

Another little wrinkle here is that I want to upgrade this to the `Text` component but will have to cut a breaking change for that since the types will break. Will probably do that separately picking up where https://github.com/artsy/palette/pull/780 left off.

----

Also my evergreen reminder that adding the accessibility dropdown to your top bar makes testing this stuff super simple:  🎩

### DIFF
--- a/packages/palette/src/elements/Toggle/Toggle.story.tsx
+++ b/packages/palette/src/elements/Toggle/Toggle.story.tsx
@@ -1,6 +1,5 @@
 import React from "react"
 import { States } from "storybook-states"
-import { Box } from "../Box"
 import { Link } from "../Link"
 import { Text } from "../Text"
 import { Toggle, ToggleProps } from "./Toggle"
@@ -28,7 +27,8 @@ export const Default = () => {
       states={[
         {},
         { expanded: true },
-        { expanded: true, disabled: false },
+        { expanded: false, disabled: true },
+        { expanded: true, disabled: true },
         { renderSecondaryAction: SecondaryAction },
         { renderSecondaryAction: SecondaryAction, disabled: true },
         {
@@ -41,7 +41,7 @@ export const Default = () => {
         },
       ]}
     >
-      <Toggle label="Example">
+      <Toggle label="Example" maxWidth={350}>
         <Text>Expanded content</Text>
       </Toggle>
     </States>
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/Toggle/Toggle.story.tsx
```diff
@@ -21,81 +21,29 @@ export default {
...
+        { renderSecondaryAction: SecondaryAction, disabled: true },
+        {
+          label: (
+            <>
+              <Text variant="mediumText">Heading</Text>
+              <Text variant="text">Subheading</Text>
+            </>
+          ),
+        },
+      ]}
```
annacarey, line 42: this syntax is so much cleaner

packages/palette/src/elements/Toggle/Toggle.tsx
```diff
@@ -28,91 +27,76 @@ export interface ToggleState {
...
+        <Flex justifyContent="space-between" alignItems="center">
+          {typeof label === "string" ? (
+            <Sans
+              size={textSize as SansSize}
+              weight="medium"
+              color="black100"
+              my={0.5}
+            >
+              {label}
+            </Sans>
```
annacarey, line 67: It's weird to me that changing this to `<Text>` doesn't work because of types? This is the only place we're still using `Sans` right?

annacarey (approved):
Love all these changes! Bummer about the `<Text>` thing. Def okay to ship but question in line about what you're seeing.

packages/palette/src/elements/Toggle/Toggle.tsx
dzucconi: When we change it have to also change the call sites where there would be type errors because it'd go from `textSize?: SansSize` to `variant?: TextVariant`. There may not be many at all of these in the wild; but it'd still be technically a breaking change. There's a bunch of breaking changes that I think might make sense to just handle in one go though.

artsyit:
:rocket: PR was released in [`@artsy/palette@13.34.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.34.2) :rocket:

PR #864, opened 2021-02-25
fix: check for empty strings in join

MounirDhahri:
Co-authored-by: David Sheldrick <djsheldrick@gmail.com>

It looks `_react.default.Children.toArray` is returning an array containing an empty string which is causing the UI to break. We simply added a check to make sure that the child actually exists.

Screenshot 2021-02-25 at 13 35 26

### DIFF
--- a/packages/palette/src/elements/Join/Join.tsx
+++ b/packages/palette/src/elements/Join/Join.tsx
@@ -27,22 +27,24 @@ interface JoinProps {
 export const Join: SFC<JoinProps> = ({ separator, children }) => {
   const childArray = React.Children.toArray(children) as any
 
-  return childArray.reduce((acc, curr, currentIndex) => {
-    acc.push(
-      React.cloneElement(curr as React.ReactElement<any>, {
-        key: `join-${currentIndex}`,
-      })
-    )
-
-    if (currentIndex !== childArray.length - 1) {
+  return childArray
+    .filter((child) => child)
+    .reduce((acc, curr, currentIndex) => {
       acc.push(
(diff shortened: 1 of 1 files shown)


MounirDhahri:
This might be an issue with react related to react, I will check out `ReactChildren.toArray` implementation

packages/palette/src/elements/Join/Join.tsx
```diff
@@ -27,7 +27,7 @@ interface JoinProps {
 export const Join: SFC<JoinProps> = ({ separator, children }) => {
   const childArray = React.Children.toArray(children) as any
 
-  return childArray.reduce((acc, curr, currentIndex) => {
+  return childArray.filter(child=> child).reduce((acc, curr, currentIndex) => {
```
starsirius, line 30: Do we need an empty space before `=>`? (Or is it actually prettier recommended style?) 🤔

MounirDhahri: oh yes, I will remove it 😅

MounirDhahri: Sorry for the big diff, prettier messed things up

MounirDhahri: picard-facepalm

starsirius (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette@13.35.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.35.0) :rocket:

PR #865, opened 2021-02-26
V3 Theme

dzucconi:
The main question I'm going back and forth with is: should you be able to use V2 tokens and variants in a V3 context tree if they happen to not overlap with the V3 tokens/variants. We could, for instance, merge the v2 theme into the v3 theme, favoring v3 keys/values. 

At the moment you can't use tokens from the old theme in the new theme; but you can use old `Text` variants (I could also block this by switching on the `theme.id` to load in the appropriate variants. I'm kind of leaning towards that but am wondering what to do about something like these `Theme/` stories, which are using the v3 variants. We could include a mapping to swap out different styles depending on the theme, which is similar to what I'm planning on doing for some of the other components, like buttons: pull out the active theme ID using `useTheme` then switch between two different styling config maps.

Everything is pretty straightforward except for the `Text` variants; which I'll leave some comments inline about. How to clean that up is dependent on which approach we wind up taking. The other thing to note there is I'm not even bothering to load the Sans/Serif tokens in the new theme either.

I started kind of just roughing in the layout from the Figma as a way of just dog fooding some of this.

localhost_6006_iframe html_id=theme--typography viewMode=story

localhost_6006_iframe html_id=theme--colors viewMode=story

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@13.35.3-canary.865.15376.0
  # or 
  yarn add @artsy/palette@13.35.3-canary.865.15376.0
  ```
</details>


packages/palette/.storybook/preview.js
```diff
@@ -1,40 +1,59 @@
...
+              <Story />
+            </Tab>
+
+            <Tab name="v3">
+              <Story />
+            </Tab>
+          </Tabs>
+        </>
+      </Theme>
+    )
```
dzucconi, line 37: This is the decorator which loads the tabs for every story and swaps out the theme. I'm persisting the active tab in local storage so that they stay active between stories.

packages/palette/src/elements/Text/tokens/v3.ts
```diff
@@ -0,0 +1,41 @@
...
+  },
+  sm: {
+    fontSize: "16px",
+    lineHeight: "24px",
+  },
+  xs: {
+    fontSize: "13px",
+    lineHeight: "18px",
+  },
+}
```
dzucconi, line 38: Instead of having a whole zoo of font-sizes, line-heights, letter-spacings; just keeping it simple and these are the only type treatments allowed. You can step outside of this by passing raw values.

We only use the tokenized sizes/line-heights in a few places in Force, so I'm thinking I can inline those values in force then do the same thing with the old text tokens which should simplify things a bit.

packages/palette/src/themes/Themes.story.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </GridColumns>
+    </Join>
+  )
+}
+
+export const Typography = () => {
+  const { theme } = useTheme()
+
+  const variants =
+    theme.id === "v2" ? theme.textVariants.large : theme.textVariants
```
dzucconi, line 162: Switching on the id like this gives you back the type safe theme. You can also pass a theme type: `useTheme<ThemeV3>()` if you know you want a specific one.

packages/palette/src/themes/v2.tsx
```diff
@@ -0,0 +1,389 @@
...
+    container: {
+      padding: 0,
+    },
+    row: {
+      padding: 0,
+    },
+    col: {
+      padding: 0,
+    },
+  },
```
dzucconi, line 152: I was thinking we could avoid using this in V3 trees, but am remembering now that the artwork grids use it. Wondering how much work it would take to drop this library. I suspect it wouldn't be too bad?

packages/palette/package.json
```diff
@@ -71,7 +71,7 @@
     "@types/react-lazy-load-image-component": "1.3.0",
     "@types/react-test-renderer": "16.8.1",
     "@types/semver": "5.5.0",
-    "@types/styled-components": "4.0.2",
+    "@types/styled-components": "4.0.3",
```
dzucconi, line 74: Looks like I didn't get this right; trying to find a version that has ThemeContext but doesn't catastrophically break all our typings (which some of the 4x releases do). Difficult to hunt down the specific version that `ThemeContext` was introduced with; could always `@ts-ignore` it.

packages/palette/src/themes/Themes.story.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </Column>
+      </GridColumns>
+
+      <Separator color="black30" my={6} />
+
+      {colors.map(([name, value]) => {
+        return (
+          <GridColumns key={name} justifyContent="space-between" my={4}>
+            <Column span={6} display="flex" alignItems="center">
+              <Box width={60} height={60} bg={name} borderRadius="50%" />
```
damassi, line 73: do we not have a 6 spacing scale?

dzucconi: width/height doesn't use the spacing scale

packages/palette/src/themes/v2.tsx
damassi: I believe our every-day artwork grids use some kind of mad science, not a formal grid.

The API between the new `GridColumns` component and the old grid is close enough to where i think we could swap it out and remove the old one, but we're looking at ~41 files minimum. Not sure about volt, etc:

damassi:
@dzucconi 

> The main question I'm going back and forth with is: should you be able to use V2 tokens and variants in a V3 context tree if they happen to not overlap with the V3 tokens/variants. We could, for instance, merge the v2 theme into the v3 theme, favoring v3 keys/values.

This was the original idea based on backwards compatibility and to keep things manageable, but I'm open for argument.

If we scope things to just V3, what would an upgrade path for a given app look like? Can you be as detailed as possible? I'm curious about happy path versus sad path options because ideally we don't find ourselves mired. 

If we were to merge, my thinking is that we don't need to do it blindly, but rather encode a migration path where if a condition is set then we get warnings and utilize a proxy (or something) to return metadata around use. For example:

```tsx
function getV2Theme() {
  const theme = {
    foo: "hello",
    bar: "hi"
  }
  
  if (!__THEME_MIGRATION__) {
    return theme 
  }
  
  const themeHandler = {
    get: function(target, prop, receiver) {
       console.warn(...) 
       return ...
    }
  }

  const themeProxy = new Proxy(theme, themeHandler);
  return themeProxy
}

<ThemeProvider theme={{
  ...getV2Theme(),
  ...v3Theme
}}>
  ...
</ThemeProvider>
```

Of course, this could all be overkill, but proxy's are helpful in this way since we can intercept lookup.

packages/palette/src/themes/Themes.story.tsx
damassi: oh duh

packages/palette/package.json
damassi: How are we still on this old styled-components version 🤯 Lets return to this first thing after we get this new version in.

dzucconi:
> This was the original idea based on backwards compatibility and to keep things manageable, but I'm open for argument.

Well this still more or less works that way; just ensures that subtrees wrapped with the v3 theme are only v3.

So you could for instance have v2 and v3 on the same page though:

```tsx
<ThemeV2>
  <Text variant="text">Some older code</Text>
  <Button variant="primaryBlack">Button that looks older</Button>

  {/*. Later... */}

  <ThemeV3>
    <Text variant="sm">Newer code</Text>
    <Button variant="primaryBlack">Newer button style</Button>
  </ThemeV3>
</ThemeV2>
```

The question isn't so much as to whether or not we should be backwards compatible — the `ThemeProviders` accommodate that, IMO. But rather should we ensure that if you're inside of a `ThemeV3` provider you can't use any v2 theme tokens or variants unless you're explicitly wrapping them in a `ThemeV2` provider.

If I backfill the v3 theme with v2 tokens then it just becomes more confusing as to what's sanctioned within v3 from the consumers viewpoint, even with a console warning (which the state of warnings/errors in Force would just be noise we ignore). Imagine if within the `ThemeV3` you could still use the `3` spacing token; but then somewhere down the line we want to actually get rid of the v2 theme in Palette. It'd be extremely difficult to hunt down these inconsistencies, particularly because they wouldn't error — 3 is valid as... 3px now all of a sudden. Whereas if they were all explicitly wrapped with a `ThemeV2` provider they'd be very easy to locate and migrate.

damassi:
>  Whereas if they were all explicitly wrapped with a ThemeV2 provider they'd be very easy to locate and migrate

So the idea would be [to take `<Theme>` out of  Boot](https://github.com/artsy/force/blob/8c760bd102db4ab882707164a679fd737e4bed8e/src/v2/Artsy/Router/Boot.tsx#L66) and decorate it where needed? Or swap the boot theme with V3 and work backwards from there?

dzucconi:
> So the idea would be to take <Theme> out of Boot and decorate it where needed? Or swap the boot theme with V3 and work backwards from there?

Yeah — either leave <Theme> alone with v2 as the default and progressively migrate pages to v3; swapping the default to v3 once some critical mass of pages is migrated. Or swap the default to v3 and decorate all old pages with v2 and migrate them that way. Either way you're explicitly making a decision about which theme to opt into.

It could also just be that this is the kind of thing that gets figured out in practice when we try to put together the kitchen sink page/start to migrate some older pages. I really can go either way with this; but am kind of just leaning towards the more strict path.

damassi:
Just so we're aligned on what strict means here:

> strict path

Boot gets V3 default, we make explicit our opting in to V2 at the root of our app trees -- yah? 

But agree, I think the proper path will shake out once we start on the molecule stuff.

dzucconi:
Strict path = we don't backfill tokens in the v3 theme. You can only use v3 tokens and variants within the v3 provider and vice versa for v2. So pruned space values no longer work, etc.

Loose path = we backfill tokens in the v3 theme. You can use all variants (text variants for instance) within v3.

In both cases v2 stays the same. The difference is how v3 behaves: can you mix together old stuff with new stuff in the same theme provider?

In either case we're going to have to choose one way to opt in at the root provider.

packages/palette/src/themes/v3.tsx
```diff
@@ -42,54 +39,54 @@ const BREAKPOINTS_SCALE = Object.assign(
  * design system from our design team
  */
 export const THEME = {
-  id: "v3",
+  // Backfill with v2
+  ...THEME_V2,
```
damassi, line 43: Aw no, what happened?

dzucconi: I'm gonna try rolling this back; I thought I was having a type issue with the theme but it turned out do to me using `as const` on the theme to convert them to readonly (which let me do a typeguard on id).

dzucconi:
OK. Finally have this passing in Force. There should be no functional/visual changes in v2, and now there's the ability to run v3 in Force. Going to look over all the UI specs then merge.

cc @damassi

damassi:
Hell yahh 🎉

PR #869, opened 2021-03-04
[v3] Banner/Message

dzucconi:
So: I made some breaking changes here in order to improve/restrict the API.
* Removed `textSize` in Message because you can just use custom children if you really need to
* Removed `textColor` and `backgroundColor` because we don't want or need that.

There are a few changes to Volt and Force. I'm going to probably be doing this a few more times: cutting a breaking release and then updating Volt/Force. If you think I should include some other apps in this cycle then point them out; but those seem like the main to be concerned with.

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@13.36.2-canary.869.15447.0
  # or 
  yarn add @artsy/palette@13.36.2-canary.869.15447.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Banner/Banner.test.tsx
+++ b/packages/palette/src/elements/Banner/Banner.test.tsx
@@ -1,41 +1,35 @@
 import { mount } from "enzyme"
 import "jest-styled-components"
 import React from "react"
+import { Theme } from "../../Theme"
 import { Banner } from "../Banner"
 
 describe("Button", () => {
-  it("pass a custom color to Banner", () => {
-    const message = "There was an error."
+  it("has default black10 background to Banner", () => {
     const wrapper = mount(
-      <Banner message={message} backgroundColor="purple100" />
+      <Theme>
+        <Banner />
+      </Theme>
     )
-    expect(wrapper.debug()).toContain("purple100")
-  })
-
-  it("has default red background to Banner", () => {
-    const message = "There was an error."
-    const wrapper = mount(<Banner message={message} />)
-    expect(wrapper.debug()).toContain("red100")
+    expect(wrapper).toHaveStyleRule("background-color", "#E5E5E5")
   })
 
   it("displays the message", () => {
     const message = "There was an error."
-    const wrapper = mount(<Banner message={message} />)
+    const wrapper = mount(<Banner>There was an error.</Banner>)
     expect(wrapper.text()).toEqual(message)
-    expect(wrapper.find("CloseButton")).toHaveLength(0)
+    expect(wrapper.find("Clickable")).toHaveLength(0)
   })
 
   it("is dismissable", () => {
-    const message = "There was an error."
-    const wrapper = mount(<Banner dismissable message={message} />)
-    expect(wrapper.find("CloseButton")).toHaveLength(1)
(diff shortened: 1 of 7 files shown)


packages/palette/src/themes/v2.tsx
```diff
@@ -93,6 +93,8 @@ export const THEME = {
     green10: "#E9F2EC",
     /** Full purple, secondary brand color. Should only used in time/transitions (on hover, active state), for highlighting vital text, and links.   */
     purple100: "#6E1EFF",
+    /** Alias of purple100 */
+    brand: "#6E1EFF",
```
dzucconi, line 97: Added this alias; which seems like it'd be useful

packages/palette/src/elements/Banner/Banner.test.tsx
```diff
@@ -1,41 +1,35 @@
...
-  it("pass a custom color to Banner", () => {
-    const message = "There was an error."
-    const wrapper = mount(
-      <Banner message={message} backgroundColor="purple100" />
-    )
-    expect(wrapper.debug()).toContain("purple100")
-  })
+const mount: typeof __mount__ = (children) => {
+  return __mount__(<Theme>{children}</Theme>)
+}
```
damassi, line 9: If this is gonna soon be everywhere might be worth turning it into something we can import

packages/palette/src/elements/Banner/Banner.tsx
```diff
@@ -1,79 +1,81 @@
...
+  },
+  error: {
+    backgroundColor: "red100",
+    color: "white100",
+  },
+  brand: {
+    backgroundColor: "brand",
+    color: "white100",
+  },
+}
```
damassi, line 31: Doing it right 👍

damassi (approved):
One suggestion; otherwise LGTM 👍

packages/palette/src/elements/Banner/Banner.test.tsx
dzucconi: Yeah; was trying to think if there's a simpler way around this. I'll come up with something 👍

packages/palette/src/Theme.tsx
```diff
@@ -49,6 +49,12 @@ export const useTheme = <T extends TTheme>() => {
   return { theme }
 }
 
+/** Returns a config specific to the current theme */
+export const useThemeConfig = <T, U>({ v2, v3 }: { v2: T; v3: U }) => {
+  const { theme = { id: "v2" } } = useTheme()
```
dzucconi, line 54: Setting a default here now so in most cases you won't have to wrap in a theme. The only reason `Button` requires it right now is because of that hover media query. Added a `mountWithTheme` function as well anyway.

damassi: > The only reason Button requires it right now is because of that hover media query.

Maybe we can make it so this isn't required? Can we skip the mq and do it a different way?

dzucconi: I actually just am PRing that separately now because it's gonna cause problems with Force specs 😬

artsyit:
:rocket: PR was released in [`@artsy/palette@14.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.0.0) :rocket:

PR #870, opened 2021-03-04
Avoids failing specs on unthemed components

dzucconi:
This should get the Force integration PRs (https://github.com/artsy/force/pull/7115) to pass

### DIFF
--- a/packages/palette/src/Theme.tsx
+++ b/packages/palette/src/Theme.tsx
@@ -49,6 +49,12 @@ export const useTheme = <T extends TTheme>() => {
   return { theme }
 }
 
+/** Returns a config specific to the current theme */
+export const useThemeConfig = <T, U>({ v2, v3 }: { v2: T; v3: U }): U | T => {
+  const { theme = { id: "v2" } } = useTheme()
+  return theme.id === "v2" ? v2 : v3
+}
+
 /** Typeguard for v2 */
 export const isThemeV2 = (theme: TTheme): theme is ThemeV2 => {
   return theme.id === "v2"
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/Button/v3/Button.tsx
```diff
@@ -116,7 +117,7 @@ const Container = styled.button<ContainerProps>`
   ${(props) => {
     // Handle pseudo classes
     return css`
-      @media ${props.theme.mediaQueries.hover} {
+      @media ${MEDIA_QUERY_HOVER} {
```
dzucconi, line 120: I guess actually I could just import this from each respective theme. I had thought this was weird initially when they were the same component.

packages/palette/src/Theme.tsx
```diff
@@ -49,6 +49,12 @@ export const useTheme = <T extends TTheme>() => {
   return { theme }
 }
 
+/** Returns a config specific to the current theme */
+export const useThemeConfig = <T, U>({ v2, v3 }: { v2: T; v3: U }) => {
+  const { theme = { id: "v2" } } = useTheme()
+  return { v2, v3 }[theme.id as "v2" | "v3"]
```
damassi, line 55: #nit Can we make this form a little more conventional

dzucconi: WHAT HAPPENS WHEN WE HAVE THREE OR FOUR THEMES?

(I kid, will make it a ternary)

artsyit:
:rocket: PR was released in [`@artsy/palette@13.36.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4013.36.1) :rocket:

PR #872, opened 2021-03-05
[v3] Pill

dzucconi:
This PR implements the new Pill component. So there's [no corresponding v2 styling](https://static.damonzucconi.com/_capture/lOiUBruu.png) (We can add it if need be later).


packages/palette/src/elements/Button/v3/Button.tsx
```diff
@@ -77,21 +77,19 @@ const Container = styled.button<ContainerProps>`
...
-    return variant({ prop: "size", variants: BUTTON_SIZES })(props)
-  }};
+  /* Handle sizing */
+  ${variant({ prop: "size", variants: BUTTON_SIZES })}
 
   ${(props) => {
     // Handle props driven states
     if (props.hover) {
-      return variant({ variants: BUTTON_VARIANTS.hover })(props)
+      return variant({ variants: BUTTON_VARIANTS.hover })
```
dzucconi, line 86: Accidentally realized you don't have to thread through the props for these variant calls. Likely for situations exactly like this.

packages/palette/src/elements/Pill/Pill.tsx
```diff
@@ -0,0 +1,155 @@
...
+const Thumbnail = styled.img`
+  display: block;
+  border-radius: 50%;
+  width: 30px;
+  height: 30px;
+  margin-right: ${themeGet("space.1")};
+  background-color: ${themeGet("colors.black30")};
+`
+
+/** A Pill is a non-cta button that's  */
```
dzucconi, line 125: Woops forgot to finish this comment

packages/palette/src/elements/Pill/Pill.story.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export const LinkExample = () => {
+  return (
+    <Pill
+      variant="artist"
+      as="a"
+      // TODO: Need a decent way of typing the threaded polymorphic `as` prop
+      // @ts-ignore
+      href="#example"
```
dzucconi, line 89: The type magic that @types/styled-components does for this is a bit over my head. Don't know how to make it work when threading the `as` prop through. Haven't been able to find any real solutions other than explicitly typing the union.

packages/palette/src/elements/Button/v3/tokens.ts
```diff
@@ -1,3 +1,4 @@
+import { DROP_SHADOW } from "../../../helpers/shadow"
```
damassi, line 1: 👍

packages/palette/src/elements/Pill/Pill.tsx
```diff
@@ -0,0 +1,159 @@
...
+        active?: boolean
+      }
+    | {
+        /** `"artist"` pills do not have an `active` state */
+        variant?: Extract<PillVariant, "artist">
+        /**
+         * Optional avatar; [1x] or [1x, 2x]
+         * Should target 30x30 @1x, 60x60 @2x
+         */
+        srcs?: [string] | [string, string]
```
damassi, line 82: Thoughts on having a little less friction here?
```
src: string | [string, string]
```
Then one can pass a simple string or an src set, and internally can handle it below.

packages/palette/src/elements/Pill/Pill.tsx
```diff
@@ -0,0 +1,159 @@
...
+  transition: color 0.25s ease, border-color 0.25s ease,
+    background-color 0.25s ease, box-shadow 0.25s ease;
+
+  ${variant({ variants: PILL_VARIANTS })}
+
+  ${(props) => {
+    return css`
+      ${props.hover && STATES.hover}
+      ${props.focus && STATES.focus}
+      ${"active" in props && props.active && STATES.active}
```
damassi, line 102: why not just `props.active && STATES.active`?

damassi (approved):
A couple tiny comments, but nothing blocking. Looks great 👍

packages/palette/src/elements/Pill/Pill.tsx
dzucconi: `variant="artist"` can't have an active state so `key in object` is used for the type-guard.

packages/palette/src/elements/Pill/Pill.tsx
dzucconi: Yeah, good idea

artsyit:
:rocket: PR was released in [`@artsy/palette@14.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.1.0) :rocket:

PR #874, opened 2021-03-08
chore(stories): adds tabs/stepper stories

dzucconi:
Just checking in these stories (in particular Stepper's) so I have the visual diff to work with when I PR later today.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.2.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.2.1) :rocket:

PR #875, opened 2021-03-09
refactor(tabs,stepper): refactors to utilize basetabs

dzucconi:
I've been meaning to do this. Was pretty annoying.

I don't think this warrants a breaking change necessarily though I did remove the `transformTabBtn` prop in favor of just a custom `Tab` component prop; which is easier to understand and the same pattern that `Carousel` and `Swiper` use. `Stepper` appears to be the only thing that is using this directly. I think we probably could get around this a few other ways at the expense of breaking the Tab API in favor of a more modular approach using a context provider, etc... but seems unnecessary for now.

I also removed the code for `autoScroll` since it looks like it's been broken for a while, presumably when React refs changed a while back (the bug is that the ref is un-forwarded). Also it's weird UX anyway, so.

<del>I'm not expecting any diffs in the visual specs, let's see how I did...</del> OK...

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@14.2.1-canary.875.15683.0
  # or 
  yarn add @artsy/palette@14.2.1-canary.875.15683.0
  ```
</details>


dzucconi:
Ah, I think the tab border line is 1px off; which of course means there's a slight diff on literally every spec now. 😬

dzucconi:
There's a few exports here that I need to check for in Force/Volt; I *think* they were maybe being used for tests. If that's the case I'll mark this as a major release and pr each.

packages/palette/src/elements/BaseTabs/BaseTab.tsx
```diff
@@ -8,15 +8,17 @@ import { Text, TextProps } from "../Text"
...
     // When displaying as an anchor tag, accept anchor props:
-    | ({ as: "a" } & (React.AnchorHTMLAttributes<HTMLAnchorElement>))
+    | ({ as: "a" } & React.AnchorHTMLAttributes<HTMLAnchorElement>)
     // Cop-out: if you pass as with a complex component, loosen up the types:
-    | ({ as: React.ReactElement<unknown> } & Record<string, unknown>))
+    | ({ as: React.ReactElement<unknown> } & Record<string, unknown>)
+  )
 
-const BaseTabActiveMixin = css`
+const baseTabActiveMixin = css`
```
damassi, line 21: 🙏

packages/palette/src/elements/Tabs/Tabs.tsx
```diff
@@ -1,278 +1,87 @@
...
-import { Flex, FlexProps } from "../Flex"
-import { Join } from "../Join"
-import { Sans } from "../Typography"
-
-import { color, media } from "../../helpers"
-import { Text } from "../Text"
+import React, { useState } from "react"
+import { flattenChildren } from "../../helpers/flattenChildren"
+import { useUpdateEffect } from "../../utils/useUpdateEffect"
+import { BaseTabs, BaseTabsProps } from "../BaseTabs"
```
damassi, line 4: Necessary to have a BaseTabs as well as a Tabs?

packages/palette/src/elements/Tabs/Tabs.tsx
```diff
@@ -1,278 +1,87 @@
...
+import { TabProps } from "./Tab"
+import { TabsTab, TabsTabProps } from "./TabsTab"
```
damassi, line 6: And necessary to have a `TabsTab` as well as a `Tab`?

damassi: If all of this division is strictly necessary then 👍 but it be a little more straight forward if there was a `Tabs` component, and inside it was a `Tabs` and `Tab`

packages/palette/src/elements/Tabs/Tabs.tsx
dzucconi: Yeah; though the naming can be improved. Most of the tabs in Force are BaseTabs.  They just have no behavior associated with them (so they are used as visual links basically). The Tabs are client-side tabs.

There's this distinction in the Figma docs between NavigationalTabs and ContentTabs which we can maybe draw the lines more clearly with.

packages/palette/src/elements/Tabs/Tabs.tsx
dzucconi: But yeah; TabsTab doesn't need to be exported from the index.

dzucconi:
Going to remove <Sup> and support count and then just cut a major release since that'll break in Volt.

packages/palette/src/elements/Tabs/Tabs.tsx
damassi: Of course, I don't mean to dismiss this work! I suspect that we're both on the same page about where this has been and is going...

dzucconi: > I don't understand why this component is so complex given the tiny API.

Two reasons:

* Shovelling props into <Tabs> using the <Tab> component. This is where you get the problem with the two different tabs, one visual/clickable; the other just part of the Tabs API.
* `Stepper` — it's close enough to Tabs that Tabs were used to build it but it requires disabling and altering the tab buttons, which is *weird*.

dzucconi:
@damassi as always thanks for your feedback. Take a look at 92198ac340e30fbc8de4cc75a5b56c66d74d1415 — also weird — instead of injecting the Tab component, just extract the tabbing logic into a hook. But I can at least get rid of the `TabsTab` and separate files.

damassi:
>  also weird

Nah this is way better! 👍 👍

dzucconi:
OK... https://github.com/artsy/force/pull/7134 will replace with the version when it's released; minor snapshot updates.

Volt is annoying: a bunch of specs are reaching into the Tab internals. Falling behind a bit so I'm punting on Volt and will come back to that.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.2.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.2.1) :rocket:

damassi:
@dzucconi - relying on component internals for tests is a pretty bad practice so i wouldn't hesitate to pend those specs and / or delete a few over zealous assertions.

PR #876, opened 2021-03-10
[v3] Tabs, Stepper

dzucconi:
localhost_6006_iframe html_id=theme--controls viewMode=story

### DIFF
--- a/packages/palette/src/Theme.tsx
+++ b/packages/palette/src/Theme.tsx
@@ -49,9 +49,18 @@ export const useTheme = <T extends TTheme>() => {
   return { theme }
 }
 
-/** Returns a config specific to the current theme */
+/** Returns a config specific to the current theme. For use in React components */
 export const useThemeConfig = <T, U>({ v2, v3 }: { v2: T; v3: U }): U | T => {
   const { theme = { id: "v2" } } = useTheme()
+  return theme.id === "v2" ? v2 : v3
+}
+
+/** Returns a config specific to the current theme. For use in styled-components */
+export const getThemeConfig = <T, U>(
+  props: Record<string, any>,
+  { v2, v3 }: { v2: T; v3: U }
+): U | T => {
+  const { theme = { id: "v2" } } = props
   return theme.id === "v2" ? v2 : v3
 }
 
--- a/packages/palette/src/elements/BaseTabs/BaseTab.tsx
+++ b/packages/palette/src/elements/BaseTabs/BaseTab.tsx
@@ -1,28 +1,26 @@
 import React from "react"
 import styled, { css } from "styled-components"
-import { color } from "../../helpers"
+import { getThemeConfig } from "../../Theme"
+import { Clickable, ClickableProps } from "../Clickable"
 import { Text, TextProps } from "../Text"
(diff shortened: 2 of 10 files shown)


packages/palette/src/elements/BaseTabs/BaseTab.tsx
```diff
@@ -32,31 +30,47 @@ const baseTabActiveMixin = css`
...
 }
 
 BaseTab.displayName = "BaseTab"
+
+/**
+ * So this is interesting: We can trick styled-components into correcting the CSS specificity (Text > Clickable)
+ * by simply including a Clickable here.
+ */
+// @ts-ignore
+const __Ignore__ = Clickable
```
dzucconi, line 76: I'm not 100% sure of the why/how of this. I'm assuming it's because of import order — though having this reference here is also important...

packages/palette/src/Theme.tsx
```diff
@@ -49,12 +49,21 @@ export const useTheme = <T extends TTheme>() => {
...
 
+/** Returns a config specific to the current theme. For use in styled-components */
+export const getThemeConfig = <T, U>(
+  props: Record<string, any>,
+  { v2, v3 }: { v2: T; v3: U }
+): U | T => {
+  const { theme = { id: "v2" } } = props
+  return theme.id === "v2" ? v2 : v3
+}
+
```
damassi, line 66: We can probably collapse this duplicate function into `useThemeConfig`, where an additional optional arg takes `props`, pushing it forward in precedence

packages/palette/src/elements/Stepper/Stepper.tsx
```diff
@@ -20,37 +22,78 @@ export const Stepper: React.FC<StepperProps> = ({
...
+    v3: {
+      fill: true,
+      joinSeparator: null,
+      inlineSeparator: <ChevronIcon fill="black60" height={10} ml={1} />,
+      verticalAlignment: "center",
+      horizontalAlignment: "flex-start",
+      checkAlignment: "right",
+      textVariant: "sm" as TextVariant,
+    },
+  })
```
damassi, line 47: Any reason this is inline with the coponent versus in its own `/tokens/v2|v3` file like the others?

damassi (approved):
Couple non-blocking qs, 👌

packages/palette/src/Theme.tsx
dzucconi: Since useTheme is a hook it both needs to be in a React component and can't be called conditionally. So I'm not sure that'd be possible.

packages/palette/src/elements/Stepper/Stepper.tsx
dzucconi: It's not the most compelling reason but I've just been doing that when they are CSS styles or actual variants which both get kind of long. All the ones that are like switches in values like this I've just been inlining since there tends to only be one or two. This one kinda just grew as I noticed some more differences between the two themes.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.3.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.3.0) :rocket:

PR #877, opened 2021-03-11
[v3] Sup, ProgressBar, CarouselBar, Progress dashes

dzucconi:


packages/palette/src/elements/ProgressBar/ProgressBar.tsx
```diff
@@ -1,39 +1,44 @@
...
+
   return (
-    <ProgressBarBackground showBackground={showBackground}>
-      <div
+    <Box
+      height="2px"
+      position="relative"
+      overflow="hidden"
+      mt={0.5}
+      mb={1}
```
dzucconi, line 29: These default margins were here and I doubt it's worth it to try to get rid of them for now. They get reset to zero in the CarouselBar variant though.

packages/palette/src/elements/ProgressDots/ProgressDots.tsx
```diff
@@ -34,19 +37,28 @@ export interface ProgressDotsProps extends BoxProps {
...
+  const bgColor = useThemeConfig({ v2: "black10", v3: "black30" })
+
   return (
     <>
       <Box
         role="presentation"
         display="flex"
         alignItems="center"
         justifyContent="center"
+        my={0.5}
```
dzucconi, line 52: This one is my fault 😞

packages/palette/src/elements/Sup/Sup.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components"
+import { Text, TextProps } from "../Text"
+
+export type SupProps = TextProps
+
+export const Sup = styled(Text)`
+  display: inline;
+  vertical-align: super;
+  font-size: 75% !important;
```
dzucconi, line 9: Conveniently the new superscript styles are specified as 75% of the current font size. I'm just replacing the old Sup tag with this. It'll be a bit different but not by much.

packages/palette/src/elements/Sup/Sup.tsx
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components"
+import { Text, TextProps } from "../Text"
+
+export type SupProps = TextProps
+
+export const Sup = styled(Text)`
```
dzucconi, line 6: Doesn't actually need to be a `<sup>` tag since we don't actually need the semantics of superscript. Regarding the role (not tag) superscript, WAI-ARIA says:

> The superscript role is intended to be used only to mark up typographical conventions that have specific meanings; not for typographical presentation for presentation's sake. In general, authors SHOULD use this role only if the absence of the superscript would change the meaning of the content.

packages/palette/src/elements/ProgressBar/ProgressBar.tsx
```diff
@@ -1,39 +1,44 @@
...
+    >
+      <Box
+        bg={highlight}
+        width="100%"
+        height="100%"
         style={{
           transition,
-          backgroundColor: color(highlight as any),
-          width: Math.round(percentComplete) + "%",
+          transform: `translateX(-${100 - percentComplete}%)`,
```
dzucconi, line 39: I changed this to animate based on transform instead of width because transform and opacity are the only GPU accelerated animations. I doubt it makes that much of a difference in this case, but why not.

I only see [one example in Force](https://github.com/artsy/force/blob/953d162ee90f7a2223d34ddf902a1c888f0bd2f8/src/v2/Apps/ViewingRoom/Components/ViewingRoomCarousel.tsx#L97) where I'll have to update a property. Not a huge deal. Should maybe have this accept a custom transitionDuration instead of the whole property?

packages/palette/src/elements/Tabs/Tabs.story.tsx
```diff
@@ -57,3 +57,32 @@ export const WithData = () => {
...
+      />
+      <Tab name="CV" />
+      <Tab
+        name={
+          <>
+            Auction results <Sup>123</Sup>
+          </>
+        }
+      />
+    </Tabs>
```
dzucconi, line 86: This is clunky, but it looks like we're going to need these superscript counts in other places as well aside from tabs. So rather than hardcode a `count` prop, I think this is fine.

packages/palette/src/elements/ProgressBar/ProgressBar.tsx
damassi: Fine either way

damassi (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette@14.4.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.4.0) :rocket:

PR #878, opened 2021-03-12
[v3] Select

dzucconi:
On the V2 instances I got rid of the serif text in favor of a reasonable sans variant. Should be OK, works the same obviously. Also consolidated them into a single implementation then deprecated LargeSelect / SelectSmall.

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -1,36 +1,14 @@
 import { themeGet } from "@styled-system/theme-get"
-import React, { SFC } from "react"
+import React from "react"
 import styled, { css } from "styled-components"
-import {
-  maxWidth,
-  MaxWidthProps,
-  minWidth,
-  MinWidthProps,
-  PositionProps,
-  space as styledSpace,
-  SpaceProps,
-  width,
-  WidthProps,
-} from "styled-system"
-import { color, space } from "../../helpers"
-import { Box } from "../Box"
-import { computeBorderColor, Required } from "../Input"
-import { Sans, Serif } from "../Typography"
-
-const computeOptionTags = (options: Option[], name?: string): JSX.Element[] => {
-  const optionTags = options.map((option, index) => {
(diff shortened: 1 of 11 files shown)


packages/palette/src/elements/Select/Select.tsx
```diff
@@ -1,254 +1,246 @@
...
+        </Text>
       )}
     </Box>
   )
 }
 
-/**
- * A small version of drop-down select menu
- */
-export const SelectSmall: SFC<SelectProps> = props => {
```
damassi, line 109: Will want to double check that we're not importing this directly in different places

dzucconi: It will still work: at the bottom of the file you'll encounter where I curry them with variant

damassi (approved):
Looks great 👍 I don't see any problems here but we'll always want to be sure that the next dev that comes along will be able to make changes, without wondering if they're in v2 or v3. Its always acceptable to duplicate and delete, versus intertwine, if that makes sense.

dzucconi:
> Its always acceptable to duplicate and delete, versus intertwine, if that makes sense.

It does, but for most things it doesn't seem worth it. In every component with differences there's some theme-specific CSS and/or some theme-specific tokens. The differences are relatively minor, strictly visual, and they are always contained inside of a `useThemeConfig` or `getThemeConfig` call.

In scenario one, we end up having to support these for a while and I'd rather have to deal with changes in a single place vs having to make it in both. There's also the problem of just potentially doubling the site of Palette's component code in the worst case.

In scenario two, we have a relatively speedy transition and you just have to unwind a few lines of assignment in most cases to delete the old code.

But yeah, I'm open to just blanket refactoring them one way or the other if this proves to be troublesome.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.5.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.5.0) :rocket:

PR #883, opened 2021-03-18
Deprecates Sans/Serif; mixes in textTransform to Text

dzucconi:
The deprecation was long overdue. :)

### DIFF
--- a/packages/palette/src/elements/Text/Text.tsx
+++ b/packages/palette/src/elements/Text/Text.tsx
@@ -6,6 +6,7 @@ import {
   compose,
   ResponsiveValue,
   style,
+  system,
   typography,
   TypographyProps,
   variant,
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Text/Text.tsx
```diff
@@ -29,8 +30,18 @@ const textColor = style({
   key: "colors",
 })
 
+type TextTransform =
+  | "none"
+  | "capitalize"
+  | "uppercase"
+  | "lowercase"
+  | "initial"
+  | "inherit"
```
damassi, line 39: Could this be pulled from general dom types? `CSS.Something`

dzucconi: I can if I pull in https://github.com/frenic/csstype — it's a dependency of styled-system anyway 🤷

damassi: Oh i see, i was thinking that was a [part of this](https://github.com/artsy/force/blob/master/tsconfig.json#L23)

artsyit:
:rocket: PR was released in [`@artsy/palette@14.7.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.7.0) :rocket:

PR #884, opened 2021-03-19
feat: add optional onClick prop to Toggle

mdole:
You can actually already use `<Toggle onClick={...}>`, but you'll get a type error. This PR adds it to `Toggle`'s interface. Current anticipated use case is adding a tracking event to a toggle in Volt so we can see when users interact with it.

Seem reasonable to folks?

### DIFF
--- a/packages/palette-docs/content/docs/elements/layout/Toggle.mdx
+++ b/packages/palette-docs/content/docs/elements/layout/Toggle.mdx
@@ -78,3 +78,14 @@ name: Toggle
     <Toggle label="Time period" />
   </>
 </Playground>
+
+<Playground title="With optional onClick">
+  <Toggle label="Click me and check your console!" expanded onClick={(e) => console.log(e)}>
+    <Sans size="3">
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+      tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
+      veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
+      commodo consequat.
+    </Sans>
+  </Toggle>
+</Playground>
--- a/packages/palette/src/elements/Toggle/Toggle.tsx
+++ b/packages/palette/src/elements/Toggle/Toggle.tsx
@@ -19,6 +19,7 @@ export interface ToggleProps extends FlexProps {
   renderSecondaryAction?: (
     toggleProps: Pick<ToggleProps, "disabled" | "expanded" | "textSize">
   ) => JSX.Element
+  onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
 }
 
 export interface ToggleState {
@@ -35,6 +36,7 @@ export const Toggle: React.FC<ToggleProps> = ({
   disabled,
   children,
   renderSecondaryAction,
+  onClick,
   ...rest
 }) => {
   const [expanded, setExpanded] = useState(defaultExpanded)
@@ -48,7 +50,10 @@ export const Toggle: React.FC<ToggleProps> = ({
   return (
     <Flex width="100%" flexDirection="column" pb={2} {...rest}>
       <Header
-        onClick={toggleExpand}
+        onClick={(e) => {
+          toggleExpand()
+          !(onClick === undefined) && onClick(e)
+        }}
         disabled={disabled}
         borderTop="1px solid"
         borderColor="black10"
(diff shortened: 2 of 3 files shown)


packages/palette/src/elements/Toggle/Toggle.tsx
```diff
@@ -19,6 +19,11 @@ export interface ToggleProps extends FlexProps {
   renderSecondaryAction?: (
     toggleProps: Pick<ToggleProps, "disabled" | "expanded" | "textSize">
   ) => JSX.Element
+  /**
+   * Optional onClick prop for anything that should fire in addition to toggling
+   * open/closed, e.g. tracking calls
```
annacarey, line 24: What do you mean by tracking calls here? Like tracking num of clicks?

annacarey (commented):
Added a non-blocking question. I am finding that comment to be a teeny confusing but I think it's because I don't know what you mean by tracking calls.

Do you think something like the following is more clear?
"Optional onClick prop to add additional behavior to a click action (beyond the built-in toggling open/closed behavior)"

packages/palette/src/elements/Toggle/Toggle.tsx
mdole: Oop yeah good question. I meant "tracking event calls", like the events we send to Segment. I like your version!

dzucconi (commented):
This is fine for now; but just a heads up: we're gonna just refactor this whole thing have it use a real button / pass through the native click event instead of this manual annotation. Like typically we wouldn't need or want to have to manually add something like onClick to this, which is a given.

packages/palette-docs/content/docs/elements/layout/Toggle.mdx
```diff
@@ -78,3 +78,14 @@ name: Toggle
     <Toggle label="Time period" />
   </>
 </Playground>
+
+<Playground title="With optional onClick">
+  <Toggle label="Click me and check your console!" expanded onClick={() => console.log("Hi there!")}>
+    <Sans size="3">
```
dzucconi, line 84: I'm sure you know but it's just weird seeing it in new code. `Sans` is long dead.

Also just so it's not surprising: this playground site isn't currently deploying.

dzucconi (changes requested):
Actually; little problem being the onClick handler is being passed to the wrong component: pull it out; pass it to the Clickable and annotate the event with the correct type.

packages/palette/src/elements/Toggle/Toggle.tsx
```diff
@@ -19,6 +19,11 @@ export interface ToggleProps extends FlexProps {
   renderSecondaryAction?: (
     toggleProps: Pick<ToggleProps, "disabled" | "expanded" | "textSize">
   ) => JSX.Element
+  /**
+   * Optional onClick prop to add additional behavior to a click action
+   * (beyond the built-in toggling open/closed behavior)
+   */
```
dzucconi, line 25: I don't really find this comment useful. More confusing/noise than anything?

packages/palette-docs/content/docs/elements/layout/Toggle.mdx
mdole: Oh interesting - the playground site isn't deploying intentionally? I know we're talking about moving to Storybooks, is it just pending that migration? Not a huge deal obviously, just curious.

dzucconi: Not intentionally, just more like: it broke during the Storybook upgrade and no one has looked at in in a while. I've never used Gatsby; some dependencies thing going on there.

Definitely interested in de-duping the examples documentation but whether or not we do something import stories into a the docs site or what, I'm not sure.

dzucconi: But in general I find the Storybook stories much more useful because they are consistent and have visual specs associated with them.

packages/palette/src/elements/Toggle/Toggle.tsx
```diff
@@ -19,6 +19,7 @@ export interface ToggleProps extends FlexProps {
   renderSecondaryAction?: (
     toggleProps: Pick<ToggleProps, "disabled" | "expanded" | "textSize">
   ) => JSX.Element
+  onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```
dzucconi, line 22: 👍

dzucconi (approved):
Thanks!

artsyit:
:rocket: PR was released in [`@artsy/palette@14.7.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.7.3) :rocket:

PR #885, opened 2021-03-19
fix(inputs): sets font-family to sans

dzucconi:
Would anyone oppose me changing the global font-family to courier or something obvious inside of storybooks so that we can be sure to catch things like this in the future?

The default font-family in most of Force is still, regrettably, Garamond.


damassi:
Not opposed! Kinda clever

artsyit:
:rocket: PR was released in [`@artsy/palette@14.7.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.7.1) :rocket:

PR #886, opened 2021-03-19
fix(divider): resets default styling when presenting as hr

dzucconi:
Looks like Chrome updated the default style sheet for HRs which is causing some Separators which present as hr to look greenish.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.7.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.7.2) :rocket:

PR #892, opened 2021-03-24
fix(labeledinput): allows for clickable labels

dzucconi:
Currently have the situation where I'd like to use this label prop to insert a clear button for a filter input. I think it makes sense for textual labels to have pointer-events none and then leave it alone for custom components.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.11.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.11.1) :rocket:

PR #894, opened 2021-03-24
fix(labeledinput): corrects tabindex

dzucconi:
OK last one 😉 

Fixes the tabindex by just moving the label into a natural order.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.12.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.12.1) :rocket:

PR #895, opened 2021-03-24
fix(select): adjusts padding to accommodate caret

dzucconi:
Re: https://artsy.slack.com/archives/C9XJKPY9W/p1616618231018200

This PR fixes a minor issue I noticed when investigating; that longer values will cut into the caret issue.

Regarding the problem mentioned in chat: I'm going to just manually pass `display="inline-flex"` to the few instances where this is a problem in Force. I don't want to change it for default `inline` variants because the current display property makes more sense when using columns for layout.

### DIFF
--- a/packages/palette/src/elements/Select/Select.story.tsx
+++ b/packages/palette/src/elements/Select/Select.story.tsx
@@ -7,9 +7,13 @@ export default {
 }
 
 const OPTIONS = [
-  { text: "First", value: "firstValue" },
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette@14.12.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.12.1) :rocket:

PR #896, opened 2021-03-25
feat(expandable): implements expandable

dzucconi:
This is a re-implementation of `Toggle` that drops the extraneous props and margins. I could also make this a breaking change and delete `Toggle` completely. But I think it'll be a bit easier to replace out the `Toggle`s in Force this way since there are quite as few and the removal of a bottom margin will have to be addressed individually.

Getting rid of the font size prop since that is only used a single place in Volt and can just be passed a custom `label` component. Ditto for `renderSecondaryAction` — used in a single place in Force and included an example in the stories... just pass a custom label.


damassi (approved):
🔥 

One q: does the off-centered ness of that focused state in the screen cap trouble you at all?

dzucconi:
Off-centeredness? Do you mean the way the text sits vertically?

The spec:



The old version also has a 5px top/bottom margin so this is pretty close to that as well. I'm assuming it's to pad out the hit area a bit.

damassi:
This bit: 

Screen Shot 2021-03-25 at 12 14 35 PM

artsyit:
:rocket: PR was released in [`@artsy/palette@14.13.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.13.0) :rocket:

PR #900, opened 2021-04-01
fix(theme): corrects media queries

dzucconi:
I noticed this while trying to use the `xs` media query inside of a styled-component using `themeGet`. The `media` util in Palette builds them but isn't theme-aware.

<del>Riddle me this: how is [this code working currently?](https://github.com/artsy/force/blob/5c79ce80607403c2e6e4de629500383d6ddcef59/src/v2/Components/NavBar/NavBar.tsx#L56-L57)</del> Answered my own question: it's not working currently 👅  and this should fix it.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.15.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.15.0) :rocket:

PR #903, opened 2021-04-02
bugfix: Ensure read more only operates on strings

damassi:
Makes it a little safer as we noticed it was blowing up in force when operating on null data.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.17.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.17.0) :rocket:

PR #907, opened 2021-04-16
fix(pagination): updates copy; adjusts aria

dzucconi:
* Designs say to use "Prev" (and this was previously "Prev")
* Some small a11y updates


artsyit:
:rocket: PR was released in [`@artsy/palette@14.18.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.18.1) :rocket:

PR #910, opened 2021-04-16
fix(pagination): correctly disables next/prev

dzucconi:
😅  😅  😅

### DIFF
--- a/packages/palette/src/elements/Pagination/__tests__/Pagination.test.tsx
+++ b/packages/palette/src/elements/Pagination/__tests__/Pagination.test.tsx
@@ -73,10 +73,10 @@ describe("Pagination", () => {
       })
 
       const prevButton = wrapper.find("[data-testid='prev']").first()
-      expect(prevButton.find("Link").prop("href")).toEqual("")
+      expect(prevButton.find("a")).toHaveLength(0)
 
       const nextButton = wrapper.find("[data-testid='next']").first()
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/Pagination/__tests__/Pagination.test.tsx
```diff
@@ -73,10 +73,10 @@ describe("Pagination", () => {
       })
 
       const prevButton = wrapper.find("[data-testid='prev']").first()
-      expect(prevButton.find("Link").prop("href")).toEqual("")
+      expect(prevButton.find("a")).toHaveLength(0)
```
dzucconi, line 76: So rather than having empty hrefs; it's better to just not be a link when "disabling"

packages/palette/src/elements/Pagination/__tests__/Pagination.test.tsx
```diff
@@ -207,10 +207,10 @@ describe("Pagination", () => {
         )
       })
 
-      it("renders the next button as disabled and calls the onNext function when clicked", () => {
+      it("renders the next button as disabled and the onNext function is not called when clicked", () => {
```
dzucconi, line 210: This spec was definitely just wrong.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.18.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.18.3) :rocket:

PR #911, opened 2021-04-20
Fixes unstyled text when using interpolation/fragments

dzucconi:
Closes: [FX-2864](https://artsyproduct.atlassian.net/browse/FX-2864)

There's now a pattern in a few places where we check to see if a prop/the children are simple text and provide default styling; but allow it to be customized as well. This was manifesting as unstyled text here:

Because the `typeof` children with some interpolated content is in fact an array.

### DIFF
--- a/packages/palette/src/elements/LabeledInput/LabeledInput.tsx
+++ b/packages/palette/src/elements/LabeledInput/LabeledInput.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useRef, useState } from "react"
+import { isSimpleChildren } from "../../helpers/isSimpleChildren"
 import { useThemeConfig } from "../../Theme"
 import { Box, splitBoxProps } from "../Box"
 import { Input, InputProps } from "../Input"
@@ -24,7 +25,7 @@ export const LabeledInput: React.ForwardRefExoticComponent<
 
   const variant: TextVariant = useThemeConfig({ v2: "small", v3: "xs" })
 
-  const isText = typeof label === "string" || typeof label === "number"
+  const isText = isSimpleChildren(label)
 
   return (
     <Box position="relative" {...boxProps}>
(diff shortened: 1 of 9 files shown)


packages/palette/src/elements/LabeledInput/LabeledInput.tsx
```diff
@@ -24,7 +25,7 @@ export const LabeledInput: React.ForwardRefExoticComponent<
 
   const variant: TextVariant = useThemeConfig({ v2: "small", v3: "xs" })
 
-  const isText = typeof label === "string" || typeof label === "number"
+  const isText = isSimpleChildren(label)
```
dzucconi, line 28: I guess `isText` might be a better name for this helper?

damassi: Yah i was gonna say the same becuse that's basically all we're checking 👍

packages/palette/src/helpers/isSimpleChildren.tsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  // Non-fragment interpolated children is an array:
+  // `>Exmaple ({2 + 2})<` => ["Example (", 4, ")"]
+  if (Array.isArray(children)) {
+    return children.every(isSimpleChildren)
+  }
+
+  return isSimple(children)
+}
```
damassi, line 31: Would be great if react shipped with a few handy utiils so devs didn't have to always write this recursive stuff

damassi (approved):
`isTextNode` (or something) might be a bit more descriptive, but nothing blocking here 👍

packages/palette/src/elements/LabeledInput/LabeledInput.tsx
dzucconi: :renamed:

artsyit:
:rocket: PR was released in [`@artsy/palette@14.19.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.19.0) :rocket:

PR #912, opened 2021-04-23
feat(theme): allows theme to be easily swapped

dzucconi:
Planning on allowing the `AppShell` theme to be configured with a `theme` key and this makes things easy.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.20.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.20.0) :rocket:

PR #916, opened 2021-05-04
WIP: Adapt Carousel to V3 / Swipe support  

damassi:
@dzucconi - This mostly works, but I can't seem to figure out how to track the index when swiping, due to the fact that there's only one `children` element. Basically, we need to do `children.children` to get the cell count and update the index.


packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -206,19 +166,27 @@ export const Carousel: React.FC<CarouselProps> = ({
...
-                {child}
-              </Cell>
-            )
-          })}
-        </Rail>
-      </Viewport>
+      <Wrapper>
+        <Viewport ref={viewportRef as any}>
+          <Swiper Rail={Rail}>
+            <Rail style={{ transform: `translateX(${offset})` }}>
```
damassi, line 172: This is the spot. `Swiper` interprets `children.length` to be 1 (correctly) but somehow we need to get in one more level deep. 

(Ideally though we'd be able to delete this component completely and somehow pass the offset into Swiper, but when I do things go haywire.)

dzucconi: Why would we need Swiper at all — Swiper is intended to be the touch version of this component itself.

packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -206,19 +166,27 @@ export const Carousel: React.FC<CarouselProps> = ({
...
-
-            return (
-              <Cell key={i} ref={ref} pr={!isLast && CELL_GAP_PADDING_AMOUNT}>
-                {child}
-              </Cell>
-            )
-          })}
-        </Rail>
-      </Viewport>
+      <Wrapper>
```
dzucconi, line 169: I wouldn't include FullBleed here, which is more of a layout concern. This makes the component less generally useful.

damassi: Thoughts on it being a shared quality of all v3 comonents?

packages/palette/src/elements/Carousel/Carousel.tsx
damassi: ?

packages/palette/src/elements/Carousel/Carousel.tsx
dzucconi: We should just handle it in the wrapper in Force rather than in each situation. In general anything in the design system shouldn't have any outer margins of any kind.

packages/palette/src/elements/Carousel/Carousel.tsx
dzucconi: I asked this before we chatted in Slack. I understand the issue here but my initial assumption was that we'd only somehow detect the cases where we don't have inertial scroll and present the arrows — so similar to what we're doing already in Force with touch. Can we maybe restate the purpose of what exactly it is we're trying to accomplish because I might have more incorrect assumptions.

damassi: A carousel that swipes and you hover over it and it shows the arrows

damassi: There's a lot of moving pieces between swiper and caruosel and so on, things that i don't totally understand. Meanwhile, its 905 mostly working except the state isn't sync'ed between swipe and arrow click. If you wouldn't mind pulling this down and looking at that, and then after -- if you feel like refactoring this to what you envision in your mind -- can can we iterate? This auctions page was supposed to ship by demo days and i'm blocked on this part.

dzucconi: As far as I can tell then, the problem we're going to face with that is that we're conflating two different modes of movement: scroll and transform. They can't be used together, really. We either refactor Carousel so that instead of transforming we smooth scroll to different positions, or we implement some kind of synthetic scroll behavior using transform. Both are going to have some trade-offs. Let's chat about this before proceeding.

dzucconi: I know it looks like the problem is the state but it's really not just that. Ping me in Slack when you're free and we can work through it.

dzucconi: The other thing to consider here is that these two components have two totally different notions of 'progress'. Swiper on the one hand is cell based — so each individual cell is a step. Carousel is page-based, so it moves through multiple cells at a time.

PR #917, opened 2021-05-06
Shelf component

dzucconi:


packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,289 @@
...
+          disabled={pageIndex === pages.length - 1}
+          aria-label="Next page"
+        >
+          <ChevronIcon direction="right" width={15} height={15} />
+        </Next>
+      </Nav>
+
+      <FullBleed
+        {...(mounted
+          ? { left: undefined, right: undefined, marginRight: undefined }
```
dzucconi, line 163: I'd love to figure out a way to do this without JS but I'm not sure it's possible; here we just fake the initial state for SSR — it bleeds off the right edge, but the left is aligned. Once it's mounted we can re-enable the full-bleed then use the offset below as the first element's padding-left and last element's padding-right.

packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+      setCursor(pages.indexOf(nearestPage))
+
+      const __progress__ = Math.ceil(
+        (viewport.scrollLeft * 100) / (pages[pages.length - 1] - 1)
+      )
+
+      setProgress(__progress__)
+      onProgress && onProgress(progress)
+    }
```
dzucconi, line 105: I had the misfortune of getting this whole thing working before trying to tackle the full-bleed overflow.

packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,289 @@
...
+        )
+      })
+
+      setCursor(pages.indexOf(nearestPage))
+
+      const __progress__ = Math.ceil(
+        (viewport.scrollLeft * 100) / (pages[pages.length - 1] - 1)
+      )
+
+      setProgress(__progress__)
```
damassi, line 103: Interesting, so whats happening here is yr setting a computed value which then outputs `progress` which is then passed up through the system. Took a sec to understand due to the shadowed var. Maybe naming it something like `computedProgress` could reduce some friction here?

packages/palette/src/elements/Shelf/Shelf.tsx
damassi: This def makes sense! though would suggest moving this block into a variable and adding this as a comment ;)

packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,289 @@
...
+  const [progress, setProgress] = useState(0)
+  const [offset, setOffset] = useState(0)
+
+  const init = () => {
+    if (containerRef.current === null || viewportRef.current === null) return
+
+    // Set page-stops
+    const { current: viewport } = viewportRef
+    const values = cells.map(({ ref }) => ref.current.clientWidth)
+    setPages(paginateCarousel({ viewport: viewport.clientWidth, values }))
```
dzucconi, line 61: This is one of those times the fix I was just looking for turned out to be simple: here I was passing in the viewport width to calculate the page-stops. Now that the viewport is full-bleed we actually want the container width which specifies that more narrow window. It works 👀  !

dzucconi: Lol, I think I spoke too soon. I'm close though.

packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,307 @@
...
+  }, [cells, offset])
+
+  useEffect(() => {
+    init()
+
+    window.addEventListener("resize", init)
+    return () => {
+      window.removeEventListener("resize", init)
+    }
+  }, [init])
```
dzucconi, line 94: I had this mostly correct; but the offset value was stale without realizing it. (I had initially only had `cells` in this dependency array — having `init` without the `useCallback` was causing infinite renders).

Note to self that frequently just including some kind of debugging visual really clarifies what's going on — in this case I added markers overlaying where the page stops were and I immediately noticed they were out of sync once I resized the browser.

packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+    if (viewport.scrollTo) {
+      viewport.scrollTo({ left: xPosition, behavior: "smooth" })
+      return
+    }
+
+    viewport.scrollLeft = xPosition
+  }
+
+  // One side-effect of setting the
```
dzucconi, line 148: Woops forgot to finish my comment here.

dzucconi:
I'll follow up with some accessibility updates; but this at least gets the thing visually correct so that we can try it out.

packages/palette/src/elements/Shelf/Shelf.tsx
damassi: > having init without the useCallback was causing infinite renders

Wow, this makes me think that the whole reason we couldn't do URL driven state updates in thet Artwork filter is that that part of the code wasn't wrapped within a `useCallback` 😠 I remember the same thing happening. Something was being triggered leading to infinite renders. (In any case, with this new artwork filter work I plan on rectifying this.)

damassi (approved):
Amazing, thanks for pushing through and getting this over the line. I'm way out of practice on this kind of stuff and the code looks clear 🙏 

There's some failures in CI but once that's all fixed up 🍏

artsyit:
:rocket: PR was released in [`@artsy/palette@14.22.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.22.0) :rocket:

PR #918, opened 2021-05-06
feat: Add alignItems prop to shelf

damassi:
This updates the `Shelf` so that we can position items within it. Current use-case prompting this change is the MyBids component, which doesn't work bottom up.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.23.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.23.0) :rocket:

PR #919, opened 2021-05-06
feat: Add showProgress prop to Shelf

damassi:
Adds a new `showProgress` prop to Shelf, so that one can show / hide the progress bar. This makes shelf a more generic `Carousel` component. Also updates stories. 

shelf

### DIFF
--- a/packages/palette/src/elements/Shelf/Shelf.story.tsx
+++ b/packages/palette/src/elements/Shelf/Shelf.story.tsx
@@ -41,3 +41,23 @@ export const Default = () => {
     </Box>
   )
 }
+
+export const CenterAlign = () => {
+  return (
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette@14.24.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.24.0) :rocket:

packages/palette/src/elements/Shelf/Shelf.story.tsx
```diff
@@ -41,3 +41,23 @@ export const Default = () => {
     </Box>
   )
 }
+
+export const CenterAlign = () => {
```
damassi, line 45: (Just remembered storybooks states, updating now)

PR #923, opened 2021-05-07
fix: [Shelf] Dont show arrows if on touch device

damassi:
Noticed that arrows would appear when interacting on phone

### DIFF
--- a/packages/palette/src/elements/Shelf/Shelf.tsx
+++ b/packages/palette/src/elements/Shelf/Shelf.tsx
@@ -304,6 +304,10 @@ const Arrow = styled(Clickable)`
     opacity: 0;
     cursor: default;
   }
+
+  @media (hover: none) {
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -304,6 +304,10 @@ const Arrow = styled(Clickable)`
     opacity: 0;
     cursor: default;
   }
+
+  @media (hover: none) {
+    display: none;
+  }
```
damassi, line 310: @dzucconi - Seems to be the right way to do this?

artsyit:
:rocket: PR was released in [`@artsy/palette@14.27.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.27.0) :rocket:

PR #925, opened 2021-05-07
fix(shelf): smoother progress

dzucconi:
Quick little fix to just tie the progress directly to scroll. I think I was stuck in thinking about this from the cell-based older components.

### DIFF
--- a/packages/palette/src/elements/ProgressBar/ProgressBar.tsx
+++ b/packages/palette/src/elements/ProgressBar/ProgressBar.tsx
@@ -42,6 +42,7 @@ export const ProgressBar: React.FC<ProgressBarProps> = ({
         style={{
           transition,
           transform: `translateX(-${100 - percentComplete}%)`,
+          backfaceVisibility: "hidden",
         }}
       />
     </Box>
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/ProgressBar/ProgressBar.tsx
```diff
@@ -42,6 +42,7 @@ export const ProgressBar: React.FC<ProgressBarProps> = ({
         style={{
           transition,
           transform: `translateX(-${100 - percentComplete}%)`,
+          backfaceVisibility: "hidden",
```
damassi, line 45: @dzucconi where o where do you discover all these mysterious and obscure css props?

dzucconi: For whatever reason without the transition this was rendering a bit weird (like a subpixel flicker kind of thing). I never understood why this works but sometimes if you're seeing some flicker from a transform; turning off the backface-visibility fixes it. 🤷

dzucconi: (Maybe doing a 3d version of the transform would be better and also solve it)

PR #930, opened 2021-05-10
fix(button): improves vertical centering in some situations

dzucconi:
That `1px` was a hold-over from the original implementation — I think it was a hack to fix the vertical centering in a few spots. Occasionally on 1x DPI screens the text is still 1px off (when the parent is also vertically centering, depending on its location on the page). This _should_ be the correct fix for that. We never have 2-line buttons so this should be fine to do. Publishing a canary to double check before merging.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.28.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.28.1) :rocket:

PR #934, opened 2021-05-18
[wip]: implements shelf scrollbar

dzucconi:
Opening a draft here to publish a canary and check this out in context and to hash out the direction.

So what `ShelfScrollMirror` does is: accepts a viewport (the parent of the rail in the carousel), then creates a new div, overflows it to the same width, then keeps the scroll positions of either in sync with one another. So when you scroll when the other moves accordingly and vice versa. The benefit of this is that it lets us decouple the scrollbar from the scrollable viewport (so that it doesn't have to be the same width). Then I'm using some CSS to customize how the scrollbar looks. This is where we start running into caveats.

* Chrome works perfectly, of course
* Firefox claims to support scrollbar CSS now but I can't get it to work yet. Need to research that a bit more. On Firefox this doesn't look terrible and at least works as a scrollbar. But it's not to spec. I've got some last resort ideas to hack around that.
* Safari works almost perfectly — there's a `scrollbar-corner` that doesn't do anything in this case occupying like ~20px on the right hand side of the scrollbar. Once you start to style it it shows up in Chrome too. My thinking is that I can style it then just offset the div to hide it.
* We probably want a bigger hit area. Can just make the scrollbar bigger then position a `pointer-events: none` cover over part of it to hide it. Transforming that could give us the animated transition you mentioned in chat @damassi.

Some of the benefits of this approach are:
* it's a native scrollbar so it works exactly how you'd expect it to
* aside from any CSS hacks; it's really simple
* since everything is happening in the DOM, no React renders need to happen

I left what I initially started with `ShelfScrollBar` in here for now. We'd have to implement the drag scrolling and click to scroll on the track etc. Probably not the worst thing in the world but before I ran into the cross-browser CSS caveats I thought I was a genius.

------

### Chrome

https://user-images.githubusercontent.com/112297/118828841-9f72d500-b88b-11eb-85db-9d5ad67ff2ae.mov

### Firefox

https://user-images.githubusercontent.com/112297/118829041-c9c49280-b88b-11eb-9d51-33e5fcb17f30.mov

### Safari

https://user-images.githubusercontent.com/112297/118829188-e4970700-b88b-11eb-91ef-9c13932b79a7.mov

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@14.29.1-canary.934.17174.0
  # or 
  yarn add @artsy/palette@14.29.1-canary.934.17174.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Shelf/Shelf.tsx
+++ b/packages/palette/src/elements/Shelf/Shelf.tsx
@@ -15,10 +15,10 @@ import { visuallyDisableScrollbar } from "../../helpers/visuallyDisableScrollbar
 import { ChevronIcon } from "../../svgs"
 import { Box, BoxProps } from "../Box"
 import { CELL_GAP_PADDING_AMOUNT, paginateCarousel } from "../Carousel"
-import { CarouselBar } from "../CarouselBar"
 import { Clickable } from "../Clickable"
 import { FlexProps } from "../Flex"
 import { FullBleed } from "../FullBleed"
+import { ShelfScrollMirror } from "./ShelfScrollMirror"
 
 /** ShelfProps */
 export type ShelfProps = BoxProps & {
@@ -53,7 +53,6 @@ export const Shelf: React.FC<ShelfProps> = ({
 
   const [mounted, setMounted] = useState(false)
   const [pages, setPages] = useState([0])
-  const [progress, setProgress] = useState(0)
   const [offset, setOffset] = useState(0)
 
   const init = useCallback(() => {
@@ -117,12 +116,6 @@ export const Shelf: React.FC<ShelfProps> = ({
       })
 
       setCursor(pages.indexOf(nearestPage))
-
-      // Sets a synthetic value between 0 and 100
-      setProgress(
-        (viewport.scrollLeft / (viewport.scrollWidth - viewport.clientWidth)) *
-          100
-      )
     }
 
     viewport.addEventListener("scroll", handler, { passive: true })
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Shelf/Shelf.tsx
```diff
@@ -273,11 +264,12 @@ const Rail = styled(Box)`
...
   width: 100%;
   height: 100%;
-  margin: 0;
+  margin-top: 0;
+  margin-left: 0;
+  margin-right: 0;
   padding: 0;
   list-style: none;
   white-space: nowrap;
-  align-items: ${(p) => p.alignItems};
```
dzucconi, line 280: This is already in `Box`

damassi (commented):
Oh yeah, aside from that mysterious CSS this is nice and simple. 

Might be worth posting some screenshots of what it (currently) looks like in FF and safari just to get a feel for the possible tradeoffs vs hacks

packages/palette/src/elements/Shelf/Shelf.tsx
damassi: rookie mistake 😅

dzucconi:
Hm, maybe drafts don't trigger canary builds?

dzucconi:
Updated with screen recordings

dzucconi:
🤔  Could maybe layer the custom scrollbar implementation over a real scrollbar and then just use `pointer-events: none` on it. That way I would only be implementing the display (which is already here in this PR) and wouldn't have to implement any of the interactivity which would be taken care of by the native control. 🤔

dzucconi:
Leaving this branch alone for posterity but closing in favor of a fully custom scrollbar implementation.

PR #942, opened 2021-06-02
Some small cleanup

pvinis:
Two things here:
- Removed `PageViewsIcon` because I saw it was from `reaction`, I searched in `force` and couldn't see it, so I guess it's safe to remove.
- Blacklisted some more names, since they are the internal ones exported, and they just make our icon list seem broken
Screenshot 2021-06-02 at 15 42 26


packages/palette-docs/content/docs/tokens/Icons.mdx
```diff
@@ -12,7 +12,7 @@ for the masterlist.
 
 <Flex flexWrap="wrap">
   {Object.entries(AllIcons).map(([iconName, IconComponent], index) => {
-    const blacklist = ["ChevronIcon", "CreditCardIcon", "Icon"]
+    const blacklist = ["ChevronIcon", "CreditCardIcon", "Icon", "G", "Title", "Circle", "Path", "Rect"]
```
jonallured, line 15: While you're here, how would you like to update this name to something nicer like `blocklist`?

pvinis:
@jonallured should i remove this file too? https://github.com/artsy/palette/blob/master/packages/palette/src/svgs/_icons.story.tsx

jonallured:
Honestly I'm not sure about how we use storybooks - I'll defer to @damassi on that point!

pvinis:
@damassi any feedback here?

damassi:
We need to add a redirect from the docs to storybooks, but its till not quite ready yet -- we've exclusively been adding to storybooks these days: https://palette-storybook.artsy.net/?path=/story/icons--all-icons

pvinis:
ok. then this PR should be ready to go I think.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.36.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.36.0) :rocket:

PR #944, opened 2021-06-04
fix(radio): allows for custom labels to occupy immediate container

dzucconi:
This is to accommodate a situation we have on the artwork sidebar where we need to float over a price on the right hand side of the container.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.33.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.33.1) :rocket:

PR #946, opened 2021-06-07
feat(button): supports `as` prop

dzucconi:
This lets us use `<Button as={RouterLink} to="/whatever">` etc. As always the typing here is a problem but since this is more of an escape hatch for a few one-offs I feel fine about that.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.34.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.34.0) :rocket:

PR #948, opened 2021-06-11
fix(popover): skips focus management on initial mount

dzucconi:
Minor bug fix.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.35.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.35.1) :rocket:

PR #949, opened 2021-06-11
fix(tooltip): avoids blowing away child props

dzucconi:
Another small bug fix. Always forget that it's possible to overwrite the existing props when using `cloneElement`

### DIFF
--- a/packages/palette/src/elements/Tooltip/Tooltip.tsx
+++ b/packages/palette/src/elements/Tooltip/Tooltip.tsx
@@ -52,11 +52,11 @@ export const Tooltip: React.FC<TooltipProps> = ({
       {React.cloneElement(children, {
         ref: anchorRef,
         tabIndex: 0,
-        onClick: handleClick,
-        onMouseOver: activate,
-        onMouseOut: deactivate,
-        onFocus: activate,
-        onBlur: deactivate,
+        onClick: compose(handleClick, children.props?.onClick),
+        onMouseOver: compose(activate, children.props?.onMouseOver),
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Tooltip/Tooltip.tsx
```diff
@@ -87,12 +97,9 @@ const truncate = (tip: string): string => {
...
-  box-shadow: ${DROP_SHADOW};
-  cursor: default;
-  pointer-events: none;
-`
+const compose = (a?: (...args: any) => any, b?: (...args: any) => any) => {
+  return (...args) => {
+    a?.(...args)
+    b?.(...args)
+  }
+}
```
damassi, line 105: We have `lodash` in the project

dzucconi: We don't :)

damassi: 🤯 wow, impressed... A first?

dzucconi: https://github.com/artsy/palette/pull/755

damassi: haha @dzucconi, of course 👌

artsyit:
:rocket: PR was released in [`@artsy/palette@14.35.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.35.2) :rocket:

PR #951, opened 2021-06-17
feat(placeholder): implements placeholder

dzucconi:
I'm going to test this on the homepage rebuild, but what I wanted to do here was just to create a comparison between the Skeleton and this technique.

I noticed on the homepage that the placeholders cause a significant amount of dropped frames. Again, turning off the animation solves this. My idea here is that rather than animate each box individually we just overlay a big fading box over the whole thing.

In the otherwise identical stories this is a difference between a solid 60fps and 30fps + dropped frames.

Rather than have these two competing implementations it'd probably make sense to just keep the Placeholder version; rename it to Skeleton, and then the existing placeholders simply won't be animated until we wrap them with the outer skeleton wrapper. Alternatively we could do the deprecate + replace dance.

-----

You can see the difference here where in I'm instead fading to red to demonstrate:

### Old:

<details>
  <br />
  
  ```sh
  npm install @artsy/palette@14.36.2-canary.951.17823.0
  # or 
  yarn add @artsy/palette@14.36.2-canary.951.17823.0
  ```
</details>


damassi (approved):
Implementation looks good, but mind adding a screencap? I _think_ I understand...

dzucconi:
@damassi — added some screencaps to the description.

I'm going to go ahead and replace Skeleton with this. This means that any previous skeletons will won't be animated until we wrap them in the animation wrapper.

damassi:
Performance is performance, but lets see how it feels on staging before shipping to prod! I winced a bit seeing the big block, compared whats currently a (mostly) nice visual implementation.

dzucconi:
It looks identical.

damassi:
> It looks identical.

Can you elaborate a bit more? Pretty drastic before / after: 

Screen Shot 2021-06-21 at 11 44 44 AM

dzucconi:
Imagine that overlay is white. and fading between 0 and 0.5 opacity placed over boxes which are `black10` — that looks the same as boxes that are `black10` fading to `black5`

dzucconi:
Sorry I didn't actually want to merge this yet; I'm gonna revert. Going to just replace the Skeleton implementation with Placeholder

damassi:
Ah, sorry about that

artsyit:
:rocket: PR was released in [`@artsy/palette@14.37.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.37.0) :rocket:

PR #952, opened 2021-06-21
Revert "feat(placeholder): implements placeholder"

dzucconi:
Reverts artsy/palette#951


dzucconi:
Eh, I'll just do it now rather than revert this revert

PR #959, opened 2021-06-29
fix(join): supports nested children

dzucconi:
Closes [WP-60](https://artsyproduct.atlassian.net/browse/WP-60)

You may want to conditionally render multiple things inside a join, which may require the use of a fragment. This change supports those cases.


artsyit:
:rocket: PR was released in [`@artsy/palette@14.38.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.38.1) :rocket:

PR #960, opened 2021-06-29
fix(carousel,swiper): supports conditional children

dzucconi:
#minor


artsyit:
:rocket: PR was released in [`@artsy/palette@14.38.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.38.2) :rocket:

PR #969, opened 2021-07-09
fix(PX-4310): get page count information from Swiper and Carousel

kiryl-zubarau:
This PR adds `onPageCountChange` event to be able to make dynamic data loading for the `Swiper` and `Carousel`.

I want to add batch loading for the artworks on the Artists page(PPP) and as far as I see now it's impossible to implement that without page count information. Please let me know if you have another option.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4310

### DIFF
--- a/packages/palette/src/elements/Carousel/Carousel.tsx
+++ b/packages/palette/src/elements/Carousel/Carousel.tsx
@@ -182,7 +182,7 @@ export const Carousel: React.FC<CarouselProps> = ({
     onChange && onChange(index)
   }, [onChange, index])
 
-  useEffect(() => {
+  useUpdateEffect(() => {
     onPageCountChange && onPageCountChange(pages.length)
   }, [onPageCountChange, pages.length])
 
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/Carousel/Carousel.tsx
```diff
@@ -180,6 +182,10 @@ export const Carousel: React.FC<CarouselProps> = ({
     onChange && onChange(index)
   }, [onChange, index])
 
+  useEffect(() => {
+    onPageCountChange && onPageCountChange(pages.length)
+  }, [onPageCountChange, pages.length])
```
dzucconi, line 187: Should this be running on initial mount as well? (See the above `useUpdateEffect` which skips that.

dzucconi (approved):
This makes sense to me. I suppose you could accomplish the same thing with an effect around `Children.count(cells)` in the parent component; though having to pull out the cells and assign them is annoying.

My only question is to whether or not we want to run this on mount.

👍

packages/palette/src/elements/Carousel/Carousel.tsx
kiryl-zubarau: Yea, make sense. missed the hook.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.43.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.43.0) :rocket:

PR #970, opened 2021-07-13
fix(useposition): silences uselayouteffect ssr warning

dzucconi:
[Shout out](https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a)

### DIFF
--- a/packages/palette/src/utils/usePosition.ts
+++ b/packages/palette/src/utils/usePosition.ts
@@ -1,7 +1,8 @@
 /**
  * Adapted from https://codesandbox.io/s/positioning-tooltip-rhplo
  */
-import { useLayoutEffect, useRef } from "react"
+import { useRef } from "react"
+import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect"
 
 export const POSITION = {
   "top-start": "top-start",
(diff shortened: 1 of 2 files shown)


packages/palette/src/utils/usePosition.ts
```diff
@@ -44,7 +45,7 @@ export const usePosition = ({
   const tooltipRef = useRef<HTMLElement | null>(null)
   const anchorRef = useRef<HTMLElement | null>(null)
 
-  useLayoutEffect(() => {
```
damassi, line 47: Or maybe just use `useEffect`? Is there a noticeable difference between the two UX-wise?

packages/palette/src/utils/usePosition.ts
```diff
@@ -44,7 +45,7 @@ export const usePosition = ({
...
+  useIsomorphicLayoutEffect(() => {
```
damassi, line 48: Or maybe just use `useEffect`? Is there a noticeable difference between the two UX-wise?

dzucconi: Oh, yeah I'm intentionally using useLayoutEffect here. It's kind of wild actually; the return from `getBoundingClientRect` is incorrect for long enough for the position to be wrong. You see it show up as a very brief flash as the popover repositions itself.

damassi: Cool 👍 Figured as much but wanted to be sure

artsyit:
:rocket: PR was released in [`@artsy/palette@14.43.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.43.1) :rocket:

PR #980, opened 2021-07-21
fix: Fix missing TextVariant export 

damassi:
An export was missed when completing the palette-tokens work, but it didnt' show up in force because renovate auto-update pr failed on type-check (and didn't notice). This fixes that.


artsyit:

PR #982, opened 2021-07-21
fix: Make sure to compile monorepo packages before publish

damassi:
Need to get rid of this lerna monstrosity. Its all good until one package depends on another


artsyit:

PR #984, opened 2021-07-26
fix(PX-4059): prevent link click if there is no next or prev page

kiryl-zubarau:
This PR doesn't allow clicking on the next or prev button if there is no next or prev page. 

JIRA - https://artsyproduct.atlassian.net/browse/PX-4059

Before

https://user-images.githubusercontent.com/79979820/126981637-beff0abd-a7a8-48f5-b077-d7589020a10b.mov

### DIFF
--- a/packages/palette/src/elements/Pagination/SmallPagination.tsx
+++ b/packages/palette/src/elements/Pagination/SmallPagination.tsx
@@ -22,11 +22,17 @@ export const SmallPagination: React.FC<PaginationProps> = (props) => {
   const handlePrevClick = (event: React.MouseEvent) => {
     if (previous) {
       onClick(previous.cursor, previous.page, event)
+    } else {
+      event.preventDefault()
     }
   }
 
   const handleNextClick = (event: React.MouseEvent) => {
-    onNext(event, nextPage)
+    if (hasNextPage) {
+      onNext(event, nextPage)
+    } else {
+      event.preventDefault()
+    }
   }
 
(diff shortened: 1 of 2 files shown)


kiryl-zubarau:
I'm not sure how to show the current component behavior, is it possible to use the local palette version in Force for example? Or maybe it makes sense to add the component to the storybook?

packages/palette/src/elements/Pagination/SmallPagination.tsx
```diff
@@ -22,11 +22,17 @@ export const SmallPagination: React.FC<PaginationProps> = (props) => {
...
+      event.preventDefault()
     }
   }
 
   const handleNextClick = (event: React.MouseEvent) => {
-    onNext(event, nextPage)
+    if (hasNextPage) {
+      onNext(event, nextPage)
+    } else {
+      event.preventDefault()
```
kiryl-zubarau, line 34: `preventDefault` is used to prevent link click.

dzucconi:
This is fine — this component should be deprecated and not used in v3 so there's no need to add it to Storybook.

artsyit:
:rocket: PR was released in [`@artsy/palette-tokens@2.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-tokens%402.1.2), [`@artsy/palette@14.45.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.45.1) :rocket:

PR #987, opened 2021-07-29
feat: Export v2/v3 theme constants

damassi:
This exports our palette theme constants so that we can access static values outside of the hook


artsyit:
:rocket: PR was released in [`@artsy/palette@14.46.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.46.0) :rocket:

PR #988, opened 2021-07-30
feat(textarea): supports v3

dzucconi:
Finally getting around to updating the textarea since it's needed for inquiries.

### DIFF
--- a/packages/palette/src/elements/TextArea/__tests__/TextArea.test.tsx
+++ b/packages/palette/src/elements/TextArea/__tests__/TextArea.test.tsx
@@ -87,14 +87,6 @@ describe("TextArea", () => {
     })
   })
 
-  it("takes a default value", () => {
-    getWrapper({ defaultValue: "This is the default value" })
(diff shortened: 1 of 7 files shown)


packages/palette/src/elements/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -87,14 +87,6 @@ describe("TextArea", () => {
     })
   })
 
-  it("takes a default value", () => {
-    getWrapper({ defaultValue: "This is the default value" })
-    expect(onChange).toHaveBeenCalledWith({
-      value: "This is the default value",
-      exceedsCharacterLimit: false,
-    })
-  })
```
dzucconi, line 96: Removing this spec since `onChange` shouldn't be called on mount. 🤞 this doesn't break something but I'll handle it if it does.

artsyit:
:rocket: PR was released in [`@artsy/palette@14.47.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.47.0) :rocket:

PR #1006, opened 2021-08-06
fix: Font getting inherited from clickable

rajsam003:
[PURCHASE-2813](https://artsyproduct.atlassian.net/browse/PURCHASE-2813)

Font style in BaseTab component was getting from Clickable component which is used as a prop. Clickable component has `font-style :"inherit"` due to which both were sharing the same styles. 

The solution is to wrap the BaseTab with Clickable instead of using it as a prop so that BaseTab gets its own style.

### DIFF
--- a/packages/palette/src/elements/Stepper/Stepper.tsx
+++ b/packages/palette/src/elements/Stepper/Stepper.tsx
@@ -56,44 +56,48 @@ export const Stepper: React.FC<StepperProps> = ({
       <BaseTabs separator={tokens.joinSeparator} fill={tokens.fill} {...rest}>
         {tabs.map((cell, i) => {
           return (
-            <BaseTab
+            <Clickable
               key={i}
-              as={Clickable}
               aria-selected={i === activeTabIndex}
               disabled={disableNavigation || i > currentStepIndex}
               onClick={handleClick(i)}
-              active={i === activeTabIndex}
-              variant={tokens.textVariant}
-              justifyContent={tokens.horizontalAlignment}
+              flex={1}
             >
-              <Flex
-                alignItems={tokens.verticalAlignment}
-                justifyContent="space-between"
-                flex={1}
+              <BaseTab
+                active={i === activeTabIndex}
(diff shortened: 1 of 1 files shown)


copperkraft (approved):
Nice solution!

dblandin:
I'm not sure how relevant the popover shifts in the UI Tests are. Seem unrelated to this change.

But I took a quick look at the stepper component in our storybooks to compare between this build and master - and it looks this change introduces some unintentional underline behavior for disabled tabs (see screen recordings below).

Also, FYI: we have a `publish-canary` CI step here that should have published a canary NPM version for you to use but the build may have run before the PR was opened. You might be able to re-push or trigger a new build to get that canary if you'd like to use it in Force before this PR is merged.

**Before**



**After**

dzucconi:
@rajsam003 I removed the `minor` tag and added the `canary` and `patch` tags. This is a fix so it'd be a patch release. And the `canary` tag tells CI to publish a canary version, so that you can validate in Force, as @dblandin mentions.

dzucconi:
You can see why the problem @dblandin points out happens here: https://github.com/artsy/palette/blob/928d334e003c922c83f381581163aff7e0d3a131/packages/palette/src/elements/BaseTabs/BaseTab.tsx#L49

The "correct" fix I'd be interested in seeing here would be to pull out a [`clickableMixin`

dzucconi:
This may seem like a trivial issue but it sort of exposes a very subtle problem that we should actually fix: how styled-components is handling ordering of classes in this case which is resulting in the specificity of Clickable winning over the BaseTab. I had previously thought this had fixed it: https://github.com/artsy/palette/blob/928d334e003c922c83f381581163aff7e0d3a131/packages/palette/src/elements/BaseTabs/BaseTab.tsx#L71-L76 which appears not to be the case.

I'm noticing another typographic issue as well with the themes for the tabs and I'd like to investigate further today. Going to close this for the time being and I'll update you when I know more.

PR #1007, opened 2021-08-06
fix(text): increases text styles specificity

dzucconi:
I just paired with @rajsam003 to figure this one out... working off https://github.com/artsy/palette/pull/1006

The way `BaseTab` is set up is that it expects to be a button, potentially. So rather than altering those styles, it'd be good to just be able to use `as={Clickable}` and have it work as you'd expect.

We can [use this recommendation from styled-components](https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity) to increase the specificity of the `Text` styles and that way we can use `Clickable` again without too much worry.

### DIFF
--- a/packages/palette/src/elements/Text/Text.tsx
+++ b/packages/palette/src/elements/Text/Text.tsx
@@ -65,37 +65,39 @@ export type TextProps = BaseTextProps &
 
 /** Text */
 export const Text = styled(Box)<TextProps>`
-  ${(props) => {
-    return getThemeConfig(props, {
-      v2: css`
-        ${variant({ variants: V2_TEXT_VARIANTS.small })}
-        ${textMixin}
+  && {
+    ${(props) => {
+      return getThemeConfig(props, {
+        v2: css`
+          ${variant({ variants: V2_TEXT_VARIANTS.small })}
+          ${textMixin}
 
         @media (min-width: ${themeGet("breakpoints.0")}) {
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Text/Text.tsx
```diff
@@ -65,37 +65,39 @@ export type TextProps = BaseTextProps &
 
 /** Text */
 export const Text = styled(Box)<TextProps>`
-  ${(props) => {
-    return getThemeConfig(props, {
-      v2: css`
-        ${variant({ variants: V2_TEXT_VARIANTS.small })}
-        ${textMixin}
+  && {
```
dzucconi, line 68: This increases specificity by doubling up the class names. `.Text` would become `.Text.Text`

dzucconi:
This was validated in Force using `yarn link`:

In Palette: 

```
yarn compile-palette
yarn link
```

In Force:

```
yarn link @artsy/palette
yarn start
# then when finished
yarn unlink-all
```

dblandin (approved):
Nice sleuthing! 🕵🏼 and thanks for the link to the styled-components documentation! 💯

artsyit:
:rocket: PR was released in [`@artsy/palette@14.53.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.53.0) :rocket:

PR #1008, opened 2021-08-06
fix(text): exports text-transform type

dzucconi:
Good catch @lilyfromseattle — once this is published you can cast as this. Alternatively I could just make this a string value but we'd lose the auto-complete 🤔


artsyit:
:rocket: PR was released in [`@artsy/palette@14.54.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4014.54.0) :rocket:

PR #1012, opened 2021-08-10
feat(multiselect): implements multi-select

dzucconi:
This roughs in `MultiSelect` — @olerichter00 you can add the canary and try it out and leave comments here about any changes. Will get a spec in before merging.

<details>
  <br />
  
  ```sh
  npm install @artsy/palette-charts@16.0.1-canary.1012.20112.0
  npm install @artsy/palette@17.0.1-canary.1012.20112.0
  # or 
  yarn add @artsy/palette-charts@16.0.1-canary.1012.20112.0
  yarn add @artsy/palette@17.0.1-canary.1012.20112.0
  ```
</details>


packages/palette/src/elements/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+    const handleResize = () => {
+      setWidth(anchorRef.current.offsetWidth)
+    }
+
+    handleResize()
+
+    window.addEventListener("resize", handleResize)
+    return () => {
+      window.removeEventListener("keydown", handleResize)
```
olerichter00, line 68:
```suggestion
      window.removeEventListener("resize", handleResize)
```

packages/palette/src/elements/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }, [])
+
+  useUpdateEffect(() => {
+    onSelect?.(selection)
+  }, [onSelect, selection])
```
olerichter00, line 74: I guess we want to run `onSelect?.(selection)` inside of the checkbox's `onSelect` callback. Updating the value with `useUpdateEffect` calls `onSelect` every time the parent component rerenders if it's used without `useRef` or `useCallback`.

dzucconi: How's that the case? It should only call it if either `onSelect` or `selection` change.

olerichter00: This will give a `Warning: Maximum update depth exceeded. `:

```
import { MultiSelect } from "@artsy/palette"
import React, { useReducer } from "react"

const sizeOptions = [
  { text: "Small (under 40cm)", value: "SMALL" },
  { text: "Medium (40cm - 100cm)", value: "MEDIUM" },
  { text: "Large (over 100cm)", value: "LARGE" },
]

export const PriceDatabaseSearch = () => {
  const [filters, setFilter] = useReducer(
    (filters, newFilter) => ({ ...filters, ...newFilter }),
    {}
  )

  return (
    <MultiSelect
      options={sizeOptions}
      title="Size"
      onSelect={sizes => {
        setFilter({ sizes: sizes.map(size => size.value) })
      }}
    />
  )
}
```

dzucconi: Ah, thanks for the example. Should be fixed now. Give it another whirl once the canary re-publishes.

olerichter00: That works - thanks for the fix!

packages/palette/src/elements/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+      setSelection(nextSelection)
+      onSelect?.(nextSelection)
+    }
+  }
+
+  return (
+    <>
+      <Container ref={anchorRef as any} onClick={onVisible} {...rest}>
+        <Text variant="sm" lineHeight={1}>
```
olerichter00, line 100: Can we change the color here to match the `Input` component and the Figma design? 

```suggestion
        <Text variant="sm" lineHeight={1} color"black60">
```

olerichter00: I just tried to pass a component for the `title` and it breaks the layout if we don't set `display: inline;`. Maybe we can wrap `{title}` with an inline component to prevent the line break.

dzucconi: We should just restrict it to string, probably. Why do you need to pass it a component? Can you give me your use-case?

dzucconi: The other thing I'll have to do is rename this prop from title to label maybe? 

 — it looks like we have to keep the title/subtitle consistent with other inputs.

dzucconi: The other thing I'll have to do is rename this prop from title to label maybe? 

 — it looks like we have to keep the title/subtitle consistent with other inputs.

dzucconi: Maybe `name`? `label` kind of seems like what `title` _should_ be called; but I'm having to keep it consistent since historically that's been title

dzucconi: I've renamed `title` to `name` then added `title` and `description` support to match the other select/inputs. I'm still not sure what the need for a custom component is for what's now `name` but let me know and we'll figure it out.

olerichter00 (approved):
Looks great and works as expected 🚀

artsyit:
:rocket: PR was released in [`@artsy/palette@17.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.1.0) :rocket:

PR #1014, opened 2021-08-11
fix(PX-4326): Don't manually override background in Pill component in SavedAddresses [WIP]

Serge0n:
Jira Ticket: https://artsyproduct.atlassian.net/browse/PX-4326
Discussion: https://github.com/artsy/force/pull/7869#discussion_r669827365

It is my first PR for `Palette` so please double check

This PR  is a continuation of fixing [PX-4326](https://github.com/artsy/force/pull/8127).

In my opinion, a good option is to add a property `disabled` which will switch off hover and focus in the `Pill` component

Also, this PR resolves this ticket to:[PX-4367]( https://artsyproduct.atlassian.net/browse/PX-4367)

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.tsx
+++ b/packages/palette/src/elements/Pill/Pill.tsx
@@ -66,6 +66,8 @@ export type PillProps = ClickableProps & {
   focus?: boolean
   /** Forces hover state */
   hover?: boolean
+  /** Swich off focus, hover  */
+  disabled?: boolean
 } & (
     | {
         variant?: Extract<PillVariant, "textRound" | "textSquare" | "filter">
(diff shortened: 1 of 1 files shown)


dzucconi:
Can you add the relevant states to the story? https://github.com/artsy/palette/blob/37586c9a870e1eccf1ef5facf079a58216c13629/packages/palette/src/elements/Pill/Pill.story.tsx#L14

dzucconi:
I left this comment in Slack but: we shouldn’t just add a prop to hack around the design system, unless I'm missing something important here. If this is a distinct state then it should have a distinct visual representation that is backed by design.

dzucconi:
Closing this for now since the existing designs don't look like `Pill` — feel free to re-open if I'm mistaken.

dzucconi:
The other thing to note: that page is still v2. Pill doesn't even have a v2 variation.

PR #1021, opened 2021-08-19
fix(text): makes variant a required prop

dzucconi:
Closes: [DSWGW-85](https://artsyproduct.atlassian.net/browse/DSWGW-85)

So the thought here is that currently we set `"text"` as the default variant, which is v2, and this leads to some subtly broken designs. It likely makes more sense just to enforce this as a required prop since you should explicitly know what variant you want to use and I suspect that most people don't know what the default is at any given moment, let alone how that matters depending on which theme context you're in.

This will be a breaking release. I'll publish a canary and just use the type errors to drive out setting the defaults. In any v3 contexts we could safely set the variant to `"sm"` and in v2 to `"text"`

### DIFF
--- a/packages/palette-docs/content/docs/elements/CSSGrid.mdx
+++ b/packages/palette-docs/content/docs/elements/CSSGrid.mdx
@@ -16,10 +16,10 @@ for layouts based on a Bootstrap-style twelve column grid.
 
 <Playground>
   <CSSGrid gridTemplateColumns="repeat(2, 1fr)" gridGap={2} my={2}>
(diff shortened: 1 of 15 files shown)


damassi (approved):

dzucconi:
OK, this is maybe a bit premature. This breaks a lot of other things that just happen to use `styled(Text)`. Stepping back, the better approach might be to just locate any examples in Force and set them to the appropriate variants and then just set the default here to `"sm"`  which is what the default should be for the global styles anyway.

PR #1031, opened 2021-09-07
fix: apply focus styles to the button when it is active

dimatretyak:
### Description
I am a user and I click on the button, the focus state remains (i.e. a blue outline around the pill) when it should disappear if I release the button

Addresses [FX-3250]

### Demo
https://user-images.githubusercontent.com/3513494/132342147-9552007b-0e8e-45ca-a904-e89c8bb5ac88.mp4

[FX-3250]: https://artsyproduct.atlassian.net/browse/FX-3250

### DIFF
--- a/packages/palette/src/elements/Button/v3/Button.tsx
+++ b/packages/palette/src/elements/Button/v3/Button.tsx
@@ -121,7 +121,7 @@ const Container = styled.button<ContainerProps>`
         }
       }
 
-      &:focus {
+      &:active {
         outline: 0;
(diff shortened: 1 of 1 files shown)


dzucconi:
So: technically this isn't a bug. This is more of a design question. We want a focus state and we don't want to remove it from all buttons. What this ticket is really suggesting is that we roll focus-visible into the design system OR design an alternative active state. Either way this change would be inappropriate. I can open a broader ticket in the design system board if design feels like this is an issue that needs addressing.

PR #1034, opened 2021-09-15
feat: extracts horizontal overflow component from base tabs

dzucconi:
This PR extracts the overflow with fade-out affordance from `BaseTabs` as a component called `HorizontalOverflow` — not the greatest name; open to suggestions.


packages/palette/src/elements/HorizontalOverflow/HorizontalOverflow.tsx
```diff
@@ -0,0 +1,100 @@
...
+    /* Hide when scrolled all the way over */
+    transition: opacity 250ms;
+    opacity: ${({ atEnd }) => (atEnd ? 0 : 1)};
+  }
+`
+
+const Viewport = styled(Box)`
+  overflow-x: auto;
+  -webkit-overflow-scrolling: touch;
+  ${visuallyDisableScrollbar}
```
dzucconi, line 48: I am visually disabling the scrollbar here now; since it's come up multiple times. I suspect we should maybe add the synthetic one at some point but this requires some design thought.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@16.4.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4016.4.0), [`@artsy/palette@17.4.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.4.0) :rocket:

packages/palette/src/elements/BaseTabs/BaseTabs.tsx
```diff
@@ -1,58 +1,10 @@
...
-  JustifyContentProps,
-  padding,
-  PaddingProps,
-} from "styled-system"
+import React, { useMemo } from "react"
 import { flattenChildren } from "../../helpers/flattenChildren"
 import { useThemeConfig } from "../../Theme"
-import { splitProps } from "../../utils/splitProps"
 import { Box, BoxProps } from "../Box"
+import { HorizontalOverflow } from "../HorizontalOverflow/HorizontalOverflow"
```
damassi, line 5: tis a good name 👍

packages/palette/src/elements/index.tsx
```diff
@@ -24,6 +24,7 @@ export * from "./Flex"
 export * from "./FullBleed"
 export * from "./Grid"
 export * from "./GridColumns"
+export * from "./HorizontalOverflow"
```
damassi, line 27: Should this be exposed in the docs or is this more of an internal component? We should be sure to add things there as new components are created, or at least stub out the file so things aren't buried unintentionally

dzucconi: Ah, yeah I forgot; I extracted this for https://github.com/artsy/force/pull/8382 so it should be documented. Will follow up

damassi: 🙏

PR #1037, opened 2021-09-17
fix(image): reduce generated classname count

dzucconi:
An attempt to reduce the number of classnames. This in particular is a problem for Shelf components where we tend to have many different sizes; each of which will have a unique classname.

The one gotcha here might be if we use responsive props for width/height anywhere. I'm going to publish a canary and take a look at the type damage and if it's not huge; make the changes.

Also removing `CleanTag` from here; if we get warnings we'll do what we did in Force which is to use `shouldForwardProp: (prop, defaultValidatorFn) => defaultValidatorFn(prop),`

<details>
  <br />
  
  ```sh
  npm install @artsy/palette-charts@16.6.1-canary.1037.20678.0
  npm install @artsy/palette@17.6.1-canary.1037.20678.0
  # or 
  yarn add @artsy/palette-charts@16.6.1-canary.1037.20678.0
  yarn add @artsy/palette@17.6.1-canary.1037.20678.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Avatar/Avatar.tsx
+++ b/packages/palette/src/elements/Avatar/Avatar.tsx
@@ -6,7 +6,9 @@ import { Image, ImageProps } from "../Image"
 import { Text } from "../Text"
 import { V2_TOKENS, V3_TOKENS } from "./tokens"
 
-export interface AvatarProps extends FlexProps, Partial<ImageProps> {
+export interface AvatarProps
+  extends FlexProps,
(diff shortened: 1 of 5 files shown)


dzucconi:
32 errors. It's not a trivial amount of work but it's manageable. 😵‍💫 

I very much want to rip the band-aid off and just re-implement it to satisfy [all of the problems](https://artsyproduct.atlassian.net/browse/DSWGW-79) — but I know if I do that we're going to be stuck with this lazy load image library for all of eternity.

Re-grouping...

PR #1038, opened 2021-09-20
fix(modal): fixes stale focusable inputs

dzucconi:
A exhaustive dependency check would have caught this so I'm adding an eslint config along with enabling that plugin (finally).


packages/palette/.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    },
+    ecmaVersion: 12,
+    sourceType: "module",
+  },
+  plugins: ["react", "react-hooks", "@typescript-eslint"],
+  rules: {
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
+  },
+}
```
damassi, line 25: How was this not already in? Ah, we were using [ts-lint](https://github.com/artsy/palette/blob/master/packages/palette/tslint.json). Can we remove refs, etc?

dzucconi: 👍

packages/palette/src/elements/Modal/ModalBase.tsx
```diff
@@ -1,9 +1,9 @@
-import React, { useEffect, useRef, useState } from "react"
+import React, { useEffect, useRef } from "react"
 import { createPortal } from "react-dom"
 import { RemoveScroll } from "react-remove-scroll"
 import styled from "styled-components"
 import { zIndex as systemZIndex, ZIndexProps } from "styled-system"
-import { useCursor } from "use-cursor"
+import { useFocusLock } from "../../utils/useFocusLock"
```
damassi, line 6: I wonder if we could rename `utils` to `hooks` now, as that's mostly whats in there

packages/palette/src/utils/useFocusLock.ts
```diff
@@ -0,0 +1,88 @@
...
+    document.addEventListener("keydown", handleKeydown)
+
+    return () => {
+      // Return the focus
+      focusedElPriorToOpen.focus()
+
+      document.removeEventListener("keydown", handleKeydown)
+    }
+  }, [handleNext, handlePrev])
+}
```
damassi, line 88: 😅 - I know you explained it a bit in standup, but whats the usecase for this again? I think I follow but curious if you could elaborate a bit.

dzucconi: Just a basic usability/accessibility thing with modals; you want to trap focus within them. Consider what'd happen if you tabbed past the last element in a modal when it wasn't locked within (which is what happens with our current login/sign up modal btw): focus moves into a part of the page that you can't click without closing the modal.

This hook locks focus within the given element, so we pass it the modal container and then while the modal is open you can only tab around to things within the dialog. When the dialog is closed, focus is yielded back to the last thing that was focused on the page.

packages/palette/src/elements/Modal/ModalBase.tsx
dzucconi: I think that makes sense. There's one function `splitProps` which is a util; so we could just have a new hooks dir and move those. Will be a separate PR.

packages/palette-docs/content/docs/hooks/useFocusLock.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+      <Box ref={containerEl}>
+        <input placeholder="First tabstop" />
+        <input placeholder="Second tabstop" />
+        <button>Third tabstop</button>
+      </Box>
+    </>
+  )
+}
+```
```
dzucconi, line 34: Docs 🦅

dzucconi:
OK, this looks good in the canary. Added specs and docs. Merging.

artsyit:
:rocket: PR was released in [`@artsy/palette@17.7.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.7.2) :rocket:

PR #1042, opened 2021-09-21
fix(button): ensures submit buttons are same font-weight

dzucconi:
Minor thing I noticed in the iOS15 update. Buttons with `type="submit"` are automatically bolded.


artsyit:
:rocket: PR was released in [`@artsy/palette@17.7.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.7.1) :rocket:

PR #1045, opened 2021-09-22
fix(hooks): improves focus lock and mutation observer hooks

dzucconi:
* Ignores mutation observer hook when mutation observer isn't defined (specifically our version of Jest in Force)
* Improves docs
* Better handling of story surrounding clicks in focus lock


artsyit:
:rocket: PR was released in [`@artsy/palette@17.7.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.7.3) :rocket:

PR #1047, opened 2021-09-24
Deprecate <Link>

damassi:
It doesn't technically *need* to be deprecated, but folks use it enough to need a pointer to `RouterLink`.


dzucconi (approved):
Yes, definitely

artsyit:
:rocket: PR was released in [`@artsy/palette@17.7.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.7.4) :rocket:

PR #1049, opened 2021-10-04
fix(horizontaloverflow): ensures internal heights match external

dzucconi:
Small fix for this component so that we can use it in places where we want a specific height without relying on padding.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@16.7.6`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4016.7.6), [`@artsy/palette@17.7.6`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.7.6) :rocket:

PR #1051, opened 2021-10-05
feat(tabs): supports autoscrolling on tab selection

dzucconi:
Closes: [DSWGW-96](https://artsyproduct.atlassian.net/browse/DSWGW-96)

`scrollIntoView` makes this nice and simple. This should just be on by default and there's no need to trigger it via prop.


artsyit:
:rocket: PR was released in [`@artsy/palette@17.8.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4017.8.0) :rocket:

PR #1057, opened 2021-10-08
fix(labeledinput): fixes height when title/description are present

dzucconi:
Since the title and description are contained within the input, we need to insert the children inside a wrapper around the input itself so that the label height is correct.


artsyit:
:rocket: PR was released in [`@artsy/palette@18.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.1.1) :rocket:

PR #1062, opened 2021-10-21
fix(autocomplete): display options if they are present regardless of query

dzucconi:
This makes more sense since the component is expecting any options passed into be pre-filtered. If you wanted to disable them for some query length you'd simply do so `{query.length !== 0 ? options : []}`


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.4.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.4.1), [`@artsy/palette@18.4.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.4.1) :rocket:

PR #1068, opened 2021-11-05
Revert "fix(tabs): add scrollIntoView for active tab"

dzucconi:
Reverts artsy/palette#1066

Few problems here — mainly that this causes components to auto-scroll into view when re-rendered.

Will follow this up with a more correct solution.


dzucconi:
OK.... I guess this never merged. https://github.com/artsy/palette/pull/1070 should fix this problem so we'll wait on that. Will re-open and merge incase we need a release before that's merged.

PR #1070, opened 2021-11-10
fix(tabs): avoids scrolling on mount

dzucconi:
So — the problem was that `scrollIntoView` was working — meaning it _was_ scrolling the active tab into view horizontally _as well as_ vertically.  This isn't really a problem when triggered via a click since to click something it's always going to be in the viewport. But this became a problem when we started using the prop to drive the active index since the initial mount is going to force the `scrollIntoView` as well.

Initially my thought was that we could use `useUpdateEffect` to skip the initial mount, but that has a few problems: one is that Force pages re-render _a lot_. Another is that if you set the initial index to something something that was off-screen it wouldn't be scrolled to initially.

Two things needed to happen:
* We need to manually scroll by getting a handle on the viewport ref and manipulating it's `scrollLeft` instead of just using `scrollIntoView`.
* We need to prevent scrolls from triggering on normal re-renders that don't change the index.

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@17.6.1-canary.1070.21704.0
  npm install @artsy/palette@18.6.1-canary.1070.21704.0
  # or 
  yarn add @artsy/palette-charts@17.6.1-canary.1070.21704.0
  yarn add @artsy/palette@18.6.1-canary.1070.21704.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/Tabs/Tabs.tsx
+++ b/packages/palette/src/elements/Tabs/Tabs.tsx
@@ -1,5 +1,12 @@
 import { TextVariant } from "@artsy/palette-tokens/dist/typography/types"
-import React, { createRef, useCallback, useEffect, useState } from "react"
+import React, {
+  createRef,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react"
 import { flattenChildren } from "../../helpers/flattenChildren"
 import { useThemeConfig } from "../../Theme"
 import { useUpdateEffect } from "../../utils/useUpdateEffect"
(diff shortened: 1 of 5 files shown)


packages/palette/src/elements/Tabs/Tabs.tsx
```diff
@@ -38,39 +45,48 @@ export const useTabs = ({
...
+    [children]
+  )
 
   const [activeTabIndex, setActiveTabIndex] = useState<number>(initialTabIndex)
-  const activeTab = tabs[activeTabIndex]
+
+  // Wraps active tab in a ref to prevent re-rendering.
+  // This means that we need to update the active tab before the index state change
+  // to catch the re-render.
+  const activeTab = useRef(tabs[activeTabIndex])
```
dzucconi, line 62: This is a little annoying but is how we can prevent scrolls from happening on arbitrary re-renders. We just need to update this ref's value before we update the `activeTabIndex` so that updates catch the render.

tam-kis (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.6.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.6.1), [`@artsy/palette@18.6.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.6.1) :rocket:

PR #1072, opened 2021-11-12
feat: Added static state to Pill

anastasiapyzhik:
If static prop is passed Pill won't be changing neither on hover nor focus.

https://user-images.githubusercontent.com/44819355/141436467-8887fc97-a68a-4e15-b9a3-eeae4408c2fd.mp4

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.story.tsx
+++ b/packages/palette/src/elements/Pill/Pill.story.tsx
@@ -11,7 +11,9 @@ export default {
 
 export const _States = () => {
   return (
-    <States<PillProps> states={[{}, { hover: true }, { focus: true }]}>
+    <States<PillProps>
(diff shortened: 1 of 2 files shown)


dzucconi:
I don't think this is the appropriate change to make but can we have some context here? It's poor UX to take something that's meant to be clickable and then pull the rug, so-to-speak.

anastasiapyzhik:
@dzucconi it will be used to display selected filters [(Figma spec)](https://www.figma.com/file/zbEFZqjKgKquZwtU8U44j9/Alerts?node-id=3475%3A6831)
On grooming it was decided to add this component to palette [(Jira ticket)](https://artsyproduct.atlassian.net/browse/FX-3543)
Do you think it's better to add a separate element, something like `Badge`?

dzucconi:
OK — yeah these don't get rid of the hover state or anything like that. We just need an example for you with an X.

PR #1076, opened 2021-11-22
feat(NX02486): adjusts the modal to the new design

araujobarret:
[NX-2486]

This PR updates the Modal to the latest changes from Design. Also implements a new idea, the shadow over sticky elements only when the content it's scrolled, getting back to scrollTop 0, doesn't show the shadow.

I kept the old narrow behavior and didn't create any breaking changes, only styling things happened here.

I wanted to create unit tests for this new hook but with Enzyme and the use case was tricky to be done, I wonder if we're going to extend RTL to Palette too, my vote is YES, definitely facilitates some neat stuff to test like hooks over Enzyme.

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/15792853/142894439-6844b8e7-c5cf-45d5-873d-5ccc51444b61.mov

</details>

<details>
<summary>Mobile</summary>

https://user-images.githubusercontent.com/15792853/142894425-5973efd5-6172-4a95-8055-32fbcd6f8ed3.mov

</details>

<details>
<summary>Hook in action Desktop/Mobile</summary>

https://user-images.githubusercontent.com/15792853/142894612-872bbad8-32f2-4872-9a70-36f52e6e10e4.mov

</details>

[NX-2486]: https://artsyproduct.atlassian.net/browse/NX-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@17.7.0-canary.1076.21932.0
  npm install @artsy/palette@18.7.0-canary.1076.21932.0
  # or 
  yarn add @artsy/palette-charts@17.7.0-canary.1076.21932.0
  yarn add @artsy/palette@18.7.0-canary.1076.21932.0
  ```
</details>

### DIFF
--- a/packages/palette/src/utils/useOnScroll.ts
+++ b/packages/palette/src/utils/useOnScroll.ts
@@ -1,0 +1,29 @@
+import { useEffect, useState, RefObject } from "react"
+
+/**
+ * useOnScroll is used to track if an element is scrolled vertically or not
+ */
+export const useOnScroll = (ref: RefObject<HTMLElement>) => {
+  const [isScrolled, setIsScrolled] = useState(false)
+
+  useEffect(() => {
+    const node = ref.current
+    if (node) {
+      const onScroll = () => {
+        if (node.scrollTop > 0) {
+          !isScrolled && setIsScrolled(true)
+        } else {
+          isScrolled && setIsScrolled(false)
+        }
+      }
+
+      node.addEventListener("scroll", onScroll, false)
+
+      return () => {
+        node.removeEventListener("scroll", onScroll, false)
+      }
+    }
+  }, [ref, isScrolled])
+
+  return { isScrolled }
+}
--- a/packages/palette/src/utils/useOnScroll.ts
+++ b/packages/palette/src/utils/useOnScroll.ts
@@ -7,21 +7,26 @@ export const useOnScroll = (ref: RefObject<HTMLElement>) => {
   const [isScrolled, setIsScrolled] = useState(false)
 
   useEffect(() => {
+    if (!("IntersectionObserver" in window)) return
+
     const node = ref.current
-    if (node) {
-      const onScroll = () => {
-        if (node.scrollTop > 0) {
-          !isScrolled && setIsScrolled(true)
-        } else {
-          isScrolled && setIsScrolled(false)
-        }
-      }
+    if (!node) return
 
-      node.addEventListener("scroll", onScroll, false)
-
-      return () => {
-        node.removeEventListener("scroll", onScroll, false)
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        setIsScrolled(!entry.isIntersecting)
+      },
+      {
+        root: null,
+        rootMargin: "0px",
+        threshold: 1.0,
       }
+    )
+
+    observer.observe(ref.current)
+
+    return () => {
+      observer.unobserve(node)
     }
   }, [ref, isScrolled])
 
--- a/packages/palette/src/utils/useOnScroll.ts
+++ b/packages/palette/src/utils/useOnScroll.ts
@@ -6,11 +6,10 @@ import { useEffect, useState, useRef } from "react"
 export const useOnScroll = () => {
   const [isScrolled, setIsScrolled] = useState(false)
   const elementRef = useRef(null)
+  const node = elementRef.current
 
   useEffect(() => {
     if (!("IntersectionObserver" in window)) return
-
-    const node = elementRef.current
     if (!node) return
 
     const observer = new IntersectionObserver(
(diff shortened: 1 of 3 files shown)


packages/palette/src/utils/useOnScroll.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useState, RefObject } from "react"
+
+/**
+ * useOnScroll is used to track if an element is scrolled vertically or not
+ */
+export const useOnScroll = (ref: RefObject<HTMLElement>) => {
```
dzucconi, line 6: I'd rather this be done with an IntersectionObserver if possible?

dzucconi:
I'd publish a canary and take a look in force to see how this sits in practice on the existing modals.

At some point we need to reconcile this Modal with ModalBase which is the correct way to do modals. This thing has a bunch of issues.

packages/palette/src/utils/useOnScroll.ts
araujobarret: I will check if it's possible 👍

araujobarret:
> I'd publish a canary and take a look in force to see how this sits in practice on the existing modals.
> 
> At some point we need to reconcile this Modal with ModalBase which is the correct way to do modals. This thing has a bunch of issues.

Thanks for the guidance here! May I ask how can we work with this Canary thing? I'm kinda curious and need to learn how to test a canary version in force before merging stuff.

dzucconi:
Generally: just add the `canary` label when opening the PR. It's added now so the next commit you push should build the canary, which will update the description of this PR when it's done.

packages/palette/src/utils/useOnScroll.ts
araujobarret: I tried but didn't succeed IntersectionObserver, from the scrollable area, even if we tweak around with `threshold` will be always marking as `isIntersecting: true`, because if we scroll we don't increase the viewing(in this case) we show more but we also hide more.
Maybe we can try to pair on it if you have time and really want to implement with IntersectionObserver 😄

dzucconi: You'd just do it with a sentinel div. We do this in Force for replicating Sticky behavior: https://github.com/artsy/force/blob/master/src/v2/Components/Sticky/Sticky.tsx#L41-L68

dzucconi: Unless I'm misunderstanding. This is just to turn on the shadow right?

araujobarret: Oh, I will try with a sentinel div, thanks for the reference!

>Unless I'm misunderstanding. This is just to turn on the shadow right?
And off when we scroll back to the initial position

araujobarret: Worked in this [commit](https://github.com/artsy/palette/pull/1076/commits/6ce31cc924738281107f5ba933ec0e7d78473b12) and in this case just observing the `isIntersecting` was enough

packages/palette/src/utils/useOnScroll.ts
```diff
@@ -7,21 +7,26 @@ export const useOnScroll = (ref: RefObject<HTMLElement>) => {
...
+    )
 
-      node.addEventListener("scroll", onScroll, false)
+    observer.observe(ref.current)
 
-      return () => {
-        node.removeEventListener("scroll", onScroll, false)
-      }
+    return () => {
+      observer.unobserve(node)
```
dzucconi, line 29: I'd also set up and return the ref inside of the hook so that in the call site you just have to pull out the ref and place it where you want to track.

araujobarret: If I got you correctly, this [commit](https://github.com/artsy/palette/pull/1076/commits/34a6633ae6e1bde716fdd80d2d182527cc0c6951) does what you asked 😄

dzucconi: Ah lol thanks

dzucconi (approved):
LGTM. Feel free to merge once you're satisfied things look good in Force.

araujobarret:
@dzucconi thanks for the valuable inputs here, I tested in force and discovered an issue with the hook, I fixed the problem with the `useEffect` dependencies, the way we use the Modal with show `false` the dependency `elementRef` is never changed, only the `elementRef.current` when the Modal is open. I'm going to merge, the results in Force are okay as far as I checked.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.8.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.8.0), [`@artsy/palette@18.8.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.8.0) :rocket:

PR #1083, opened 2021-12-06
feat(modaldialog): implements new modal

dzucconi:
So our existing modal has a _number_ of problems — accessibility and otherwise.

`ModalBase` was implemented a while ago to build a correct modal implementation without a UI to serve as a base for new modals in the application. We never got around to deprecating the dialog UI modal in here. Unfortunately there's also a `Dialog` component that needs deprecating too.

This PR implements a new modal dialog UI and deprecates the old. It also more closely matches the design spec and fixes some visual bugs.

I'll publish a canary and work through integration in Force tomorrow before merging.

https://user-images.githubusercontent.com/112297/144932229-6e638c9f-cd30-45d9-8942-865f34d48ac9.mov

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@17.10.0-canary.1083.22064.0
  npm install @artsy/palette@18.10.0-canary.1083.22064.0
  # or 
  yarn add @artsy/palette-charts@17.10.0-canary.1083.22064.0
  yarn add @artsy/palette@18.10.0-canary.1083.22064.0
  ```
</details>


packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
```diff
@@ -0,0 +1,100 @@
...
+}) => {
+  const {
+    sentinel: topSentinel,
+    isSentinelVisible: isAtTop,
+  } = useSentinelVisibility()
+
+  const {
+    sentinel: bottomSentinel,
+    isSentinelVisible: isAtBottom,
+  } = useSentinelVisibility()
```
dzucconi, line 36: There are two shadows that we independently turn on and off.

> Shadows, similar to material, only appears as content goes behind.

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
```diff
@@ -0,0 +1,100 @@
...
+          transition: "box-shadow 250ms",
+          boxShadow: isAtTop ? DROP_SHADOW : undefined,
+        }}
+      >
+        {(title || hasLogo) && (
+          <Box m={2}>
+            {hasLogo && (
+              <ArtsyLogoBlackIcon
+                display="block"
+                style={{ width: "75px", height: "26px" }}
```
dzucconi, line 60: This manually sizes the icon using with the exact aspect ratio to avoid excessive margin.

packages/palette/src/elements/ModalDialog/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./ModalDialog"
+export * from "./ModalDialogBackdrop"
+export * from "./ModalDialogDialog"
```
dzucconi, line 3: This is modular so that you could, for example, swap in a new backdrop or custom dialog.

packages/palette/src/utils/useSentinelVisibility.tsx
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }, [isSentinelVisible])
+
+  return {
+    isSentinelVisible,
+    sentinel: (
+      <Box ref={ref as any} position="relative" width="100%" height={0} />
+    ),
+  }
+}
```
dzucconi, line 41: `useOnScroll` sounds like a hook for attaching something to a scroll handler. Renamed this to be pretty explicit about what it's doing and instead of returning the ref, just return out the whole sentinel.

packages/palette/src/utils/useSentinelVisibility.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { useEffect, useState, useRef } from "react"
+import { Box } from "../elements/Box"
+
+/**
+ * Used to track if an element is scrolled vertically or not
```
dzucconi, line 5: Need to update this comment actually

packages/palette/src/elements/ModalDialog/ModalDialog.tsx
```diff
@@ -0,0 +1,34 @@
...
+  hasLogo,
+  onClose,
+  title,
+  ...rest
+}) => {
+  const [boxProps, modalProps] = splitBoxProps(rest)
+
+  return (
+    <ModalDialogBackdrop onClose={onClose} {...modalProps}>
+      <ModalDialogDialog
```
araujobarret, line 23: Since it's used internally it's not an issue, but the name sounds redundant ♻️

packages/palette/src/utils/useSentinelVisibility.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    const element = ref.current
+    if (!element) return
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        setSentinelVisibility(!entry.isIntersecting)
+      },
+      {
+        root: null,
```
araujobarret, line 22: isn't better to extend it to make it more customizable? I would say at least the threshold and margin

packages/palette/src/utils/useSentinelVisibility.tsx
araujobarret: make a lot of sense, thanks for the improvement!

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
```diff
@@ -0,0 +1,100 @@
...
+              />
+            )}
+
+            {hasLogo && title && <Spacer mt={2} />}
+
+            {title && <Text variant="lg">{title}</Text>}
+          </Box>
+        )}
+
+        <Clickable p={2} ml="auto" onClick={onClose} aria-label="Close">
```
araujobarret, line 70: this is good, all modals should have the close option by default

araujobarret (approved):
Looks good, I guess removes a lot of overhead inside the old Modal, we were overusing and shifted a lot of complex from there 💯 

Only 2 things are missing IMO:

1 - Follow up tickets to upgrade Force with the newer approach and remove the old modal from here(maybe they were already on Jira)

2 - Transitional animation between show state of the modal, from the story and the ModalDialog I couldn't see, also tested locally and it's not showing.

dzucconi:
Re: 1. I'm not overly concerned with making follow up tickets. Generally just deprecating and using the new modal in new code seems fine. 

Re: 2. That current animation was just inherited from the old design system. I don't think the animation is a blocker here but I'll consult with Jonathan on the specifics.

packages/palette/src/utils/useSentinelVisibility.tsx
dzucconi: Not really, since this isn't a generic intersection observer hook. We're not using this for lazy loading or anything like that. It's a pretty specific use-case: is the sentinel in the viewport or not.

packages/palette/src/elements/ModalDialog/ModalDialog.tsx
dzucconi: Yeah, I dislike this but am reaching for alternatives. I wish I could just take over the `Dialog` component. We have to keep these exports namespaced, regardless.

araujobarret:
Cool, I'm fine with 1.
About the 2nd point I also don't consider it a blocker, mentioned because the deprecated Modal at least has a translateY sliding up, but on dismiss has no animation AFAIR

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
damassi: Whats a good dead-clear synonym for "sentinel"

packages/palette/src/elements/ModalDialog/ModalDialog.story.tsx
```diff
@@ -0,0 +1,125 @@
...
+    <>
+      <Button onClick={() => setOpen(true)}>Open Modal</Button>
+
+      {open && (
+        <ModalDialog onClose={() => setOpen(false)} maxWidth={550} {...rest}>
+          {children ? (
+            children
+          ) : (
+            <Text variant="sm">
+              Lorem ipsum, dolor sit amet consectetur adipisicing elit. Eaque,
```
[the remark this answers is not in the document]
damassi: Did not know about `repeat` 👍

packages/palette/src/elements/ModalDialog/ModalDialog.tsx
damassi: `ModalDialogContent`? `ModalDialogWindow`?

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+          <CloseIcon fill="black100" display="block" />
+        </Clickable>
+      </Flex>
+
+      <Box
+        px={2}
+        pb={2}
+        flex={1}
+        overflow="auto"
+        style={{ WebkitOverflowScrolling: "touch" }}
```
damassi, line 82: Supposed to be caps case?

packages/palette/src/utils/useDidMount.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const useDidMount = (defaultMounted = false) => {
+  const [isMounted, toggleMounted] = useState(defaultMounted)
+
+  useEffect(() => {
+    toggleMounted(true)
+  }, [])
+
+  return isMounted
+}
```
damassi, line 11: We'll want to add this hook to the docs and perhaps follow-up in Force as we have something similar there.

packages/palette/src/utils/useSentinelVisibility.tsx
damassi: I think this has a docs entry

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
dzucconi: Sentinel is pretty clear and a commonly used word for this sort of thing: https://developers.google.com/web/updates/2016/04/intersectionobserver — this is also the nomenclature Google uses.

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
dzucconi: Yeah — that's how you go about doing vendor prefixes, weirdly.

packages/palette/src/utils/useDidMount.ts
dzucconi: 👍 Will add a docs pass now

packages/palette/src/elements/ModalDialog/ModalDialog.tsx
dzucconi: `ModalDialogContent` works for me. Will update.

packages/palette/src/elements/ModalDialog/ModalDialogDialog.tsx
damassi: Cool, lets stick with google defs 👍

packages/palette/src/elements/ModalDialog/ModalDialog.story.tsx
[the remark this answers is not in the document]
dzucconi: Ah, yeah — same!

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.11.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.11.0), [`@artsy/palette@18.11.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.11.0) :rocket:

PR #1094, opened 2022-01-09
security(colors.js): Lock version to 1.4 due to corruption

damassi:
See https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/ 

We don't import this lib directly, but it is an ephemeral dep, which still runs the risk of an automatic update. 

Screen Shot 2022-01-09 at 12 58 45 PM

### DIFF
--- a/packages/palette-charts/package.json
+++ b/packages/palette-charts/package.json
@@ -35,6 +35,10 @@
     "url": "https://github.com/artsy/palette/issues"
   },
   "homepage": "https://github.com/artsy/palette#readme",
+  "resolutions": {
+    "//": "Locking colors to 1.4.0 because 1.4.4 has been deliberately corrupted",
+    "colors": "1.4.0"
+  },
   "peerDependencies": {
     "react": "^16.2.0",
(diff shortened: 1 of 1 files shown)


packages/palette-charts/package.json
```diff
@@ -35,6 +35,10 @@
     "url": "https://github.com/artsy/palette/issues"
   },
   "homepage": "https://github.com/artsy/palette#readme",
+  "resolutions": {
+    "//": "Locking colors to 1.4.0 because 1.4.4 has been deliberately corrupted",
```
pvinis, line 39: 🙌

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.13.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.13.0), [`@artsy/palette@18.13.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.13.0) :rocket:

PR #1099, opened 2022-01-14
feat(FilterSelect): Adds new <FilterSelect> component 

damassi:
This moves the FilterSelect component over from force. Currently being used in the artwork filter under `nationality`, will also be used for `artist` (are there any other spots where this would be valuable in the filter?) and the new `/works-for-you` page. 

Will follow up with a PR with docs (for some reason something is blowing things up). 

multiselect-check

#### API: 

```tsx
<FilterSelect
  placeholder="Filter by artist name"
  query="Some query" 
  initialItemsToShow={6}
  order={[
    ["country", "name"],
    ["asc", "asc"],
  ]}
  renderItemLabel={(item) => {
    return `${item.name}, ${item.country}`;
  }}
  onChange={(state) => {
    console.log(state);
  }}
  items={[
    {
      label: "Barbara Kruger",
      value: "barbara-kruger",
      country: "American",
    },
    {
      label: "Carrie Mae Weems",
      value: "carrie-weems",
      country: "American",
    },
    {
      label: "Daniel Arsham",
      value: "daniel-asham",
      country: "American",
    },
  ]}
/>

```


packages/palette/package.json
```diff
@@ -116,6 +116,7 @@
     "@artsy/palette-tokens": "^2.6.0",
     "@seznam/compose-react-refs": "^1.0.6",
     "@styled-system/theme-get": "^5.1.2",
+    "lodash": "4.17.5",
```
damassi, line 119: This is being pulled in in a few places as a transient dep, so properly added it

dzucconi: We should not be using this in Palette

dzucconi:
I still think this is a bad idea to add to Palette for a number of reasons. Let's chat on Tuesday.

damassi:
@dzucconi - Right now we're filtering `nationalities`; next we're going to filter `artists` (which is missing). And now I need to filter a totally separate artists section on `/works-for-you` page. This component is simple and useful, and will be used in 3 spots in Force. There's no harm with it being here. It's also been substantially simplified from the nationality facet implementation. We can agree to disagree on this one 🤷

damassi:
Bigger issue is lodash, which is always very annoying not to have on hand, but makes sense to remove 👍 . Turns out we don't have lodash in the project after all.

.yarnrc
```diff
@@ -1 +1,6 @@
-unsafe-disable-integrity-migration false
\ No newline at end of file
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+unsafe-disable-integrity-migration false
+yarn-path ".yarn/releases/yarn-1.19.0.cjs"
```
damassi, line 6: For some reason i could no longer yarn add etc, due to https://github.com/yarnpkg/yarn/issues/7807. This fixed it

packages/palette/package.json
```diff
@@ -116,6 +116,10 @@
     "@artsy/palette-tokens": "^2.6.0",
     "@seznam/compose-react-refs": "^1.0.6",
     "@styled-system/theme-get": "^5.1.2",
+    "lodash.intersection": "^4.4.0",
+    "lodash.orderby": "^4.6.0",
+    "lodash.reject": "^4.6.0",
+    "lodash.uniqby": "^4.7.0",
```
damassi, line 122: @dzucconi - thoughts on this? I could go into these packages and copy in the code, or i can just pluck the modules. This seems reasonable. Unlike some I really don't like like writing stdlib-esque code from scratch 😄

dzucconi: I guess it's fine though it's still adding like ~45kb.

dzucconi: Though maybe they have some shared deps and they don't all just add up? (like each of those are exactly 12.3kb)

packages/palette/src/elements/FilterSelect/Components/FilterSelectContext.tsx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+const initialState: FilterSelectState = {
+  filteredItems: [],
+  initialItemsToShow: INITIAL_ITEMS_TO_SHOW,
+  isFiltered: false,
+  items: [],
+  onChange: (x) => x,
+  order: [
+    ["age", "name"],
```
dzucconi, line 104: Seems odd this would have opinions on fields to sort by?

damassi: ahh sorry, i copied this over from the example as i was tidying up. Lemme update

packages/palette/package.json
damassi: yeah, i bet all the primitives underneath are shared

damassi:
Merging - if there's any more comments will follow up

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.13.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.13.0), [`@artsy/palette@18.13.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.13.0) :rocket:

PR #1102, opened 2022-01-18
feat: add disabled state for Pill

dimatretyak:
### Storybook
storybook

### Documentation
https://user-images.githubusercontent.com/3513494/149978387-84ce3baa-c14e-468f-a577-c63477032d55.mp4

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.tsx
+++ b/packages/palette/src/elements/Pill/Pill.tsx
@@ -69,6 +69,10 @@ const Container = styled(Clickable)<PillProps>`
         ${states.focus}
         ${"active" in props && props.active && states.active}
       }
+
+      &:disabled {
+        pointer-events: none;
+      }
     `
   }}
 `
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Pill/Pill.tsx
```diff
@@ -69,6 +69,10 @@ const Container = styled(Clickable)<PillProps>`
         ${states.focus}
         ${"active" in props && props.active && states.active}
       }
+
+      &:disabled {
+        pointer-events: none;
+      }
```
dzucconi, line 75: It should look visibly different from a non-disabled Pill. Can we get design input on that?

dimatretyak: > It should look visibly different from a non-disabled Pill

question: if we want pill to have no visual changes with hover/focus/click effects, how would it be better to implement this? For example, we have a pill with the artist's name, nothing should happen to it when you click on it or for hover/focus effects

https://user-images.githubusercontent.com/3513494/150110807-1b629d11-bcca-411d-9051-fa7780ed1845.mp4

PR #1112, opened 2022-02-01
feature: Lodash Full

icirellik:
Fixes part of a recent bundle size regression by switching the 
pre-bundled versions of lodash functions out with the full version
so that will be tree shaken out during the force build.

~40 KiB (gzip) reduction


damassi:
Oh interesting, thanks for digging into this. Yeah i had the full lodash with this assumption, but assumed that this individual function as lib import would still be smaller.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.16.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.16.1), [`@artsy/palette@18.16.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.16.1) :rocket:

PR #1116, opened 2022-02-03
Docs updates; clean up; fixes

dzucconi:
This PR:

* Cleans up the overall docs site layout which would break at some widths.

* Adds new docs pages for un-documented components
* Updates existing docs pages to reflect current APIs/practices
* Deprecates a number of older components and removes docs associated with them
* Some minor fixes to existing components as I encountered them.

-----

Writing docs in this system is _not pleasant_. Which we should address at some point.

* Notably `Playground` is extremely buggy and will crash if you so much as look at it funny. A few components were simply un-documentable because of the shape of their props API. In such cases I just dropped a link to the Storybook page.
* Whenever the `Playground` would crash it'd take Gatsby down with it.
* When it does crash (frequently) Gatsby is _slowwww_ to boot back up.
* Frequently saves (hot reloads) would simply freeze the Chrome tab.


packages/palette/src/elements/StaticCountdownTimer/StaticCountdownTimer.tsx
```diff
@@ -11,7 +11,10 @@ import { Sans } from "../Typography"
 
 const FIVE_HOURS_IN_SECONDS = 60 * 60 * 5
 
-/** StaticCountdownTimer */
+/**
+ * StaticCountdownTimer
+ * @deprecated
```
damassi, line 16: I know we've got a hook in force now, we should mention it.

packages/palette/src/elements/TimeRemaining/TimeRemaining.tsx
```diff
@@ -10,7 +10,10 @@ function padWithZero(num: number) {
   return num.toString().padStart(2, "0")
 }
 
-/** TimeRemaining */
+/**
+ * TimeRemaining
+ * @deprecated
```
damassi, line 15: ditto

damassi (approved):
Couple minor comments, but gracias for giving this good pass 🙏 🙏

dzucconi:
I _think_ Chromatic might be having some issues. Eventually it runs (and the specs are fine) but the actual job here is timing out. I don't _think_ there's anything in this PR that should be contributing. Going to open up a simpler PR to double check.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.18.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.18.0), [`@artsy/palette@18.18.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.18.0) :rocket:

PR #1118, opened 2022-02-04
fix(ModalBase): Make SSR friendly

damassi:
Noticed that pages in force were erroring out if they include the ModalDialog, but aren't guarded against on the server.

### DIFF
--- a/packages/palette/src/elements/Modal/ModalBase.tsx
+++ b/packages/palette/src/elements/Modal/ModalBase.tsx
@@ -55,7 +55,14 @@ export const DEFAULT_MODAL_Z_INDEX = 9999
  * Low-level modal that has no opinions about layout/overlay
  * Modals content using a portal, locks scroll.
  */
-export const ModalBase: React.FC<ModalBaseProps> = ({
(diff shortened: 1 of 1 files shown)


dzucconi (approved):
Definitely — but which pages actually do this? I guess if you wanted to auto-open a modal without having to enable it via `useEffect`?

damassi:
In the new /auction/id rebuild there are some sub routes that are triggered that exist entirely within the modal, where `<ModalDialog>` is the top-most component of the route, like `auction2/sale-name/register`, or `auction2/sale-name/confirm-registration`. Everything is URL driven, and these routes can also stand alone (jumped directly to, triggering the ssr pass).

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.18.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.18.0), [`@artsy/palette@18.18.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.18.0) :rocket:

PR #1120, opened 2022-02-04
Scrolling marquee for black owned galleries project

lilyfromseattle:
[GRO-771]

After some discussion design decided they wanted the original ribbon design from below to be a scrolling marquee. 

[Artsy 3.0 Design System](https://www.figma.com/file/m6gDpKHEWDbYJyrwsVZDBr/Artsy-3.0-Design-System?node-id=6873%3A7707)

https://user-images.githubusercontent.com/5643895/152605502-e7dc88b6-da31-4e4a-b417-29d167241216.mov

[GRO-771]: https://artsyproduct.atlassian.net/browse/GRO-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/palette/src/elements/Marquee/Marquee.story.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return <Marquee marqueeText="Black Owned" noDivider />
+}
+
+export const ExtraFast = () => {
+  return <Marquee marqueeText="Black Owned" speed="5s" />
+}
+
+export const ExtraSlow = () => {
+  return <Marquee marqueeText="Black Owned" speed="20s" />
+}
```
dzucconi, line 30: Not at all blocking but you could use our `States` component to make this a little less verbose and easier to read for consumers.

https://github.com/artsy/palette/blob/03efe61df5146dbdfaebeb1b778c6827efba4ddf/packages/palette/src/elements/Button/Button.story.tsx#L15-L26

Outputs:

packages/palette/src/elements/Marquee/Marquee.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export const Marquee: React.FC<MarqueeProps> = ({
+  marqueeText,
+  noDivider = false,
+  speed = "10s",
+  color,
+  ...rest
+}) => {
+  const size: TextVariant = useThemeConfig({ v2: "small", v3: "xs" })
```
dzucconi, line 70: Totally fine to leave this — I'm of the opinion that we shouldn't even bother supporting `v2` in new components. Though I appreciate the attention to detail.

dzucconi (approved):
Looks great! Glad this worked out.

dzucconi (changes requested):
Oh — one thing I forgot; you'll have to add this to the elements/index file so that it's importable.

packages/palette/src/elements/Marquee/Marquee.tsx
```diff
@@ -0,0 +1,113 @@
...
+  variant?: keyof typeof VARIANTS
+  speed?: string
+  color?: string
+  marqueeText: string
+  noDivider?: boolean
+}
+
+export const Marquee: React.FC<MarqueeProps> = ({
+  marqueeText,
+  noDivider = false,
```
jonallured, line 65: **nitpick [non-blocking]:** prefer truthy flag names

I have a slight preference for flag names to be truthy so that you don't have to have double negatives and break your brain haha. So what if this was `divider` instead and the default was `true` and then in the actual code we'd have:

```
{divider &&
  <StyledText>...
}
```

*format via [Conventional Comments](https://conventionalcomments.org/)*

jonallured (commented):
OMG I love it!! 😝 Marquee is back baby!!

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/marquee

packages/palette/src/elements/Marquee/Marquee.tsx
dzucconi: +

lilyfromseattle: I had that originally but changed it so that you could pass `noDivider` instead of `divider = false` but happy to change it back!

dzucconi (approved):
👍

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.20.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.20.0), [`@artsy/palette@18.20.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.20.0) :rocket:

PR #1122, opened 2022-02-08
GRO-770: Added New Cultural Badges Component

The-Beez-Kneez:
This is to add in the first Cultural Badge component to support the cultural badging that came out of the 2022 Jan Hackathon.

Leadership and Marketing are interested in trying to get the cultural badges put in place for Black Owned Galleries by Feb 17th, 2022.

Added in  badge Component (example of just badge's appearance):
Screen Shot 2022-02-08 at 11 20 33 AM

Storybook for Component:
Screen Shot 2022-02-08 at 11 30 44 AM


packages/palette/src/elements/CulturalBadge/CulturalBadge.tsx
```diff
@@ -0,0 +1,20 @@
...
+}) => {
+  return (
+    <>
+      <Box backgroundColor={"black100"} width={100} mb={1}>
+        <Text variant="xs" color="white" textAlign={"center"}>
+          {badgeCategory}
+        </Text>
+      </Box>
+    </>
+  )
```
dzucconi, line 19: A few things:
* Can get rid of the outer fragment, not needed
* This can be a single `Text` tag, no need to wrap in a `Box`, `Text` accepts everything `Box` does.
* Don't hardcode the width but rather use padding. (`px`)
* The color should be `white100`
* Accept `children` instead of `badgeCategory`
* Remove the bottom margin (in general never include outer margins on a component like this, especially in palette)
* You can get rid of the interpolation `{"black100"}` becomes just `"black100"`

dzucconi: Also: let's name this `Badge`. One can imagine these being used for other things. The `Cultural` prefix is too specific.

The-Beez-Kneez: Updated.

dzucconi:
Added the `EntityHeader` updates we paired on in accbaf748bac7da6ce8ab675c8b72ee2e805479b @The-Beez-Kneez

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@17.22.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4017.22.0), [`@artsy/palette@18.22.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4018.22.0) :rocket:

PR #1129, opened 2022-02-15
Marquee QA updates

lilyfromseattle:
Made some small adjustments to the Marquee component in response to QA feedback

The component was highlighting weird (see screenshot below) after I made my changes but it worked and wasn't showing any errors so I'm assuming its just a weird vscode issue, but lmk if you see anything that doesn't look right.

Screen Shot 2022-02-15 at 4 08 03 PM

https://www.notion.so/artsy/Marquee-is-running-too-fast-needs-to-be-1-2-speed-91ae28a484a6415dbc043a5eb57165b8
https://www.notion.so/artsy/4427eddd8ba84f11b2b147dedb6f2ed0?v=061f852c191a4858a805281aefc4e155

### DIFF
--- a/packages/palette/src/elements/Marquee/Marquee.tsx
+++ b/packages/palette/src/elements/Marquee/Marquee.tsx
@@ -2,9 +2,9 @@ import React, { useState, useEffect, useRef } from "react"
 import styled, { keyframes } from "styled-components"
 import { variant } from "styled-system"
 import { TextVariant } from "@artsy/palette-tokens/dist/typography/types"
-import { useThemeConfig } from "../../Theme"
 import { Flex } from "../Flex"
 import { Box, BoxProps } from "../Box"
+import { media } from "../../helpers"
 import { Text } from "../Text"
 
 const VARIANTS = {
@@ -41,11 +41,18 @@ const StyledText = styled(Text)`
   white-space: nowrap;
 `
 
-const Inner = styled.div<{ speed: string; offset: number }>`
+const Inner = styled.div<{
+  speed: string
+  mobileSpeed: string
+  offset: number
+}>`
   animation: ${move} ${(props) => props.speed} infinite linear;
   cursor: default;
   user-select: none;
   white-space: nowrap;
+  ${media.xs`
+    animation: ${move} ${(props) => props.mobileSpeed} infinite linear;
+  `};
 `
 
 const Item = styled.div`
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/Marquee/Marquee.tsx
```diff
@@ -55,19 +62,22 @@ const Item = styled.div`
 export interface MarqueeProps extends BoxProps {
   variant?: keyof typeof VARIANTS
   speed?: string
+  mobileSpeed?: string
   color?: string
   marqueeText: string
+  textSize?: TextVariant
```
dzucconi, line 68: If we're using `TextVariant` I'd name this `textVariant`

packages/palette/src/elements/Marquee/Marquee.tsx
```diff
@@ -55,19 +62,22 @@ const Item = styled.div`
...
   divider?: boolean
 }
 
 export const Marquee: React.FC<MarqueeProps> = ({
   marqueeText,
+  textSize = "sm",
   divider = true,
-  speed = "10s",
+  speed = "30s",
+  mobileSpeed = "10s",
```
dzucconi, line 77: A better API would be to match the responsive props array style. So `speed={["10s", "30s"]}

dzucconi: It might be a bit tricky to implement but basically you'd loop over it in the `Inner` component. I can help with that if it proves ellusive.

lilyfromseattle: 👍

packages/palette/src/elements/Marquee/Marquee.tsx
lilyfromseattle: 👍

lilyfromseattle:
@dzucconi Do you have any idea why the highlighting is weird? (see above screenshot) Everything is working

dzucconi:
I think sometimes styled-components with types `styled.div<{ whatever }>` breaks syntax highlighting. It's not something to worry about though; just a bug in whatever extension is doing the highlighting.

lilyfromseattle:
What's the deal with the UI tests? Can I just ignore?

dzucconi:
The UI tests are valid — you click through via "details" and have to approve or reject any changes.

I'd like to hold off on merging this until we update the props API since it's difficult to change a shipped API; happy to pick up those changes or pair as well.

PR #1130, opened 2022-02-17
fix: tooltip/positioning improvements

dzucconi:
When more complex tooltips or dropdowns load they may change their dimensions. In which case we should attempt to reposition the element incase it escapes the bounds of the viewport. Additionally when it flips we should maintain the same offset.

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@17.24.0-canary.1130.23142.0
  npm install @artsy/palette@18.24.0-canary.1130.23142.0
  # or 
  yarn add @artsy/palette-charts@17.24.0-canary.1130.23142.0
  yarn add @artsy/palette@18.24.0-canary.1130.23142.0
  ```
</details>


dzucconi:
So for posterity, re: https://github.com/artsy/palette/pull/1130/commits/c26dfeb0088d2f28d94ac840e1a15658a5ad2391

The goal was to portal the dropdown panel because we don't want any styles of where it's inserted to pollute the panel. Generally this isn't that much of a problem because we typically will insert it parallel within a fragment but, right now in working on the styled HTML of articles, we need to insert it directly into some styled node, whether that's an `h2` or a `p` or what have you. So some styles may leak over, in particular `letter-spacing` would be a common one.

To get around this we can simply portal the panel out. But this presents a problem for focus management: we need to yield focus back to the "parent" anchor when the dropdown is closed. This wasn't a problem before when it was being inserted as a sibling.

Since it's portaled to the end of the document (or beginning either way the outcome is similar), you have this issue where when you tab off the last node in the panel your actually tabbing off the end of the page, since it's been appended to the end of the body. And once that happens you don't actually get a `keyup` that fires. This makes it impossible to yield focus back to the parent anchor.

When modals are portaled out we use the focus lock to trap focus within that pane. This is what I'm doing here but it's semi-problematic because these panels are triggered not with a click but with a hover, and typically don't include affordances like an explicit close button.

What I've noticed is that for hoverable instances like this it's common to _not_ bring them into the tab order but rather to manually manage active items using arrow keyboard navigation (up or down). So for instance: Google's Material UI menus work this way. This matches how a native select would operate.

When you're not displaying something menu-like — say, for instance when you hover an avatar on Twitter you get the additional context popdown that includes a follow button and bio — the panel is not tabbable at all. This isn't much of an accessibility problem because that information can be accessed by just navigating to the anchor.

I think the correct solution here would be to do this: remove the dropdown panel from the tab order completely and require that consumers of it manage keyboard navigation manually or only display redundant information as a 'bonus' to pointer device users.

For now, until I have a moment to implement a keyboard navigable panel for the main navbar dropdowns, I'm going with a bit of a compromise: lock focus within the pane, as if it were a modal, and if you're keyboard navigating, `esc` will yield focus back to the parent anchor. The pane, once visible has an `aria-label` that announces that it can be closed with escape.

damassi:
If i'm reading correctly, this portal solution could also fix the search dropdown in palette docs, right?

dzucconi:
@damassi — sorta; the main problem there is that they use absolute positioning so the stacking context is inherited. Like if they were positioning it with fixed it'd also be fine.

packages/palette/src/utils/usePortal.ts
```diff
@@ -17,7 +16,13 @@ export const usePortal = () => {
...
 
   const createPortal = useCallback((children: React.ReactNode) => {
-    return __createPortal__(children, appendEl.current)
+    if (typeof window === "undefined") return null
+
+    // May execute before effect runs and appendEl is set
+    const el = appendEl.current ?? document.createElement("div")
+    appendEl.current = el
+
+    return __createPortal__(children, el)
```
dzucconi, line 25: This should obviate https://github.com/artsy/palette/pull/1118/files as well

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.0.0), [`@artsy/palette@19.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.0.0) :rocket:

PR #956, opened 2021-06-25
fix: Remove react-spring

damassi:
This removes react spring from the app per [this thread](https://artsy.slack.com/archives/CA8SANW3W/p1624453974296000).

In short, react spring was causing a memory leak in our application. We've removed animation from the charts and replaced animation in modal with css, which additionally removes a boat load of complexity. 

modal

PR #1133, opened 2022-02-17
fix: remove permanent transparency from DonutChart labels

pepopowitz:
Addresses [PX-4907]

In #956, we removed react-spring, and as part of this work we removed logic that was setting the opacity of DonutChart labels from 0 to 1. The effect of this was that the labels remained at `opacity:0`, and never appeared. 

This PR removes opacity from the DonutChart. Since the previous PR removed the "flourish" from this component, I don't personally think it is worth preserving the fading-in effect on the labels. Instead, they are now permanently visible. 

It's been a while since I contributed here in palette so please let me know if I interpreted incorrectly and need to manually generate a changelog entry 😅 

## Before this PR vs After this PR

image

cc @artsy/px-devs 

[PX-4907]: https://artsyproduct.atlassian.net/browse/PX-4907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette-charts/src/elements/DonutChart/DonutChart.tsx
+++ b/packages/palette-charts/src/elements/DonutChart/DonutChart.tsx
@@ -28,7 +28,6 @@ export const DonutChart: React.FC<DonutChartProps> = ({
   margin = space(3),
 }) => {
   const [hoverIndex, setHoverIndex] = useState(-1)
-  const [labelFadeIn] = useState(false)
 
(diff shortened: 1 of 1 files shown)


damassi:
This looks good! And no worries about the changelog 👍

PR #1135, opened 2022-02-22
fix(useclickoutside): wait for stack to clear before adding event

dzucconi:
This is a weird one — https://github.com/artsy/palette/commit/bb715ff204d1604fb6b0d4179ff1df6e0cefcad2 introduced some state into the `usePosition` hook, which should be fine. But the re-render that state initiates causes it to compete with the click handler that `useClickOutside` registers. So you click to make the popover visible, but that click propagates to the useClickOutside handler which then hides the popover. The solution is to wait for the stack to clear after visibility changes before adding the handler.

As far as catching issues like this with tests: I think we need to add a Cypress test suite. We definitely need more fidelity for these very UI-centric specs and they would be much easier to write.

### DIFF
--- a/packages/palette/src/elements/Popover/Popover.tsx
+++ b/packages/palette/src/elements/Popover/Popover.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react"
+import React, { useCallback, useEffect, useState } from "react"
 import styled from "styled-components"
 import { DROP_SHADOW } from "../../helpers"
 import { isText } from "../../helpers/isText"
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.0.2), [`@artsy/palette@19.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.0.2) :rocket:

PR #1143, opened 2022-03-18
fix(filterselect): avoids calling onchange on mount

dzucconi:
This actually provides a fix for the double page view logging on `/works-for-you`. Generally speaking, when you manually implement `onChange` you want to use `useUpdateEffect` so that it's not called on mount. In this case https://github.com/artsy/force/blob/bb2c444e9f555fc58817782782df2eb6232d7ea4/src/v2/Apps/WorksForYou/WorksForYouApp.tsx#L46-L54 — `onChange` was being called on mount causing a redirect to itself, which was logging a pageview since the middleware was getting executed with the referrer change.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.1.2), [`@artsy/palette@19.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.1.2) :rocket:

damassi:
Thank you! i haven't used that hook much.

PR #1144, opened 2022-03-21
fix(tabs): cached tabs rendering

dzucconi:
So this is still a problem, sort of.

Currently the auction pages use the `Tabs` component to display the associated auction. We actually very rarely navigate from a given page to another page of the same type, which is why I think this was never noticed. When that happens though the children of the tabs change, the overall `Tabs` component gets reused without the re-render. Since the child re-renders the tab *should* update and re-render but it doesn't. Two reasons for that. One is that we put the active tab in a ref https://github.com/artsy/palette/pull/1070 to prevent re-rendering on mount to prevent it being scrolled to. Looking at this now I feel like I have a few different ideas on how to solve this but I'm having trouble getting a reliable reproduction of the original issue. (Which I'll come back to. I've included a story which reproduces the new issue.)

Regardless: what was extra confounding was that even giving the `Tabs` a `key` that changed when the parent sale changed prevented a re-render. This was deeply puzzling until I figured out that `Join` was actually blowing away our provided keys. So there's a bug fix for that in 4b251cdeb83a11066bc0c2630a4c6830a0e2f2b6.

So for the time being I'll at least be able to fix the immediate problem on the page and will circle back to resolving the tab issue completely.

### DIFF
--- a/packages/palette/src/elements/BaseTabs/BaseTabs.tsx
+++ b/packages/palette/src/elements/BaseTabs/BaseTabs.tsx
@@ -53,3 +53,5 @@ export const BaseTabs: React.ForwardRefExoticComponent<
     )
   }
 )
+
+BaseTabs.displayName = "BaseTabs"
--- a/packages/palette/src/elements/Join/Join.tsx
+++ b/packages/palette/src/elements/Join/Join.tsx
@@ -30,26 +30,18 @@ export const Join: React.FC<JoinProps> = ({ separator, children }) => {
 
   return (
     <>
-      {elements.reduce((acc, element, currentIndex) => {
-        acc.push(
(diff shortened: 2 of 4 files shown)


dzucconi:
For my future self: I believe the initial scrolling issue was on artwork pages; the page would auto-scroll to the tabbed section

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.1.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.1.3), [`@artsy/palette@19.1.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.1.3) :rocket:

PR #1149, opened 2022-03-30
GRO-836: Adding design update for gallery attributes, until further approval on designs met

The-Beez-Kneez:
Screen Shot 2022-03-30 at 11 50 35 AM

### DIFF
--- a/packages/palette/src/elements/Badge/Badge.tsx
+++ b/packages/palette/src/elements/Badge/Badge.tsx
@@ -10,12 +10,12 @@ const VARIANTS = {
     color: "black100",
   },
   defaultDark: {
-    backgroundColor: "black100",
-    color: "white100",
+    backgroundColor: "black10",
+    color: "black100",
   },
   brand: {
-    backgroundColor: "brand",
-    color: "white100",
+    backgroundColor: "black10",
+    color: "black100",
   },
 }
 


dzucconi (commented):
Rather: let's just get rid of the variants all together

The-Beez-Kneez:
Would prefer to leave them for now, as this is an ongoing design discussion. They keep going back and forth on it, and have yet to reach a solid design to represent a multi attribute scenario.

dzucconi:
You could just remove the keys for the other variants and leave the default variant then, since nothing is using them.

The-Beez-Kneez:
But if we receive details in the next few weeks for varied styling of multiple attributes, we can just use the code here.

The plan is to support multiple of these, I think it best the logic is there so we do not have to put it back in when we end up with 3 Gallery Attribute categories.

dzucconi:
I disagree — there's no value in having placeholder code. It's not hard to add real variants when we want to. We probably wouldn't even name them this.

I'd leave the `default` variant alone and just remove the other keys.

dzucconi:
(I'm not stressing about it either way though. If you feel strongly just approve the UI tests and merge.)

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.2.0), [`@artsy/palette@19.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.2.0) :rocket:

PR #1153, opened 2022-04-11
fix: add white-space: nowrap to pill component

gkartalis:
# Description

This pr addresses [FX-3883](https://artsyproduct.atlassian.net/browse/FX-3883)

- adds white-space: nowrap; to Pill component

Reasoning behind this:
Noticed that in force when you visit [Search Results for 'banksy' | Artsy](https://www.artsy.net/search?term=banksy) ,
and the width of your browser is <= 1226 the Artist Series pill is displayed like this:

55540c1e-3efd-4198-bb21-405093adb7f3

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.tsx
+++ b/packages/palette/src/elements/Pill/Pill.tsx
@@ -47,6 +47,7 @@ const Container = styled(Clickable)<PillProps>`
   justify-content: center;
   text-align: center;
   text-decoration: none;
+  white-space: nowrap;
   border: 1px solid ${themeGet("colors.black15")};
   transition: color 0.25s ease, border-color 0.25s ease,
     background-color 0.25s ease, box-shadow 0.25s ease;


dzucconi:
So to be clear what's happening here is that this occurs in cases where we use complex children instead of strings. From your screenshot it's clear that this is the case because of the existence of a count node. As you can see the string case is handled: https://github.com/artsy/palette/blob/f959a4846a1d79e20cec681624c7c2bc4ac0499d/packages/palette/src/elements/Pill/Pill.tsx#L112-L114

I do think we'd benefit from a story that replicates this problem, as it's a subtle one.

gkartalis:
@dzucconi thanks! I added also a story for Long example with children 🤔.

Let me know what you think!

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.4.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.4.1), [`@artsy/palette@19.4.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.4.1) :rocket:

PR #1161, opened 2022-05-04
feat(icon): add white Guarantee icon

tam-kis:
Removed white Guarantee icon. Using only one Guarantee icon with color props.
Screenshot 2022-05-04 at 14 20 03


dzucconi:
these shouldn't have any color. they should be fillable by any system color.

packages/palette/src/svgs/GuaranteeWhiteIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+/** GuaranteeWhiteIcon */
+export const GuaranteeWhiteIcon: React.FC<IconProps> = ({
+  title = "Guarantee White",
+  ...props
+}) => {
+  return (
+    <Icon {...props} viewBox="0 0 18 18">
+      <Title>{title}</Title>
+      <Path
+        fill="#fff"
```
dzucconi, line 13: see other icons for example. should get rid of the other version too

dzucconi (changes requested):

packages/palette/src/svgs/GuaranteeWhiteIcon.tsx
tam-kis: You mean `color(props.fill!)` , it says `color` is deprecated? Also the other one has 2 paths of 2 different colors.

dzucconi:
its fine for this purpose (not really but we dont have a choice until i rebuild the icon pipeline). u can add a class to the accent and target it via a styled tag

packages/palette/src/svgs/GuaranteeWhiteIcon.tsx
tam-kis: @dzucconi I've updated it to use one Guarantee icon with the color props.

dzucconi (approved):

packages/palette/src/svgs/GuaranteeWhiteIcon.tsx
dzucconi: Thanks!

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@18.6.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4018.6.0), [`@artsy/palette@19.6.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4019.6.0) :rocket:

PR #1163, opened 2022-05-05
fix(ssr): Fix docs ssr

damassi:
Addresses https://artsyproduct.atlassian.net/browse/WP-98

This fixes SSR rendering in our docs that fell out when we upgraded Gatsby from 2 to 3. 

cc @artsy/grow-devs

### DIFF
--- a/packages/palette-docs/package.json
+++ b/packages/palette-docs/package.json
@@ -98,7 +98,7 @@
     "gatsby-plugin-s3": "^0.3.2",
     "gatsby-plugin-sharp": "^3.11.0",
     "gatsby-plugin-sitemap": "^4.7.0",
-    "gatsby-plugin-styled-components": "^4.11.0",
+    "gatsby-plugin-styled-components": "4.0.0",
     "gatsby-remark-autolink-headers": "^4.8.0",
     "gatsby-remark-prismjs": "^5.8.0",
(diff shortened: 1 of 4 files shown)


packages/palette-docs/package.json
```diff
@@ -98,7 +98,7 @@
     "gatsby-plugin-s3": "^0.3.2",
     "gatsby-plugin-sharp": "^3.11.0",
     "gatsby-plugin-sitemap": "^4.7.0",
-    "gatsby-plugin-styled-components": "^4.11.0",
```
damassi, line 101: This looks like a semver fail: 4.0 = Gatsby 3 (version we're on); 4.11 = Gatsby 4.

packages/palette-docs/src/layouts/MainLayout.tsx
```diff
@@ -79,59 +80,62 @@ const Layout = (props) => {
...
-            display={["none", "block"]}
-            borderLeft="1px solid"
-            borderColor="black10"
-            pl={2}
-          >
-            <TableOfContents headings={headings} />
-          </Column>
-        </GridColumns>
+    <>
+      <GlobalStyles />
```
damassi, line 84: hat-tip to: https://www.ronnieho.dev/blog/styled-components/ssr-styled-components-gatsby/

PR #1166, opened 2022-05-20
fix(spinner): supports box props; respects theme

dzucconi:
Fixes an annoyance with the `Spinner` component. Can now turn off the default assumption that it should be positioned absolutely. Would like to make that default sensible (static) but, for another day.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@19.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4019.0.2), [`@artsy/palette@20.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4020.0.2) :rocket:

PR #1171, opened 2022-05-27
fix(WP-52): search navigates to correct pathname on local

rquartararo:
[WP-52](https://artsyproduct.atlassian.net/browse/WP-52)

I overrode the default DocSearch navigate behavior and replaced it with Gatsby navigate. This should make the search always navigate to the correct pathname, even when running locally.


damassi:
Awesome, tyyyy 🙏

PR #1176, opened 2022-06-06
fix(responsivebox): utilize native aspect-ratio property in place of padding-bottom hack

dzucconi:
We're seeing an issue where: when `ResponsiveBox` components are used, within a single wrapper, within the context of a CSS column, images don't load. This appears to be a Chrome specific bug which has been introduced recently. My theory is that it has to do with the `height: 0`. Either way; it's not a bad reason to use this as a motivator to upgrade to just using native `aspect-ratio`, [which is well supported](https://caniuse.com/mdn-css_properties_aspect-ratio) in all the browsers we're concerned with. A nice benefit here is that we can drop 2 extra DOM nodes per instance.

### DIFF
--- a/packages/palette/src/elements/ResponsiveBox/ResponsiveBox.story.tsx
+++ b/packages/palette/src/elements/ResponsiveBox/ResponsiveBox.story.tsx
@@ -1,6 +1,8 @@
 import React, { useEffect, useRef, useState } from "react"
+import styled from "styled-components"
 import { Box } from "../Box"
 import { Text } from "../Text"
+import { Image } from "../Image"
 import {
   ResponsiveBox,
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@19.0.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4019.0.3), [`@artsy/palette-tokens@3.0.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-tokens%403.0.1), [`@artsy/palette@20.0.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4020.0.3) :rocket:

PR #1180, opened 2022-06-08
feat: add ability to open Dropdown by click

dimatretyak:
### Description
For the [FX-4007 task](https://artsyproduct.atlassian.net/browse/FX-4007), it is necessary to add the ability to open the menu by click (not hover effect). This PR adds such an opportunity for `Dropdown` component

### Demo
https://user-images.githubusercontent.com/3513494/172599110-c740a0f0-21c0-4695-9983-faf39d524f61.mp4

### DIFF
--- a/packages/palette/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/palette/src/elements/Dropdown/Dropdown.tsx
@@ -159,8 +159,7 @@ export const Dropdown: React.FC<DropdownProps> = ({
       document.removeEventListener("keyup", handleKeyUp)
       document.removeEventListener("click", handleClick)
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [panelRef])
+  }, [panelRef, openDropdownByClick])
 
   const activeRef = useRef(false)
 
--- a/packages/palette/src/elements/Dropdown/Dropdown.tsx
+++ b/packages/palette/src/elements/Dropdown/Dropdown.tsx
@@ -42,12 +42,13 @@ export interface DropdownProps extends BoxProps {
   offset?: number
   /** Should the dropdown panel always be present in the DOM (vs removed when invisible) */
   keepInDOM?: boolean
+  openDropdownByClick?: boolean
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/Dropdown/Dropdown.tsx
```diff
@@ -132,13 +142,24 @@ export const Dropdown: React.FC<DropdownProps> = ({
...
     document.addEventListener("keydown", handleKeyDown)
     document.addEventListener("keyup", handleKeyUp)
+    document.addEventListener("click", handleClick)
 
     return () => {
       document.removeEventListener("keydown", handleKeyDown)
       document.removeEventListener("keyup", handleKeyUp)
+      document.removeEventListener("click", handleClick)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 162: Why disable this?

dimatretyak: If we remove this, we get the following warning

image

dzucconi: Theoretically if we say, enabled this prop later on, this would break. Just as a general rule we should follow exhaustive-deps.

dimatretyak: > Theoretically if we say, enabled this prop later on, this would break

Yes, but it seems to me that few people will do so, and for this reason I added `eslint-disable-next-line`. But I have nothing against your suggestion and I will specify `openDropdownByClick` in `deps

dimatretyak: Done

dzucconi (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@19.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4019.1.0), [`@artsy/palette@20.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4020.1.0) :rocket:

PR #1183, opened 2022-06-15
fix(useportal): guards on removal

dzucconi:
Wraps the teardown here in a try/catch just in case something else has removed the node.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@20.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4020.1.1), [`@artsy/palette@21.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4021.1.1) :rocket:

PR #1194, opened 2022-07-14
chore(v2): drops support for v2 completely

dzucconi:
I wanted to get rid of the v2 `Text` variants because those can still be used under any theme. So getting rid of them means I have to drop v2 in its entirety.

Opening for canary and Chromatic specs. If specs look good I'll drop the theme switcher from the stories configuration as well (which will trigger a change in every single story).


damassi (approved):
There's a few references to `THEME_V3` in force and forque, so we'll need to be sure to update that after this lands 👍

dzucconi:
@damassi I'm not deleting that yet; this just drops usage internally. I guess the 'completely' here isn't really complete.

dzucconi:
(And I'll update force where needed using the canary)

PR #1198, opened 2022-07-28
fix(BorderedRadio): remove label hover and select underline 

rquartararo:
[TX-500](https://artsyproduct.atlassian.net/browse/TX-500)

This PR removes the hover and select underline from bordered radio labels to align with the [figma](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=6726%3A3446).


dzucconi:
The design system doesn't say anything about removing the underline on the hover state. (This should also have a visual spec specified in a story if it was the case)

PR #1200, opened 2022-08-25
fix(TX-609): radio button alignment 

rquartararo:
[TX-609](https://artsyproduct.atlassian.net/browse/TX-609)

This PR fixes the alignment of the radio button so that it is in line with the main label when there is a subtitle present. 

### After

Screenshot 2022-08-25 at 11 31 29 AM

### Before
Screenshot 2022-08-25 at 12 27 33 PM

### DIFF
--- a/packages/palette/src/elements/Radio/Radio.tsx
+++ b/packages/palette/src/elements/Radio/Radio.tsx
@@ -1,5 +1,6 @@
 import React from "react"
 import styled, { css } from "styled-components"
+import { Spacer } from "../.."
 import { Flex, FlexProps } from "../../elements/Flex"
 import { isText } from "../../helpers/isText"
 import { Text } from "../Text"
@@ -73,17 +74,17 @@ export const Radio: React.FC<RadioProps> = ({
       error={error}
       {...rest}
     >
-      <RadioDot
-        disabled={disabled}
-        hover={hover}
-        focus={focus}
-        selected={selected}
-        error={error}
-        mr={1}
-      />
-
       <Flex flexDirection="column" flex={1}>
         <Flex alignItems="center" flex={1}>
+          <RadioDot
+            disabled={disabled}
+            hover={hover}
+            focus={focus}
+            selected={selected}
+            error={error}
+            mr={1}
+          />
+
           {isText(title) ? (
             <Text variant="sm-display" lineHeight={description ? undefined : 1}>
               {title}
@@ -94,9 +95,12 @@ export const Radio: React.FC<RadioProps> = ({
         </Flex>
 
         {isText(description) ? (
-          <Text variant="xs" color="black60">
-            {description}
-          </Text>
+          <Flex>
+            <Spacer ml="30px" />
+            <Text variant="xs" color="black60">
+              {description}
+            </Text>
+          </Flex>
         ) : (
           description
         )}
@@ -127,7 +131,7 @@ const Container = styled(Flex)<{
           ${RADIO_STATES.hover}
 
           // Radio
-          > div:first-of-type {
+          > div:nth-of-type(2) {
             ${props.selected
               ? RADIO_DOT_STATES.hover.selected
               : RADIO_DOT_STATES.hover.resting}
@@ -139,7 +143,7 @@ const Container = styled(Flex)<{
         ${RADIO_STATES.focus}
 
         // Radio
-        > div:first-of-type {
+        > div:nth-of-type(2) {
           ${props.selected
             ? RADIO_DOT_STATES.focus.selected
             : RADIO_DOT_STATES.focus.resting}


packages/palette/src/elements/Radio/Radio.tsx
```diff
@@ -94,9 +95,12 @@ export const Radio: React.FC<RadioProps> = ({
         </Flex>
 
         {isText(description) ? (
-          <Text variant="xs" color="black60">
-            {description}
-          </Text>
+          <Flex>
+            <Spacer ml="30px" />
```
dzucconi, line 99: I don't see this in the design? Can you highlight in the screenshot where this shows up? Nor do I think we should be using a value of `30px` here.

packages/palette/src/elements/Radio/Radio.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react"
 import styled, { css } from "styled-components"
+import { Spacer } from "../.."
```
dzucconi, line 3: Your test is failing because of this import path.

packages/palette/src/elements/Radio/Radio.tsx
```diff
@@ -139,7 +143,7 @@ const Container = styled(Flex)<{
         ${RADIO_STATES.focus}
 
         // Radio
-        > div:first-of-type {
+        > div:nth-of-type(2) {
```
dzucconi, line 146: I realize this is more or less how this was but it's a tad more brittle. Can we target by nesting the targeted component? `${RadioDot} {`

araujobarret:
Good job, but 2 things are missing:
1 - When the state is selected we still see the hover effect, from the ticket's AC this should be also fixed
2 - The radio(1st screenshot) with larger text doesn't look aligned at the top when the as the smaller(2nd screenshot)
Screenshot 2022-08-25 at 15 51 14
Screenshot 2022-08-25 at 15 51 35

packages/palette/src/elements/Radio/Radio.tsx
dzucconi: Yeah, I'd absolutely avoid that. You could use a simple CSS grid here.

dzucconi:
> @araujobarret I made a comment on the ticket bc the hover effect looked to be an intentional part of the focus effect. @dzucconi could you confirm that this is an error?

This state isn't specified in Figma. Some text on a ticket is not sufficient cause to remove it. The active&hover state would need to be specified in the design system Figma. Until that's done I would simply ignore any ticket stating otherwise!

As an aside native controls have a hover state when selected. I don't see why we would remove it.

araujobarret:
> > @araujobarret I made a comment on the ticket bc the hover effect looked to be an intentional part of the focus effect. @dzucconi could you confirm that this is an error?
> 
> This state isn't specified in Figma. Some text on a ticket is not sufficient cause to remove it. The active&hover state would need to be specified in the design system Figma. Until that's done I would simply ignore any ticket stating otherwise!
> 
> As an aside native controls have a hover state when selected. I don't see why we would remove it.

@dzucconi I see the Design System but we have no rules established there if a state overwrites/doesn't another, I got that call from design people, I don't know how to properly have it represented in Figma this case. Maybe a selected/hovered state?

dzucconi:
It's just a compound state. Nothing preventing that from being added there. Have design add it. There is a disconnect between design system governance and team product design. Just because something comes from design doesnt mean the system should be modified. (Though it can be of course!)

araujobarret:
@dzucconi gotcha and agree! Going to get back to them with this request to add the compound state 👍 

In regards to the other thing I've mentioned, the alignment of the dot, isn't supposed to be at the start?
image

dzucconi:
I suppose it seems like it should be center aligned with the first line's height.

packages/palette/src/elements/Radio/Radio.tsx
damassi: Worth noting that 30px isn't in the [spacing scale of the design system](https://palette.artsy.net/atoms/spacing/#all-spacing-values) either, so that should be a bit of a flag

damassi:
> Just because something comes from design doesnt mean the system should be modified. (Though it can be of course!)

Seconding this! We have this [Design System 3.1 Figma file](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=604%3A1770) and any change made here **has** to be reflected there (and approved internally by Product Design, aka Jimmy) before we can work it into Palette. 

This might seem overly strict, but we're trying to steady the ship here around process amongst an ever-shifting design team. Once things settle down there I think we'll have a better pipeline to make changes here, but short of that individual engineers need to keep their guard up and ensure that the Figma doc remains canonical.

araujobarret:
@rquartararo actually we don't have one with a large variant does it get glitchy when breaking the line using a medium variant?
The new offer submission flow (mockup linked in the ticket) has exactly one use case of it.

No clues on how to solve that and respect the baseline of the first line 😓

dzucconi:
You would (probably) have to step outside of CSS and measure the line height with the aid of a ref.

dzucconi:
It might be a bit weird since you only know the container and will have to reach into the child. Actually, maybe the idea of restricting the freedom here is a good one. If we have a specific set of variants in the system and restrict to that; then that would also be fine. The problem there is just going to be motivating design to codify the changes.

dzucconi:
I wouldn't add anything to Palette until it's represented in the design system Figma.

damassi:
Once it's in figma, it takes all the stress out 😄

PR #1202, opened 2022-08-29
feat: make ShowMore more flexible

mdole:
Allowing clients to update the styles on the Clickable wrapper and Text component means we can use this component in cases like https://www.figma.com/file/9DMfKOi5RuuBD2f49eB9nV/Artwork-page?node-id=1825%3A96724

### DIFF
--- a/packages/palette/src/elements/ShowMore/ShowMore.tsx
+++ b/packages/palette/src/elements/ShowMore/ShowMore.tsx
@@ -1,11 +1,15 @@
 import React, { Children, isValidElement, useState } from "react"
-import { Clickable } from "../Clickable"
-import { Text } from "../Text"
+import { Clickable, ClickableProps, splitClickableProps } from "../Clickable"
+import { Text, TextProps } from "../Text"
 
-export interface ShowMoreProps {
+export interface ShowMoreProps
+  extends ClickableProps,
+    Pick<TextProps, "variant"> {
   children: React.ReactNode
   initial?: number
   expanded?: boolean
+  hideText?: string
+  showMoreText?: string
 }
 
 export const INITIAL_ITEMS_TO_SHOW = 6
@@ -14,10 +18,15 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
   initial = INITIAL_ITEMS_TO_SHOW,
   children,
   expanded = false,
+  hideText = "Hide",
+  showMoreText = "Show more",
+  ...rest
 }) => {
   const [isExpanded, setExpanded] = useState(expanded)
   const nodes = Children.toArray(children).filter(isValidElement)
   const hasMore = nodes.length > initial
+
+  const [clickableProps, { variant }] = splitClickableProps(rest)
 
   return (
     <>
@@ -25,12 +34,10 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
 
       {hasMore && (
         <Clickable
-          mt={1}
-          textDecoration="underline"
-          textAlign="left"
           onClick={() => setExpanded((visibility) => !visibility)}
+          {...clickableProps}
         >
-          <Text variant="xs">{isExpanded ? "Hide" : "Show more"}</Text>
+          <Text variant={variant}>{isExpanded ? hideText : showMoreText}</Text>
         </Clickable>
       )}
     </>
--- a/packages/palette/src/elements/Clickable/Clickable.tsx
+++ b/packages/palette/src/elements/Clickable/Clickable.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import styled from "styled-components"
 import { compose, ResponsiveValue, system } from "styled-system"
+import { splitProps } from "../../utils/splitProps"
 import { boxMixin, BoxProps } from "../Box"
 
 const cursor = system({ cursor: true })
@@ -12,6 +13,8 @@ export type ClickableProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
     cursor?: ResponsiveValue<string>
     textDecoration?: ResponsiveValue<string>
   }
+
+const clickableMixin = compose(boxMixin, cursor, textDecoration)
 
 /**
  * Clickable is a utility component useful for wrapping things like <div>s
@@ -27,7 +30,7 @@ export const Clickable = styled.button<ClickableProps>`
   font: inherit;
   text-align: inherit;
 
-  ${compose(boxMixin, cursor, textDecoration)}
+  ${clickableMixin}
 
   &:disabled {
     cursor: default;
@@ -38,3 +41,5 @@ Clickable.defaultProps = {
   cursor: "pointer",
   type: "button",
 }
+
+export const splitClickableProps = splitProps<ClickableProps>(clickableMixin)
--- a/packages/palette/src/elements/FilterSelect/FilterSelect.tsx
+++ b/packages/palette/src/elements/FilterSelect/FilterSelect.tsx
@@ -83,7 +83,14 @@ const _FilterSelect: React.FC = () => {
           <FilterSelectResultItem key={item.value} {...item} />
         ))
       ) : (
-        <ShowMore expanded={expanded} initial={initialItemsToShow}>
+        <ShowMore
+          expanded={expanded}
+          initial={initialItemsToShow}
+          variant={"xs"}
+          textDecoration={"underline"}
+          mt={1}
+          textAlign="left"
+        >
           {itemsSorted.map((item) => {
             return <FilterSelectResultItem key={item.value} {...item} />
           })}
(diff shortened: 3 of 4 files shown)


dimatretyak (approved):
Good improvements 👍

packages/palette/src/elements/ShowMore/ShowMore.tsx
```diff
@@ -14,6 +18,8 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
   initial = INITIAL_ITEMS_TO_SHOW,
   children,
   expanded = false,
+  clickableProps = { mt: "1", textDecoration: "underline", textAlign: "left" },
+  textProps = { variant: "xs" },
```
dzucconi, line 22: This is clunky/ugly. We have a utility that lets you split off props so that you can just use `...rest` https://github.com/artsy/palette/blob/main/packages/palette/src/utils/splitProps.ts

packages/palette/src/elements/ShowMore/ShowMore.tsx
```diff
@@ -25,12 +31,10 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
 
       {hasMore && (
         <Clickable
-          mt={1}
-          textDecoration="underline"
-          textAlign="left"
```
dzucconi, line 30: You can leave these here.

packages/palette/src/elements/ShowMore/ShowMore.tsx
```diff
@@ -25,12 +31,10 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
...
           onClick={() => setExpanded((visibility) => !visibility)}
+          {...clickableProps}
         >
-          <Text variant="xs">{isExpanded ? "Hide" : "Show more"}</Text>
+          <Text {...textProps}>{isExpanded ? "Hide" : "Show more"}</Text>
```
dzucconi, line 37: Ditto; leave the defaults in place; they get overwritten by the spread.

packages/palette/src/elements/ShowMore/ShowMore.tsx
```diff
@@ -1,11 +1,15 @@
...
-import { Text } from "../Text"
+import { Clickable, ClickableProps } from "../Clickable"
+import { BaseTextProps, Text } from "../Text"
 
 export interface ShowMoreProps {
   children: React.ReactNode
   initial?: number
   expanded?: boolean
+  // Discourage overwriting the onClick handler since it's used for the show
+  // more button
```
dzucconi, line 10: You don't have to omit `onClick`; it can be augmented fine.

packages/palette/src/elements/ShowMore/ShowMore.tsx
dzucconi: (Though; I can see that being a lil odd to pull off in practice.)

dzucconi: It's less about disallowing and more about not letting the props API become a mess. As an aside this seems like a not well designed component/not really appropriate for Palette. But that's another story.

dzucconi: Like: it'd make more sense to simply remove _all_ text styling and let it inherit from its parent. Since it's not 'designed'

mdole:
Thanks for the quick review @dimatretyak @dzucconi !

> To be honest I'm not even sure if this is the right component to use for that use case, since this is used on a list and it seems like you'd want to just line-clamp it with a button to toggle off the line-clamp?

Hm yeah good thought. In this case it works nicely because we want to cap after 4 artists, not a specific number of lines of text.

I'll touch base with Kseniya about it though; artist names do vary widely in length so I could see the argument for a number of lines.

Either way, I think it'll be an improvement for this component to be more usable next time we reach for it, so I'd say worth keeping the changes.

packages/palette/src/elements/ShowMore/ShowMore.tsx
mdole: > As an aside this seems like a not well designed component/not really appropriate for Palette. But that's another story.

This got me wondering about where it's used - turns out it's not actually being consumed by any components outside of Palette (within Palette, it's used for `FilterSelect`, and I see you & @damassi [butted heads](https://github.com/artsy/palette/pull/1099) over that choice when it was introduced). Force has its [own version](https://github.com/artsy/force/blob/main/src/Components/ArtworkFilter/ArtworkFilters/ShowMore.tsx) of the component, which I assume is where this one came from in the first place.

I take your point about whether it makes sense as part of our design system. We should either make it a first-class component (basically what this PR does, though maybe it would also need more thought around use-cases), or make it inaccessible outside of Palette and let clients create their own.

IMO it's still a useful "util" component (want to hide/show a few elements easily? done), but this isn't a hill I'm prepared to die on. If you still think it's an anti-pattern, I'll update this PR to remove it from exported components and do something similar in Force.

dzucconi:
> turns out it's not actually being consumed by any components outside of Palette

If it's the case that it's just a util then I feel even more strongly that it should have no design opinions. Strip all design from it and have it inherit from its parent context. It'd be strictly a util since this doesn't have a spec in Figma.

I'd also add a story that explicitly dealt with your use-case and correct comma insertion.

I've never seen a `fooProps`-style slot interface that was necessary or made any kind of sense. I also think generally we've made the components in Palette _too_ flexible. We should be removing things rather than adding more ways to introduce inconsistencies.

packages/palette/src/elements/ShowMore/ShowMore.tsx
damassi: I'd take @dzucconi's design sys guidance over mine any day 😄 

It feels like a read more should be in palette, just because we use it all over, but yeah I agree with him that ideally it wouldn't be styled; and / or it could be more clever in that way. So far it has worked well enough tho!

packages/palette/src/elements/ShowMore/ShowMore.tsx
```diff
@@ -14,23 +16,27 @@ export const ShowMore: React.FC<ShowMoreProps> = ({
...
+  hideText = "Hide",
+  showMoreText = "Show more",
+  ...rest
 }) => {
   const [isExpanded, setExpanded] = useState(expanded)
   const nodes = Children.toArray(children).filter(isValidElement)
   const hasMore = nodes.length > initial
 
+  const [textProps] = splitTextProps(rest)
+  const [clickableProps] = splitClickableProps(rest)
```
mdole, line 28: @dzucconi how do you feel about this?

I did try an approach where `ShowMore` just had a `<div>` instead of `<Text>` so it could inherit styles without explicit props. The problem I found was that wrapping `<ShowMore>` in e.g. styled `<Text>` would also result in those styles being applied to all `children` components. 

Not the end of the world, but means anywhere this is used the children would likely have to specify their own styles to override the styles we only want applied to `ShowMore`, which is a headache.

This approach removes the styles from `ShowMore` itself while allowing parents to pass in styles without an explicit `clickableProps/textProps` interface. It's a little odd because any props shared by both sets of props get passed down to both `Text` and `Clickable` - if someone were to do something like add `mt` to the list of allowed `Text` props in `splitTextProps`, then we'd suddenly start doubling the top margin everywhere we pass one in.

dzucconi: > if someone were to do something like add mt to the list of allowed Text props in splitTextProps, then we'd suddenly start doubling the top margin everywhere we pass one in.

That makes it a no go then — you could just split off `variant` without bothering with `splitProps` and everything else goes to the container?

I'm also willing to be pragmatic if this is holding up something — but I'd really recommend setting up a story that accomplishes your use-case with the comma insertion because looking at it I think it's probably going to be more cumbersome than just some one off component in Force.

mdole: >That makes it a no go then — you could just split off variant without bothering with splitProps and everything else goes to the container?

Oh yeah, good call! Updated, that looks better and simplifies things a lot.

> I'd really recommend setting up a story that accomplishes your use-case with the comma insertion because looking at it I think it's probably going to be more cumbersome than just some one off component in Force.

Yep this PR includes it already! And yeah, this was more meant to be an improvement to the component than a time-saver. Just trying to do it the "right" way, but turns out I'm a bit out of touch with best patterns in Palette 😅 thanks for helping me navigate this.

Screen Shot 2022-08-30 at 17 40 34

dzucconi (approved):
Looks good! — but lol at this being way more work than just doing a one-off in Force without gaining any consistency benefits in Palette.

mdole:
lol yeah 🤷‍♂️ live and learn I guess

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@23.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4023.1.1), [`@artsy/palette@24.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4024.1.1) :rocket:

PR #1204, opened 2022-09-01
refactor(Select): add useForwardRef

laurabeth:
This PR updates Palette's `<Select />` to accept & leverage reference objects.

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -1,17 +1,11 @@
 import { themeGet } from "@styled-system/theme-get"
-import React, {
-  forwardRef,
-  ForwardRefExoticComponent,
-  Ref,
-  useRef,
-} from "react"
+import React, { forwardRef, ForwardRefExoticComponent, Ref } from "react"
 import styled, { css } from "styled-components"
 import { Box, BoxProps, splitBoxProps } from "../Box"
 import { Flex } from "../Flex"
 import { Text } from "../Text"
 import { Variant } from "./types"
 import { SELECT_STATES } from "./tokens"
-import composeRefs from "@seznam/compose-react-refs"
 
 export interface Option {
   value: string
@@ -56,10 +50,8 @@ export const Select: ForwardRefExoticComponent<
       onSelect,
       ...rest
     },
-    forwardedRef
+    ref
   ) => {
-    const ref = useRef<HTMLSelectElement | null>(null)
-
     const [boxProps, selectProps] = splitBoxProps(rest)
 
     return (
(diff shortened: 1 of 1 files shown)


laurabeth:
@dzucconi I'm working on a project w/ react-hook-forms which works best if form elements accept refs. I used `Button.ts` as an example to get this done, but I'm not 100% sure it's correct. If you're not otherwise occupied, mind taking a look? Thanks.

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -30,102 +36,114 @@ export interface SelectProps
...
+      required,
+      selected,
+      title,
+      variant = "default",
+      onSelect,
+      ...rest
+    },
+    forwardedRef
+  ) => {
+    const ref = useRef<HTMLSelectElement | null>(null)
```
dzucconi, line 61: This ref isn't being used

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -30,102 +36,114 @@ export interface SelectProps
...
-      {error && typeof error === "string" && (
-        <Text variant="xs" mt={0.5} color="red100">
-          {error}
-        </Text>
-      )}
-    </Box>
-  )
-}
+            <select
+              ref={composeRefs(ref, forwardedRef) as any}
```
dzucconi, line 115: You won't need this compose here; just pass the forwarded ref.

packages/palette/src/elements/Select/Select.tsx
laurabeth: fixed

packages/palette/src/elements/Select/Select.tsx
laurabeth: fixed

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -30,102 +30,112 @@ export interface SelectProps
...
-        <div>
-          {title && (
-            <Text
-              variant="xs"
-              lineHeight={
-                variant === "inline" && description === undefined
-                  ? 1
-                  : undefined
+export const Select: ForwardRefExoticComponent<
+  SelectProps & { ref?: Ref<HTMLElement> }
```
dzucconi, line 34: `Ref<HTMLSelectElement>`; I believe the Button example is looser because we change it using the `as` prop. (This is also probably fine; depending on your usecase)

dzucconi (approved):

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@23.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4023.1.2), [`@artsy/palette@24.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4024.1.2) :rocket:

PR #1206, opened 2022-09-07
feat: add ArtsyGuarantee new icons

gkartalis:
### Description

Adds three new icons for the redesigned Artwork Screen in force
- [x] SafetyIcon
- [x] SecureLockIcon
- [x] VerifyIcon

Figma spec [here](https://www.figma.com/file/9DMfKOi5RuuBD2f49eB9nV/Artwork-page?node-id=1419%3A37823)
Jira ticket [here](https://artsyproduct.atlassian.net/browse/FX-4151)


gkartalis:
@damassi @dzucconi is there a reason why we disallow the use of `fill="none"` 🤔

packages/palette/src/svgs/SafetyIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const SafetyIcon: React.SFC<IconProps> = (props) => {
+  return (
+    <Icon viewBox="0 0 24 24" fill="none" {...props}>
+      <G fillRule="evenodd" clipPath="url(#a)" clipRule="evenodd">
+        <Path
+          fill="#000"
+          d="M6.745 17.505a27.926 27.926 0 003.948 3.102 2.287 2.287 0 002.564 0 29.072 29.072 0 003.728-3.142 11.347 11.347 0 003.11-4.934.814.814 0 00-.61-.976.814.814 0 00-.977.602 9.826 9.826 0 01-2.678 4.168 26.87 26.87 0 01-3.517 2.947.635.635 0 01-.725 0 26.253 26.253 0 01-3.72-2.923 9.508 9.508 0 01-2.841-4.208 13.839 13.839 0 01-.18-3.143c0-.716.074-1.383.123-1.88.31 0 .692.065 1.099.081a7.97 7.97 0 002.442-.22 8.14 8.14 0 002.393-1.286c.408-.3.766-.594 1.042-.814.277.236.635.538 1.042.814a7.88 7.88 0 002.443 1.287.814.814 0 10.431-1.572 6.342 6.342 0 01-1.872-1.025 14.398 14.398 0 01-1.335-1.116.977.977 0 00-1.343 0c-.428.396-.876.768-1.344 1.116A6.643 6.643 0 018.08 5.417a6.325 6.325 0 01-1.954.154 15.213 15.213 0 01-1.628-.146.977.977 0 00-1.091.87c-.04.408-.147 1.474-.18 2.67-.073 1.188 0 2.38.22 3.55a11.095 11.095 0 003.298 4.99z"
+        />
+        <Path
+          fill="#1023D7"
```
dzucconi, line 14: There shouldn't be _any_ fill colors?

packages/palette/src/svgs/SecureLockIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <Icon
+      fill="none"
+      viewBox="0 0 24 24"
+      {...props}
+    >
+      <Path
+        stroke="#000"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+        strokeWidth="1.5"
```
dzucconi, line 16: Strokes should be outlined

packages/palette/src/svgs/SafetyIcon.tsx
gkartalis: hmm, I used [this guide](https://palette.artsy.net/atoms/icons/#adding-new-icons) to add the new icons, and [this](https://svg2jsx.com/) website came up with the result.

After that I just converted path -> Path, svg -> Icon and so on, got rid of some other stuff like  ```
```
xmlns="http://www.w3.org/2000/svg"
      width="24"
      height="24"
```

and this was the end result, happy to fix the rest but not sure how 😅

dzucconi: Use https://jakearchibald.github.io/svgomg/ which is linked there?

packages/palette/src/svgs/SecureLockIcon.tsx
dzucconi: The designers should be giving you outlines btw — this one isn't on us.

packages/palette/src/svgs/SafetyIcon.tsx
gkartalis: I did use that but with the default settings, so I guess I have to actually change some of the settings 🤔

dblandin: Interesting about the multi-colored element of these icons. Should we support multi-colored icons? If so, would we consider extending `IconProps` to include something like `accentFill`?

Seems like an icon that disables external colors would be less desirable?

dzucconi: The only true multi-colored icon is the Google logo, which shouldn't be altered. The few others are just white variants which are just wrong & should be removed. So I'm assuming this is a mistake? You can also simply target sub-paths with CSS if need be and manually hang some classes in the markup if you know you want an accent somewhere.

dzucconi: (And yes: our lack of an SVG pipeline is a huge problem and I'll work to resolve it when I return from leave! We can reduce the size of this package *immensely*)

gkartalis:
Closing this since two of the three icons already exist. Will open a new pr with the lock icon with strokes outlined

PR #1211, opened 2022-10-17
fix(Box): add gridArea mixin

laurabeth:
A CSSGrid wrapper component already exists within Palette. This PR adds styled-system's `GridArea` mixin to the base Box component, which will allow palette tokens to be positioned within a grid according to SS responsive layouts paradigm.


dzucconi (approved):
Fine with this.

Though perhaps we should just mixin all of `grid` then completely delete the awkward `CSSGrid` component. (Though only do that if we intend to first open a PR with it's removal in Force)

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@23.5.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4023.5.0), [`@artsy/palette@24.5.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4024.5.0) :rocket:

PR #1212, opened 2022-10-18
fix(shelf): adds label to scrollbar thumb

dzucconi:
A minor fix to address a kind of odd Lighthouse callout. We skip over this button with the tab index so the only time this gets announced is if you use a pointer device to access it.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@23.5.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4023.5.1), [`@artsy/palette@24.5.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4024.5.1) :rocket:

PR #1218, opened 2022-10-27
feat: deny change position when scrolling for `Popover` component

dimatretyak:
### Description
ToDo


packages/palette/src/elements/Popover/Popover.tsx
```diff
@@ -40,6 +41,7 @@ export const Popover: React.FC<PopoverProps> = ({
   visible: _visible = false,
   children,
   popover,
+  denyFlipPosition,
```
dzucconi, line 44: It would be better to name this the positive case 'flipPosition' and then set a default to be 'true'

dzucconi (commented):
What is the use case for this?

PR #1220, opened 2022-11-01
fix(horizontaloverflow): accommodate half pixel differences

dzucconi:
Depending on if an element is centered, the actual detected difference in scroll might be a half pixel. Here we loosen the detection to happen within 1px.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@25.0.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4025.0.1), [`@artsy/palette@26.0.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4026.0.1) :rocket:

PR #1229, opened 2022-11-26
fix: Avatar border issue on Safari

MrSltun:
This PR fixes the Avatar border-radius issue on Safari

https://user-images.githubusercontent.com/20655703/204067452-54351208-209f-49e9-b168-2ede45f0b2bb.mov

https://user-images.githubusercontent.com/20655703/204067454-1e45a2aa-7180-4c56-a5b9-dea70d0a3782.mov

### DIFF
--- a/packages/palette/src/elements/Avatar/Avatar.tsx
+++ b/packages/palette/src/elements/Avatar/Avatar.tsx
@@ -6,6 +6,12 @@ import { Flex, FlexProps } from "../Flex"
 import { Image, ImageProps } from "../Image"
 import { Text } from "../Text"
 
+/**
+ * On Safari, border-radius with overflow hidden on an image causes a rendering issue
+ * The image shows as a square at first, then the border-radius kicks in and it becomes a circle
+ * This is a css hack that solves the issue
+ * src: https://gist.github.com/ayamflow/b602ab436ac9f05660d9c15190f4fd7b
+ */
 const Container = styled(Flex)`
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/Avatar/Avatar.tsx
```diff
@@ -1,10 +1,15 @@
 import { TextVariant } from "@artsy/palette-tokens/dist/typography/v3"
 import React from "react"
+import styled from "styled-components"
 import { splitBoxProps } from "../Box"
 import { Flex, FlexProps } from "../Flex"
 import { Image, ImageProps } from "../Image"
 import { Text } from "../Text"
 
+const Container = styled(Flex)`
+  -webkit-mask-image: -webkit-radial-gradient(white, black);
```
dzucconi, line 10: Wow, would love a comment here explaining the problem and why this fixed it though!

dzucconi (approved):

packages/palette/src/elements/Avatar/Avatar.tsx
dzucconi: Interesting: https://gist.github.com/ayamflow/b602ab436ac9f05660d9c15190f4fd7b?permalink_comment_id=3941712#gistcomment-3941712 — I find this comment a bit more helpful but I'm wondering if I can also track down an actual bug report in the Webkit bug tracker.

PR #1230, opened 2022-11-30
add new banner background colour and text alignment

dariakoko:
In this PR, I add a `blue10` banner variant and  `justifyContent` property settings
Reason: we keep having [the blue10 banner with left text alignments in figma](https://www.figma.com/file/MbVDbVu5eF6PUCzjCENx4g/My-Collection-Landing-Page?node-id=723%3A32490&t=xt7xtwlElstNgzI9-0). Why not adding it to palette 

On the screenshot, the second banner shows test alignment settings and the last one - new colour

image

### DIFF
--- a/packages/palette/src/elements/Banner/Banner.story.tsx
+++ b/packages/palette/src/elements/Banner/Banner.story.tsx
@@ -16,6 +16,7 @@ export const Default = () => {
         { variant: "success" },
         { variant: "error" },
         { variant: "brand" },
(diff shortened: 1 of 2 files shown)


dzucconi:
Hi, Looks like you just want the Message component: https://palette-storybook.artsy.net/?path=/story/components-message--default

We also will only adjust these components when there is updates to the actual design system.

dariakoko:
🤦‍♀️ 
thank you

PR #1231, opened 2022-12-02
feat: remove v2

pvinis:
removing everything v2 from palette


packages/palette-tokens/src/helpers.ts
```diff
@@ -1,4 +1,5 @@
-import tokens, { Color, ColorValue, SpacingUnit } from "./index";
+import { THEME_V3 } from ".";
```
dzucconi, line 1: This will break

packages/palette/src/elements/Label/Label.tsx
```diff
@@ -1,4 +1,4 @@
-import { Text } from "@artsy/palette"
+import { Text } from "../Text"
```
dzucconi, line 1: 😬

packages/palette/src/elements/StackableBorderBox/StackableBorderBox.tsx
```diff
@@ -8,7 +8,7 @@ import { BorderBoxProps } from "../BorderBox/BorderBoxBase"
  * A stackable border box is a BorderBox that shares borders with its siblings.
  */
 export const StackableBorderBox = styled(BorderBox)<BorderBoxProps>`
-  padding: ${space(3)}px;
+  padding: ${space(4)}px;
```
dzucconi, line 11: Ugh

packages/palette/src/index.tsx
```diff
@@ -3,6 +3,7 @@ export * from "./helpers"
 export * from "./svgs"
 export * from "./Theme"
 export * from "./utils"
+export * from "./themes/types"
```
dzucconi, line 6: hm?

dzucconi:
So we need to publish a canary and then validate that Force type checks and that there's no show stopper bugs. TBH I think we might have to be a bit more tactical than just ripping it out all at once but we'll see.

packages/palette/src/Theme.tsx
```diff
@@ -52,24 +43,19 @@ export const useTheme = <T extends TTheme>() => {
...
-export const isThemeV2 = (theme: TTheme): theme is ThemeV2 => {
-  return theme.id === "v2"
-}
+// export const getThemeConfig = <T, U>(
+//   props: Record<string, any>,
+//   { v2, v3 }: { v2: T; v3: U }
+// ): U | T => {
+//   const { theme = { id: "v2" } } = props
+//   return theme.id === "v2" ? v2 : v3
+// }
```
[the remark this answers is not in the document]
pvinis: i thought ill keep it for.. v4, but I'm to to remove. you prefer to remove?

packages/palette/src/elements/Label/Label.tsx
pvinis: this was wrong, right? importing from @artsy/palette within palette?!

packages/palette/src/Theme.tsx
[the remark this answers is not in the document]
dzucconi: No need to leave commented code around. That's what git is for!

packages/palette/src/elements/StackableBorderBox/StackableBorderBox.tsx
pvinis: this comes from https://github.com/artsy/eigen/blob/main/src/palette/Theme.tsx#L246, where the designers said to take 3->4. but im cool either way.

packages/palette/src/elements/Label/Label.tsx
dzucconi: You are indeed correct. Surprised this was working at all.

packages/palette-tokens/src/helpers.ts
pvinis: why?

pvinis, line 1:
```suggestion
import { THEME_V3 } from "./index";
```

packages/palette/src/index.tsx
pvinis: some files need `Color` and `SpacingUnit` types, so I thought I'll export them. you have a better idea?

pvinis: i should clarify, files outside palette. like palette-charts

packages/palette/src/Theme.tsx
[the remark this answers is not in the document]
pvinis: love it.

packages/palette/src/index.tsx
dzucconi: How were they getting them previously?

pvinis: they were coming from the `v2.ts` files that are now gone.
For example, for `Color`:
- https://github.com/artsy/palette/blob/main/packages/palette-tokens/src/themes/v2.tsx#L373
- https://github.com/artsy/palette/blob/main/packages/palette-tokens/src/themes/index.ts#L13
- https://github.com/artsy/palette/blob/main/packages/palette-tokens/src/index.ts#L13

just very "export everything" style, and now these are at least only coming from `themes/types`, which then come from tokens. feels better.

dzucconi: But they were coming from the tokens package; where as this is in the parent palette package. I guess it's just less confusing if it's coming from the one

packages/palette-tokens/src/helpers.ts
pvinis: wow i had to fix a bunch of those. how.. how are they breaking even. feels like a bad configuration. I've never seen this break anywhere else.

packages/palette/src/index.tsx
dzucconi: Well that's the thing; everything should come from tokens since that's what gets shared. Tokens should actually be used by Eigen, for instance.

dzucconi: If it's not then we don't even need it separate. cc @damassi

dzucconi: Oh right, and charts uses it too.

dzucconi: I would just leave things as they were.

pvinis: shouldnt it be in here then? https://github.com/artsy/palette/blob/main/packages/palette-charts/package.json#L117-L124

dzucconi: Hm, probably, I guess since it's a transitive dependency of Palette it doesn't yell.

pvinis: alright. in that case, i can remove this, and let it be imported from tokens 👍

pvinis: i think in that case we should remove this file too then, kinda useless.
```
import {
  Breakpoint as BreakpointV3,
  Color as ColorV3,
  SpacingUnit as SpacingUnitV3,
} from "@artsy/palette-tokens/dist/themes/v3"

export type Color = ColorV3
export type SpacingUnit = SpacingUnitV3
export type Breakpoint = BreakpointV3
```

pvinis: preference on
```
import { SpacingUnit } from "@artsy/palette-tokens"
```
or
```
import { SpacingUnit } from "@artsy/palette-tokens/dist/themes/v3"
```
?

PR #1236, opened 2022-12-09
fix(modaldialog): addresses long titles; focus-visible close state

dzucconi:
Addressing some minor issues that arose when Q/Aing the new auth dialog.

### DIFF
--- a/packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx
+++ b/packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx
@@ -8,6 +8,8 @@ import { CloseIcon } from "../../svgs/CloseIcon"
 import { ArtsyLogoBlackIcon } from "../../svgs/ArtsyLogoBlackIcon"
 import { useSentinelVisibility } from "../../utils/useSentinelVisibility"
 import { Spacer } from "../Spacer"
+import styled from "styled-components"
+import { themeGet } from "@styled-system/theme-get"
 
 export interface ModalDialogContentProps
   extends BoxProps,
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/Text/Text.tsx
```diff
@@ -65,9 +65,17 @@ export const overflowEllipsisMixin = css`
...
+export const hyphenate = css`
+  word-wrap: break-word;
+  word-break: break-word;
+  hyphens: auto;
+`
+
 /** TextProps */
 export type TextProps = BaseTextProps &
   BoxProps & {
+    hyphenate?: boolean
```
dzucconi, line 78: New `Text` feature.

packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx
```diff
@@ -104,3 +109,13 @@ export const ModalDialogContent: React.FC<ModalDialogContentProps> = ({
     </Flex>
   )
 }
+
+const Close = styled(Clickable)`
+  transition: opacity 250ms;
```
damassi, line 114: Don't we have this stored somewhere as a CONST?

dzucconi: We don't. I don't really think we need to, unless there was some agreed upon values and guidelines for animations and transitions in the DS Figma (there probably should be) — I'm just basing values off of personal experience with what feels good. In this case it's just a focus so, actually, we should maybe not have any focus just so it's a bit more apparent.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@26.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4026.1.0), [`@artsy/palette@27.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4027.1.0) :rocket:

PR #1250, opened 2023-01-30
feat(CX-3348): Update Input components

MrSltun:
This PR resolves [CX-3348]

> [!IMPORTANT]
> This is a major version update, please take a look at **Features & Breaking Changes** section

## Description
This PR updates the input components in Palette to match the new design

## Features & Breaking Changes
- Input, LabelledInput, Select, MultiSelect, and TextArea now come with a title that floats when active/completed 
- New PhoneInput element that includes a searchable list of country codes and an input
- Input now has a new prop called `showCounter` to render the character limit under the Input

### ⚠️ Breaking Changes ⚠️
- Select element prop `variant` is no longer supported
- The initial Select value should be an empty string (`""`) to enable title animation on select
- Adding `title` to an element will add 10px top padding to the element, preventing the title from being cut

### New Input ✨
<details>
<summary>Details</summary>

#### Screenshots
| Status | Screenshot |
| -- | -- |
| default | image |
| with title | image |
| with description | image image |
| with a placeholder (no title) | image |
| with a placeholder (with title - focused) | image |
| required | image |
| hover | image |
| focused | image |
| finished typing | image |
| disabled | image |
| with error message | image |
| with character counter | image image |
| with prefix and title | image |
| with prefix and title - focused | image |
| with suffix and title | image | 
| with suffix and title - focused | image |

#### Demos of Moving Title
| Input | Video |
| -- | -- |
| Normal Input | <video src="https://user-images.githubusercontent.com/20655703/216614557-f475e0e9-bd55-406e-9e72-27f4cce7edd5.mov" /> |
| Labelled input | <video src="https://user-images.githubusercontent.com/20655703/216615036-709cc9c2-1c0c-4fc8-bb77-5400d355b423.mov" /> |
</details>

### New Phone Input ✨

<details><summary>Details</summary>

#### Screenshots
| Status | Screenshot |
| --- | --- |
| Empty | image |
| With a placeholder | image |
| Disabled | image |
| Filled | image |
| Hover | image |
| Focused | image |
| Required | image |
| With the list open | image |
| With the list open and filtered | image |
| With errors | image |

#### Demo

| Jounrey | Video |
| -- | -- |
| ⌨️ & 🖱️ | <video src="https://user-images.githubusercontent.com/20655703/233412661-97e58cc1-2632-4229-ac42-f739c2cd3774.mov" width="670" /> | 
| ⌨️ ONLY | <video src="https://user-images.githubusercontent.com/20655703/233412825-04ad00c7-db69-4047-a50f-12a24c0fa5c4.mov" /> |

</details> 

### New TextArea ✨
<details><summary>Details</summary>

#### Screenshots
| Status | Screenshot |
| -- | -- |
| default | image |
| with title | image |
| with description | image image |
| with placeholder (no title) | image |
| with placeholder (with title - focused) | image |
| required | image |
| hover | image |
| focused | image |
| finished typing | image |
| disabled | image |
| disabled (with value) | image |
| with error message | image |
| with character counter | image image |
| with character counter (over limit) | image |

</details> 

### New Select ✨

<details><summary>Details</summary>

#### Screenshots
| Status | Screenshot |
| -- | -- |
| default | image |
| default (with value) | image |
| with title | image |
| with title (with value) | image |
| with title and description | image image |
| required | image |
| hover | image |
| focused | image |
| with error message | image |
| disabled | image |
| disabled (with value) | image |

#### Demo Video
<video src="https://github.com/artsy/palette/assets/20655703/b6ae8eed-8371-437d-9ad9-7303dfef3378" />

</details> 

### New MultiSelect ✨

<details><summary>Details</summary>

#### Screenshots
| Status | Screenshot |
| -- | -- |
| default | image |
| default (with value) | image |
| with title | image |
| with title (with value) | image |
| with title and description | image image |
| required | image |
| hover | image |
| focused | image |
| with error message | image |
| disabled | image |

#### Demo Video
<video src="https://github.com/artsy/palette/assets/20655703/fa1d8df4-a2fb-4454-b54b-ce41915a99c6" />	

</details> 

[CX-3330]: https://artsyproduct.atlassian.net/browse/CX-3330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@36.0.0-canary.1250.29384.0
  npm install @artsy/palette@37.0.0-canary.1250.29384.0
  # or 
  yarn add @artsy/palette-charts@36.0.0-canary.1250.29384.0
  yarn add @artsy/palette@37.0.0-canary.1250.29384.0
  ```
</details>

[CX-3348]: https://artsyproduct.atlassian.net/browse/CX-3348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun:
And as a side-effect, LabeledInput changed to match the new design as well 😄

image

Let's hope I didn't break anything else 🌚

MounirDhahri (approved):
niiice! thanks Sultan.
@dzucconi Should we wait for the other input fields (credit card and money) to be done before releasing this to users to avoid UI conflicts ?

MrSltun:
@MounirDhahri I'm merging this to a working branch called [`in-progress/new-input`](https://github.com/artsy/palette/tree/in-progress/new-input) as @dzucconi suggested, so that we can proceed with the work and release a canary version of Palette from the working branch to test it on Force before releasing it to the public :)

Please correct me if I got it wrong! @dzucconi

MounirDhahri:
Sounds great, forgot that we spoke about it Already! let's proceed with that then once Damon approves

dzucconi:
Yeah: we can't merge this until _every_ input-like component is converted and we validate Force. (I've added the `canary` tag).

(Looking now though!)

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -14,6 +15,7 @@ export interface InputProps
     > {
   active?: boolean
   description?: string
+  descriptionVariant?: TooltipVariant
```
dzucconi, line 18: I don't think we should support a configurable variant on this. They should all be the same. Safe to remove this prop.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -44,30 +47,24 @@ export const Input: React.ForwardRefExoticComponent<
...
-                {description}
-              </Text>
-            )}
-          </>
+        {description && (
+          <Tooltip
+            pointer
+            content={description}
+            variant={descriptionVariant}
+            placement={"top-end"}
```
dzucconi, line 59: `style` No need for the braces

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -44,30 +47,24 @@ export const Input: React.ForwardRefExoticComponent<
...
+          >
+            <Text variant="xs" color="black60" textAlign="right">
+              <u>What is this?</u>
+            </Text>
```
dzucconi, line 63: 👍 The `Tooltip` component clones this child and adds `tabIndex={0}` to ensure this is keyboard focusable.

packages/palette/src/elements/Input/Input.tsx
MrSltun: yeah, that one slipped 😅😬

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -44,30 +47,24 @@ export const Input: React.ForwardRefExoticComponent<
   ) => {
     const [boxProps, inputProps] = splitBoxProps(rest)
 
+    const inputName = inputProps.name || "palette-input"
```
dzucconi, line 50: `suggestion` It's possible we want to make this a required prop rather than a fallback. I don't really think there should be a fallback value though, either way.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -105,14 +112,14 @@ const StyledInput = styled.input<StyledInputProps>`
...
-  border-radius: 0;
+  border: 1px solid ${themeGet("colors.black15")};
+  border-radius: 3px;
   transition: border-color 0.25s, color 0.25s;
   font-family: ${themeGet("fonts.sans")};
   ${systemHeight};
 
   ::placeholder {
-    transition: color 0.25s;
+    transition: color 0.25s, opacity 0.2s;
```
dzucconi, line 122: Is there a spec for the animation? As far as transition values go, etc.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -142,6 +154,25 @@ const StyledInput = styled.input<StyledInputProps>`
...
 `
+
+const StyledLabel = styled.label<StyledInputProps>`
+  position: absolute;
+  top: 50%;
+  left: 5px;
+  padding: 0 5px;
+  background-color: ${themeGet("colors.white100")};
+  transform: translateY(-50%);
+  transition: 0.2s cubic-bezier(0.65, 0.05, 0.36, 1);
```
dzucconi, line 175: Same question regarding spec. Where is this curve coming from?

dzucconi: We don't want to `transition` *all* properties; so make sure you add `transform` here.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -142,6 +154,25 @@ const StyledInput = styled.input<StyledInputProps>`
...
+        }
+      `}
     `
   }};
 `
+
+const StyledLabel = styled.label<StyledInputProps>`
+  position: absolute;
+  top: 50%;
+  left: 5px;
```
dzucconi, line 171: Hm, is there a way to get at our token value — I know `0.5` has some issues. `themeGet("space['0.5']")` maybe?

packages/palette/src/elements/Label/Label.tsx
```diff
@@ -1,4 +1,4 @@
-import { Text } from "@artsy/palette"
+import { Text } from "../Text"
```
dzucconi, line 1: 🙏

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -128,6 +135,11 @@ const StyledInput = styled.input<StyledInputProps>`
         ${INPUT_STATES.hover}
       }
 
+      &:not(:placeholder-shown) {
+        ${INPUT_STATES.completed}
```
dzucconi, line 139: Smart!

dzucconi (commented):
Looking really good!

One minor general comment I have is that we might want to `opacity` transition the placeholder in, simultaneously as the label transitions its `translation`, so that there isn't an overlap.

packages/palette/src/elements/Input/Input.tsx
dzucconi: Actually, t's totally valid, of course, to have inputs without names, for instance when they are part of a non-form component — like an autocomplete UI or something.

MrSltun: I'm concerned if we have multiple inputs on a page without names, would the labels be associated with the same input?

example:

```html
<input name="palette-input" ... />
<label for="palette-input" ...>

...

<input name="palette-input" ... />
<label for="palette-input" ...>

```

Is that even possible to have two inputs with the same name? and if so, should we add a random digit at the end of the fallback input name?

dariakoko (approved):
Great work Sultan!

packages/palette/src/elements/Input/Input.tsx
MrSltun: I don't think there are any specific specs for the animation speed, so I'll change it to be .25 to match the current speed we have 👍

packages/palette/src/elements/Input/Input.tsx
MrSltun: For the curve, I was playing around with [this tool](https://cubic-bezier.com) to be as close to [the material design guidelines](https://m3.material.io/styles/motion/easing-and-duration/applying-easing-and-duration#d4348fa7-87b9-4dda-be70-c126111fd63c) 

But for the transition, I'll specify the properties that are going to be changed :) (font size, color, left, top, etc)

dzucconi: Wondering if we can do this label position transition using *only* `transform`, actually? There's no real reason we couldn't. See: https://web.dev/animations-guide/#layout

packages/palette/src/elements/Input/Input.tsx
MrSltun: I tried multiple ways, but none were successful 😞

dzucconi: We should probably just introduce some named aliases for these, like `space.half-x` or something. Anyway; fine for now.

dzucconi: (You also could just pass these using `attrs`)

packages/palette/src/elements/Input/Input.tsx
MrSltun: I managed to update the size (using `scale`) and the position (using `translate(x , y)`), but I'm not sure how to animate the changing of the text color with transform, do you know how we could do that?

In the meantime, I'll update the transition property to only update the animating ones, not _all_

dzucconi: Nice — you *could* get clever using `opacity` since that is also one of the properties that's hardware accelerated but for now I'd continue to just transition on `color`. We might want to tag these with [`will-change`](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change), but it'd probably be good to just measure and optimize later.

I think these animations are fast enough and nothing else generally will be happening on screen as far as layout goes so I'm not anticipating this being much of a problem. Just, in general, if something can be achieved with a `transform` then we should do that. 👍

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -208,9 +208,9 @@ const StyledLabel = styled.label<StyledInputProps>`
   left: 5px;
   padding: 0 5px;
   background-color: ${themeGet("colors.white100")};
-  transform: translate(0, -50%) scale(1);
+  transform: translateY(-50%);
   transition: 0.25s cubic-bezier(0.64, 0.05, 0.36, 1);
-  transition-property: color, transform, padding;
+  transition-property: color, transform, padding, font-size;
```
MrSltun, line 215: Changed these to fix this issue from showing up

### Before
image

After: 
image

dzucconi:
Was wondering what happened to this — can you rebase though instead of the merge?

packages/palette/src/elements/Input/tokens.ts
```diff
@@ -48,8 +83,25 @@ export const INPUT_STATES: Record<State, any> = {
     ::placeholder {
       color: ${themeGet("colors.black30")};
     }
+
+    &:not(:placeholder-shown) + label {
+      color: ${themeGet("colors.black30")};
+      transform: translateY(-150%);
```
damassi, line 89: All of the animation styling in here look great (checked out [the review app](https://new-input.artsy.net/)) 👍 

I wonder if there's any way we might be able to improve the perf a bit tho. It _could_ just be my machine, but it feels like the frame rate is a bit slow (or rather, could be faster). I know translate Y pushes to the gpu, but 🤔 

@MrSltun , @dzucconi - thoughts? Maybe its just fine... not too sure. 


https://github.com/artsy/palette/assets/236943/2273bffb-d8dc-4d81-9915-dba848da7e26

dzucconi: I'd be interested to know if the color is what's slowing you down; otherwise no.

Totally fine on my M2 with 32GB of ram though 😆

damassi: Oh interesting, i just went back to the tab and its all ok! It was due to low battery on my computer. Looks perfect now. 

Ignore!

damassi: So eager for this to launch ⭐

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -76,14 +79,43 @@ export const Input: React.ForwardRefExoticComponent<
             error={!!error}
             required={required}
             height={(height ?? 50) as any}
+            name={inputName}
+            title={title}
+            suffixOffset={suffixOffset}
+            prefixOffset={prefixOffset}
+            onChange={(e) => {
+              inputProps.onChange?.(e)
+              if (inputProps.maxLength) countChars(e)
```
damassi, line 88: #superminor - 
```tsx
if (...) {
  // use brackets :) 
}
```

damassi (approved):
Nice work @MrSltun - and your CSS skills are impressive :)

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -20,6 +23,9 @@ export interface InputProps
   hover?: boolean
   required?: boolean
   title?: string
+  prefixOffset?: number
+  suffixOffset?: number
```
dzucconi, line 27: I feel strongly that we need to get rid of these props. We extend BoxProps so just use them — I'm confused by this

packages/palette/src/elements/Input/Input.tsx
dzucconi: Again: we really cannot have a default name here.

packages/palette/src/elements/Input/Input.tsx
```diff
@@ -123,11 +161,24 @@ const StyledInput = styled.input<StyledInputProps>`
...
       ${props.disabled && INPUT_STATES.disabled}
       ${props.error && INPUT_STATES.error}
+      ${!!props.prefixOffset &&
+      css`
+        padding-left: ${props.prefixOffset}px;
+      `}
+      ${!!props.suffixOffset &&
+      css`
+        padding-right: ${props.suffixOffset}px;
+      `}
```
dzucconi, line 171: I'm going to need some kind of visual explanation for the existence of this?

damassi:
One thing I noticed @MrSltun is that checking out the order flow in the review app is that the drop down doesn't seem to be selectable: 

Screenshot 2023-10-25 at 10 25 42 AM

damassi:
@dzucconi - do you have any blocking comments, or can we do a follow-up here?

dzucconi:
I would really address the prefixOffset props stuff before merging. Changing props APIs down the road is hard.

We also need to drop the default name

damassi:
No prob, just wanted to be sure 👍 I shall contain my eagerness

packages/palette/src/elements/Input/Input.tsx
MrSltun: These were added to support `LabeledInput` since we used to have a default offset the label that was only added to the right of the element


In the new designs we have two options for labels, either left or right, and since we use our `Input` for it, we need to add an offset for the input and for the label positioning if the label is on the left

This is where it is used: [code](https://github.com/artsy/palette/blob/mrsltun/CX-3330/new-input-component/packages/palette/src/elements/LabeledInput/LabeledInput.tsx#L36)

dzucconi: I'd probably just split off `pl` and `pr` though? We don't want these exposed publicly. We could also dunder name them to discourage usage.

dzucconi: Why accept arbitrary values if we know the exact contexts in which they are used?

packages/palette/src/elements/Input/Input.tsx
MrSltun: Just removed the default name :)

packages/palette/src/elements/Input/Input.tsx
MrSltun: This will be removed

packages/palette/src/elements/Input/Input.tsx
dzucconi: Cool! Does the `labelOffset` actually change? Or can it just be a boolean?

MrSltun: It's a number that adds a padding-left to the label, and that number depends on the width of the text in the `LabeledInput`

dzucconi: Ah OK — got it thanks

dzucconi (approved):

damassi:
Alright @MrSltun - thanks for all of your hard work on this, for real. Gonna merge it and we can follow up with a merge on the review app tomorrow and I can update Volt 2 first, then Forque and Volt 1.

artsyit:

PR #1254, opened 2023-02-07
feat(types): Constrain spacing unit types

damassi:
Per convo with @pvinis and mobile plat, this constrains our spacing scale numeric values to only those within the design system, and if we need to step out of it we can use "3px" style values:

```tsx
<Box m={2} /> // good 
<Box m={3} /> // type-error
<Box m="3px" /> // good  
```

Opening a canary to see how it goes. 

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@28.3.0-canary.1254.25961.0
  npm install @artsy/palette@29.3.0-canary.1254.25961.0
  # or 
  yarn add @artsy/palette-charts@28.3.0-canary.1254.25961.0
  yarn add @artsy/palette@29.3.0-canary.1254.25961.0
  ```
</details>

### DIFF
--- a/packages/palette/src/themes/types.ts
+++ b/packages/palette/src/themes/types.ts
@@ -10,5 +10,8 @@ import {
 } from "@artsy/palette-tokens/dist/themes/v3"
 
 export type Color = ColorV2 | ColorV3
-export type SpacingUnit = SpacingUnitV2 | SpacingUnitV3
+export type SpacingUnit = SpacingUnitV2 | SpacingUnitV3 | `${number}px`
 export type Breakpoint = BreakpointV2 | BreakpointV3
+
+// Constrains styled-system's spacing unit to those defined by us
+export type BaseSpaceProps = { space: Record<SpacingUnit, any> }
--- a/packages/palette/src/elements/Box/Box.tsx
+++ b/packages/palette/src/elements/Box/Box.tsx
@@ -20,6 +20,7 @@ import {
   textAlign,
   TextAlignProps,
 } from "styled-system"
+import { BaseSpaceProps } from "../../themes/types"
 import { splitProps } from "../../utils/splitProps"
 
 export interface BoxProps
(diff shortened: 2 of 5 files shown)


packages/palette/src/themes/types.ts
```diff
@@ -1,14 +1,9 @@
...
   Breakpoint as BreakpointV3,
   Color as ColorV3,
   SpacingUnit as SpacingUnitV3,
 } from "@artsy/palette-tokens/dist/themes/v3"
 
-export type Color = ColorV2 | ColorV3
-export type SpacingUnit = SpacingUnitV2 | SpacingUnitV3
-export type Breakpoint = BreakpointV2 | BreakpointV3
+export type Color = ColorV3
+export type SpacingUnit = SpacingUnitV3 | `${number}px`
```
dzucconi, line 8: If this was to actually constrain the types, then we would have only be able to use the union of v2 and v3 before, which was not the case.

damassi: Updated, had to pass into the generic slot

pvinis:
do you wanna change this type to `${number}px` too? https://github.com/artsy/palette/blob/main/packages/palette-tokens/src/themes/v3.tsx#L137

pvinis (approved):

damassi:
I'm still debating whether to change it there! Since this is mostly a palette addition, not a token thing.  What do you think @pvinis / @dzucconi ?

dzucconi:
Integrate the canary in Force and see how painful it is?

damassi:
looks good inside of palette, not sure whats up with the canary types tho... will return to this on Fri

pvinis:
is this still useful to be open? i feel like we had this already, maybe on palette-mobile? if its useful, lets keep it, but if its done on palette too, lets close this.

damassi:
Ah, this fell totally off the radar. Lemme add a ticket to the board and pick this back up when i'm back from OOO 👍

PR #1257, opened 2023-02-15
fix(TX-1071): pass the collapsed state into children

rquartararo:
[TX-1071](https://artsyproduct.atlassian.net/browse/TX-1071)

This PR does not fix the above issue, however, it will allow us to more easily set the `tabIndex` for hidden child elements. 

I've also created a story as a noticed there wasn't one. 

[TX-1071]: https://artsyproduct.atlassian.net/browse/TX-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/Collapse/Collapse.story.tsx
+++ b/packages/palette/src/elements/Collapse/Collapse.story.tsx
@@ -10,21 +10,19 @@ export default {
 }
 
 export const Default = () => {
-  const [open, setOpen] = useState(false)
+  const [openCollapse, setOpenCollapse] = useState(false)
 
   return (
-    <>
-      <States<CollapseProps> states={[{ open: false }]}>
-        <>
-          <Button onClick={() => setOpen(!open)}>
-            {open ? "Close" : "Open"}
-          </Button>
-          <Spacer y={1} />
-          <Collapse open={open}>
-            <Text>This collapse is open.</Text>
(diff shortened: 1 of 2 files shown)


dzucconi (commented):
So there's some minor changes I have with regards to how the story is set up — which are a little hard to explain so I can push something. But first: it appears that this component doesn't even control its internal state and needs to be driven by external state anyway. So we don't even need to do this, right?

(As an aside I can see why this component has been neglected and is only still hanging around on the order page. It kind of doesn't really _do_ anything in terms of UI, it just transitions some content. Seems like we should maybe just move this over to Force since it's a ❄️ on that page.)

rquartararo:
@dzucconi Yeah originally I was just going to use the external state to set the tabIndex, but I thought using the component's open state would make the code a bit more readable 🤷‍♀️. I agree that this can be moved over to force!

dzucconi:
In this case there's just no internal state, so that just winds up being misdirection.

rquartararo:
Okay so lets close this and work on getting the component into force / removing it from Palette. @AlicanAkyuz is going to take over from here as I am traveling tomorrow and will be out for two weeks. Thanks for the help!

PR #1260, opened 2023-02-22
feat: add `titleVariant` prop for modals

dimatretyak:
‼️ TODO: Add screenshots

### DIFF
--- a/packages/palette/src/elements/ModalDialog/ModalDialog.tsx
+++ b/packages/palette/src/elements/ModalDialog/ModalDialog.tsx
@@ -26,7 +26,10 @@ export const ModalDialog: React.FC<ModalDialogProps> = ({
 }) => {
   const isMounted = useDidMount({ clearCallStack: true })
 
-  const [{ width, ...boxProps }, modalProps] = splitBoxProps(rest)
+  const [
+    { width, ...boxProps },
+    { titleVariant, ...modalProps },
+  ] = splitBoxProps(rest)
 
   return (
     <ModalBase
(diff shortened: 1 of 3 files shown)


packages/palette/src/elements/ModalDialog/ModalDialog.tsx
```diff
@@ -26,7 +26,9 @@ export const ModalDialog: React.FC<ModalDialogProps> = ({
 }) => {
   const isMounted = useDidMount({ clearCallStack: true })
 
-  const [{ width, ...boxProps }, modalProps] = splitBoxProps(rest)
+  const [{ width, ...boxProps }, modalProps] = splitModalDialogContentProps(
+    rest
+  )
```
dimatretyak, line 31: maybe we don't need `splitModalDialogContentProps` here and it will be enough to do something like that. what do you think about this?

```javascript
const [
  { width, ...boxProps },
  { titleVariant, ...modalProps },
] = splitModalDialogContentProps(rest)
```

packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx
```diff
@@ -71,7 +80,7 @@ export const ModalDialogContent: React.FC<ModalDialogContentProps> = ({
                 {hasLogo && title && <Spacer y={2} />}
 
                 {title && (
-                  <Text variant="lg-display" lineClamp={6} hyphenate>
+                  <Text variant={titleVariant} lineClamp={6} hyphenate>
```
dimatretyak, line 83: Waiting for a response from the design team about this feature

dzucconi:
Yeah I'm not so sure this is what we want to do. I could see adding `title: string | JSXElement` to just customize it for a snowflake. Though I'd rather not have multiple title variants unless there's an _actual_ systematic need.

PR #1261, opened 2023-02-24
feat(popover): adds prop to disable focus management

dzucconi:
With the progressive onboarding we're kind of using these like tooltips with external triggers. This allows us to ignore the typical focus management the `Popover` does.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@28.6.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4028.6.0), [`@artsy/palette@29.6.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4029.6.0) :rocket:

PR #1267, opened 2023-03-03
Revert "Revert "fix(usefocuslock): uses a more accurate method of determining focusable""

dzucconi:
Reverts artsy/palette#1266

Reverts and patches `useFocusLock` to properly observe `disabled` attribute changes.

PR #1268, opened 2023-03-03
Revert "Revert "Revert "fix(usefocuslock): uses a more accurate method of determining focusable"""

dzucconi:
Reverts artsy/palette#1267

Sigh. It looks like this is manifesting some kind of issue where nothing is focusable. This is sporadic and only surfaced by Integrity. I think I'll just replace this with a library dedicated to focus locking instead of our own thing here.

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -122,7 +122,6 @@
     "react-lazy-load-image-component": "1.5.5",
     "react-remove-scroll": "2.5.5",
     "styled-system": "^5.1.5",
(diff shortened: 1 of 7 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@28.7.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4028.7.3), [`@artsy/palette@29.7.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4029.7.3) :rocket:

PR #1271, opened 2023-03-10
feat(TX-943): update link hover and visited colors

rquartararo:
This PR updates the link styles to reflect the new design in [Figma](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=10105%3A18647&viewport=-2947%2C-4962%2C0.88&t=8CqNEOnea3jWNq5U-0). 

I went through Volt to see if there would be any issues with these changes. I’ve identified a couple of places where links will have a “visited color” when not desired and will update them. 

Once this is done, we can use as={Link} to components in force to get these new styles where needed.

### DIFF
--- a/packages/palette/src/elements/Link/Link.tsx
+++ b/packages/palette/src/elements/Link/Link.tsx
@@ -1,3 +1,4 @@
+import { themeGet } from "@styled-system/theme-get"
 import styled from "styled-components"
 import { color } from "../../helpers"
 import { Color } from "../../Theme"
@@ -6,6 +7,7 @@ import { boxMixin, BoxProps } from "../Box"
 type UnderlineBehaviors = "default" | "hover" | "none"
 
 export interface LinkProps extends BoxProps {
+  visitedColor?: Color
   hoverColor?: Color
   noUnderline?: boolean
   underlineBehavior?: UnderlineBehaviors
@@ -29,17 +31,22 @@ const backwardsCompatCompute = (state: string, props: LinkProps) => {
 /**
  * Basic <a> tag styled with additional LinkProps
  *
- * @deprecated Do not use this component! 
- * Tip: If working on Force, please use RouterLink.  
+ * Tip: If working on Force, please use RouterLink.
  */
 export const Link = styled.a<LinkProps>`
-  color: ${color("black100")};
+  color: inherit;
   transition: color 0.25s;
   text-decoration: ${(props) => backwardsCompatCompute("normal", props)};
   &:hover {
     text-decoration: ${(props) => backwardsCompatCompute("hover", props)};
     color: ${(props) =>
-      props.hoverColor ? color(props.hoverColor) : color("black100")};
+      props.hoverColor ? color(props.hoverColor) : themeGet("colors.blue100")};
+  }
+  &:visited {
+    color: ${(props) =>
+      props.visitedColor
+        ? color(props.visitedColor)
+        : themeGet("colors.blue150")};
   }
   ${boxMixin};
 `


packages/palette/src/elements/Link/Link.tsx
```diff
@@ -29,17 +31,22 @@ const backwardsCompatCompute = (state: string, props: LinkProps) => {
 /**
  * Basic <a> tag styled with additional LinkProps
  *
- * @deprecated Do not use this component! 
- * Tip: If working on Force, please use RouterLink.  
+ * Tip: If working on Force, please use RouterLink.
  */
 export const Link = styled.a<LinkProps>`
-  color: ${color("black100")};
+  color: inherit;
```
rquartararo, line 37: Correct me if I’m wrong but if we are using as={Link}  we can’t also pass it a prop to specify the color. So I changed this to `inherit` as the design specifies that the link color should be either black100 or black60 depending on the text color of the component, but let me know if you have a better solution to this.

dzucconi: `currentColor` would be better here.

packages/palette/src/elements/Link/Link.tsx
```diff
@@ -6,6 +7,7 @@ import { boxMixin, BoxProps } from "../Box"
 type UnderlineBehaviors = "default" | "hover" | "none"
 
 export interface LinkProps extends BoxProps {
+  visitedColor?: Color
   hoverColor?: Color
   noUnderline?: boolean
```
dzucconi, line 12: IMO we should make a breaking change and actually fix this props API. None of these props make a particular sense to me.

dzucconi: Would just get rid of them all and then mixin `textDecoration` like `Clickable` does.

rquartararo: Yeah good idea we can definitely replace noUnderline and underlineBehavior with it, but does textDecoration support hover and active colors?

dzucconi: No but what would be the purpose of this component if you want to go against its styling? The goal of any of these components should be to make adhering to the style guide easy. Snowflakes can just do their own thing.

rquartararo: There are a few instances in Volt where we would want to change the visited color as the links are being used more as buttons, but I guess could just change them to styled links.

rquartararo:
looks like we use `underlineBehavior="hover"` on Pagination 🤔

dzucconi:
@rquartararo well that can just be a one-off for that component — OR if we're actually aiming for consistency, get rid of it. A question for design though.

rquartararo:
I went ahead and replaced the <Link> with a styled <a> tag for `Pagination` as Design likes it how it is, also we would still need to deal with the hover color issues. I'm going to start fixing the breaking changes in Volt and Force now.

rquartararo:
I’m having an issue with using as={Link} with `RouterLink` in Force. The styles are mixing correctly, however, it also ignores the “to” prop from RouterLink. I guess it’s now expecting an “href” to use.

dzucconi:
The types for `as` are just completely broken in Palette and Force. It should work but you'll likely have to `@ts-ignore` it. This is non-ideal but I don't think anyone is up to the challenge of upgrading our `@types` packages anytime soon to fix this.

rquartararo:
@ts-ignore removes the type error, but the link still doesn't work

dzucconi:
It might need to be the other way around: `Link as={RouterLink}`?

rquartararo:
Yeah this does the trick, but if we change the "to" to "href" to turn it into a `Link` won't that make it into the 'RouterUnawareLink`? Maybe we should rename "href" to "to" on`Link`, or add both options?

dzucconi:
I'm not following. `RouterLink` should use `to` because it's not just a string, it accepts an object configuration as well, which we sometimes use. There's a problem or?

rquartararo:
Yeah sorry was reading RouterLink wrong. It is reading as router aware. This makes it a little more confusing for devs if they should use `RouterLink` or `Link as {RouterLink}.` I can try to make it more clear in the component description.

rquartararo:
@dzucconi checked Volt-v2 and Forque for breaking changes and we are good as they don’t use this component. There are some breaking changes in Volt-v1 but I forgot that we can’t upgrade it so I guess we can deal with that when the time comes. 

Force PR is ready so let me know if all is good and I will merge.

dzucconi:
Thanks for all the diligence here. Looks good!

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@30.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4030.0.0), [`@artsy/palette@31.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4031.0.0) :rocket:

PR #1288, opened 2023-04-28
fix(range): removes labelled tuple for time being

dzucconi:
Not a super critical piece of typing, this is a TS 4.0 feature that Volt doesn't have yet and it's causing some issues with the upgrade.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@30.6.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4030.6.1), [`@artsy/palette@31.6.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4031.6.1) :rocket:

PR #1290, opened 2023-05-03
feat(CX-3394): Update text area component

MrSltun:
This PR will be merged to `mrsltun/CX-3330/new-input-component`

This PR resolves [CX-3394]

### Description
This PR updates the text area component to match the new design

### Screenshots
| Status | Screenshot |
| -- | -- |
| default | image |
| with title | image |
| with description | image image |
| with placeholder (no title) | image |
| with placeholder (with title - focused) | image |
| required | image |
| hover | image |
| focused | image |
| finished typing | image |
| disabled | image |
| disabled (with value) | image |
| with error message | image |
| with character counter | image image |
| with character counter (over limit) | image |

[CX-3339]: https://artsyproduct.atlassian.net/browse/CX-3339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CX-3394]: https://artsyproduct.atlassian.net/browse/CX-3394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/TextArea/TextArea.tsx
+++ b/packages/palette/src/elements/TextArea/TextArea.tsx
@@ -127,7 +127,7 @@ export const TextArea: React.ForwardRefExoticComponent<
         )}
 
         {error && typeof error === "string" && (
-          <Text variant="xs" color="red100" mt={0.5}>
+          <Text variant="xs" color="red100" mt={0.5} ml={1}>
             {error}
           </Text>
         )}
--- a/packages/palette/src/elements/TextArea/TextArea.tsx
+++ b/packages/palette/src/elements/TextArea/TextArea.tsx
@@ -1,9 +1,10 @@
 import { themeGet } from "@styled-system/theme-get"
 import React, { useCallback, useMemo, useState } from "react"
 import styled, { css } from "styled-components"
+import { RequiredField } from "../../shared/RequiredField"
 import { Box, BoxProps, splitBoxProps } from "../Box"
-import { Spacer } from "../Spacer"
-import { Text } from "../Text/Text"
+import { Text } from "../Text"
+import { Tooltip } from "../Tooltip"
 import { TEXTAREA_STATES } from "./tokens"
 
 export interface TextAreaProps
@@ -54,6 +55,8 @@ export const TextArea: React.ForwardRefExoticComponent<
 
     const [value, setValue] = useState(defaultValue)
 
+    const textAreaName = inputProps.name || "palette-text-area"
+
     const characterLimitExceeded = useCallback(
       (text: string) => {
         return Boolean(characterLimit && text.length > characterLimit)
@@ -80,29 +83,12 @@ export const TextArea: React.ForwardRefExoticComponent<
 
     return (
       <Box width="100%" {...boxProps}>
-        {(title || description) && (
-          <>
-            <div>
-              {title && (
-                <Text variant="xs">
-                  {title}
-                  {required && (
-                    <Box as="span" color="brand">
-                      *
-                    </Box>
-                  )}
-                </Text>
-              )}
-
-              {description && (
-                <Text variant="xs" color="black60">
-                  {description}
-                </Text>
-              )}
-            </div>
-
-            <Spacer y={0.5} />
-          </>
+        {!!description && (
+          <Tooltip pointer content={description} placement="top-end">
+            <Text variant="xs" color="black60" textAlign="right">
+              <u>What is this?</u>
+            </Text>
+          </Tooltip>
         )}
 
         <Box position="relative">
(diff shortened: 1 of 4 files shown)


packages/palette/src/elements/TextArea/TextArea.tsx
```diff
@@ -54,6 +55,8 @@ export const TextArea: React.ForwardRefExoticComponent<
 
     const [value, setValue] = useState(defaultValue)
 
+    const textAreaName = inputProps.name || "palette-text-area"
```
dzucconi, line 58: I don't think we want a default name

packages/palette/src/elements/TextArea/TextArea.tsx
```diff
@@ -193,6 +191,25 @@ const StyledTextArea = styled.textarea<StyledTextAreaProps>`
...
   }}
 `
+const StyledLabel = styled.label`
+  position: absolute;
+  top: 24px;
+  left: 6px;
+  padding: 0 5px;
+  background-color: ${themeGet("colors.white100")};
+  transform: translateY(-50%);
+  transition: 0.25s cubic-bezier(0.64, 0.05, 0.36, 1);
```
dzucconi, line 211: Where does the specification for this transition come from?

packages/palette/src/elements/TextArea/TextArea.tsx
MrSltun: I'm just using it for the label for the textarea, I believe it's useful for the screen reader according to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

dzucconi: I mean inputs should have names but we should just mimic the native API here where they are optional. There are reasons for that and having a default is not helpful for screen readers in this instance.

packages/palette/src/elements/TextArea/TextArea.tsx
dzucconi: Ah, cool — yeah. Let's make sure we get some of this documented within some motion guidelines.

packages/palette/src/elements/TextArea/TextArea.tsx
```diff
@@ -115,27 +101,33 @@ export const TextArea: React.ForwardRefExoticComponent<
             onChange={handleChange}
             defaultValue={defaultValue}
             required={required}
+            name={textAreaName}
+            title={title ? (typeof title === "string" ? title : "") : undefined}
```
olerichter00, line 105: is there a reason why we only allow titles of the type string? 

Maybe instead of restricting titles to the type "string" here, what do you think about updating the type of the `TextArea`'s `title` prop instead?

packages/palette/src/elements/TextArea/TextArea.tsx
```diff
@@ -80,29 +83,12 @@ export const TextArea: React.ForwardRefExoticComponent<
...
-                </Text>
-              )}
-            </div>
-
-            <Spacer y={0.5} />
-          </>
+        {!!description && (
+          <Tooltip pointer content={description} placement="top-end">
+            <Text variant="xs" color="black60" textAlign="right">
+              <u>What is this?</u>
```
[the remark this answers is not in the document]
MrSltun: I agree with you, I'll discuss this with the design team and see how could we implement it, I used the tooltip as a starting point for both the textarea and the input, but maybe we should allow customizability for this label

packages/palette/src/elements/TextArea/TextArea.tsx
MrSltun: The reason behind this is that the `title` prop in the HTML textarea only accepts strings, but the `title` prop in the Palette Textarea component accepts ReactNode

I didn't want to update the type for the `title` because that might cause breaking changes for the TextArea to not accept React nodes anymore, but I believe updating the type would be fine, thank you for raising this! :)

dzucconi: Yeah, it should be fine and this whole thing should be considered a breaking change so if we want to fix any interfaces to make them make more sense, now is the time to do it.

olerichter00 (approved):
I think only this comment is still open: https://github.com/artsy/palette/pull/1290/files#r1183776687

packages/palette/src/elements/TextArea/TextArea.tsx
MrSltun: Hmm, maybe instead of a name we could have a default `id`?

dzucconi: Neither, actually.

dzucconi: (The reason you can't have default `id`s is because they need to be unique per page, and synchronizing random IDs with SSR hydration is non-trivial)

Just remove the default name and we're good.

artsyit:

PR #1295, opened 2023-05-12
feat(NX-3251): add drawer element

araujobarret:
Related also to https://github.com/artsy/volt-v2/pull/424 and [NX-3251]

This PR creates a drawer element following some specs we used in upcoming features that the NX team is working on.

The idea here is based on the concept of having a provider with a hook linked to toggle the Drawer on/off. Since we don't have any specs somethings in this PR we can make them flexible once we have the official specs for the Drawer, like the flex of the Overlay and Drawer Content.

Desktop

https://github.com/artsy/palette/assets/15792853/d0e0cd80-f970-4c2d-8061-3d9fbc68964e

Mobile

https://github.com/artsy/palette/assets/15792853/8adb3f67-ca71-44f8-9414-73d2e1ae89d7

### DIFF
--- a/packages/palette/src/elements/Drawer/Drawer.tsx
+++ b/packages/palette/src/elements/Drawer/Drawer.tsx
@@ -1,0 +1,98 @@
+import React, { FC } from "react"
+import { Box, Flex, breakpoints } from "@artsy/palette"
+import styled, { css } from "styled-components"
+import { zIndex } from "styled-system"
+import { useDrawer } from "./useDrawer"
+
+export interface DrawerProps {
+  anchor?: "left" | "right"
+  flexContent?: number
+  flexOverlay?: number
+}
+
+export const Drawer: FC<DrawerProps> = ({
+  children,
+  anchor = "right",
+  flexContent = 1,
+  flexOverlay = 3,
+}) => {
+  const { isOpen, toggle } = useDrawer()
+  const activeClass = isOpen ? "active" : ""
+
+  return (
+    <Container
+      position="fixed"
+      zIndex={9999}
+      className={activeClass}
+      anchor={anchor}
+      isOpen={isOpen}
+    >
+      <Box
+        flex={[1, flexContent]}
+        height="100%"
+        backgroundColor="white100"
+        overflowY="scroll"
+        className={activeClass}
+      >
+        {children}
+      </Box>
+
+      <Overlay
+        backgroundColor="black100"
+        height="100%"
+        display={["none", "flex"]}
+        flex={flexOverlay}
+        onClick={toggle}
+        className={activeClass}
+        data-testid="drawer-overlay"
+      />
+    </Container>
+  )
+}
+
+const Container = styled(Flex)<DrawerProps & { isOpen: boolean }>`
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  transition-property: transform;
+  transition-timing-function: cubic-bezier(
+    0.075,
+    0.82,
+    0.165,
+    1
+  ); /* easeOutCirc */
+  align-items: center;
+  justify-content: center;
+  ${zIndex}
+
+  ${(props) => {
+    return css`
+      transform: translateX(${props.anchor === "left" ? -100 : 100}%);
+      flex-direction: ${props.anchor === "left" ? "row" : "row-reverse"};
+      transition-duration: ${props.isOpen ? "300ms" : "6000ms"};
+      transition-delay: ${props.isOpen ? 0 : 100}ms;
+
+      /* If the screen width is not too big a shorter delay is better */
+      @media (max-width: ${breakpoints.sm}) {
+        transition-duration: ${props.isOpen ? "300ms" : "600ms"};
+      }
+    `
+  }}
+
+  &.active {
+    transform: translateX(0);
+  }
+`
+
+const Overlay = styled(Box)`
+  opacity: 0;
+  transition: opacity 50ms ease;
+  pointer-events: none;
+
+  &.active {
+    opacity: 0.5;
+    pointer-events: auto;
+    transition: opacity 300ms ease 200ms;
+  }
+`
--- a/packages/palette/src/elements/Drawer/DrawerProvider.tsx
+++ b/packages/palette/src/elements/Drawer/DrawerProvider.tsx
@@ -1,0 +1,29 @@
+import React, { FC, useCallback, useState, createContext } from "react"
+
+export const DrawerContext = createContext<{
+  isOpen: boolean
+  toggle: () => void
+  close: () => void
+}>({
+  isOpen: false,
+  toggle: () => {
+    // NOOP
+  },
+  close: () => {
+    // NOOP
+  },
+})
+
+export const DrawerProvider: FC = ({ children }) => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  const toggle = useCallback(() => setIsOpen((prev) => !prev), [])
+
+  const close = useCallback(() => setIsOpen(false), [])
+
+  return (
+    <DrawerContext.Provider value={{ isOpen, toggle, close }}>
+      {children}
+    </DrawerContext.Provider>
+  )
+}
--- a/packages/palette/src/elements/Drawer/Drawer.story.tsx
+++ b/packages/palette/src/elements/Drawer/Drawer.story.tsx
@@ -1,0 +1,72 @@
+import React from "react"
+import { Drawer } from "./Drawer"
+import { DrawerProvider } from "./DrawerProvider"
+import { useDrawer } from "./useDrawer"
+import { Flex } from "../Flex"
+import { Button } from "../Button"
+
+export default {
+  title: "Components/Drawer",
+  parameters: {
+    layout: "fullscreen",
+  },
+  decorators: [
+    (Story) => (
+      <DrawerProvider>
+        <Story />
+      </DrawerProvider>
+    ),
+  ],
+}
+
+const Layout: React.FC = ({ children }) => {
+  const { toggle } = useDrawer()
+
+  return (
+    <>
+      {children}
+
+      <Flex width="100%" height="100%" flexDirection={["column", "row"]}>
+        <Flex flex={1} />
+        <Flex flex={2} flexDirection="column">
+          <Button size="small" onClick={toggle}>
+            Open Drawer
+          </Button>
+        </Flex>
+        <Flex flex={1} />
+      </Flex>
+    </>
+  )
+}
+
+const Content = () => {
+  const { toggle } = useDrawer()
+
+  return (
+    <Flex justifyContent="center" py={2}>
+      <Button size="small" onClick={toggle}>
+        Close Drawer
+      </Button>
+    </Flex>
+  )
+}
+
+export const Default = () => {
+  return (
+    <Layout>
+      <Drawer>
+        <Content />
+      </Drawer>
+    </Layout>
+  )
+}
+
+export const Left = () => {
+  return (
+    <Layout>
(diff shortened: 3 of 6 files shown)


packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,98 @@
...
+  ); /* easeOutCirc */
+  align-items: center;
+  justify-content: center;
+  ${zIndex}
+
+  ${(props) => {
+    return css`
+      transform: translateX(${props.anchor === "left" ? -100 : 100}%);
+      flex-direction: ${props.anchor === "left" ? "row" : "row-reverse"};
+      transition-duration: ${props.isOpen ? "300ms" : "6000ms"};
```
araujobarret, line 73: I'm not an animation ninja, but if I use a duration like 300ms when it's not open it goes stupidity fast, any insights why are more than welcome

dzucconi: I _really_ feel strongly that designers should be designing this.

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { FC } from "react"
+import { Box, Flex, breakpoints } from "@artsy/palette"
+import styled, { css } from "styled-components"
+import { zIndex } from "styled-system"
+import { useDrawer } from "./useDrawer"
+
+export interface DrawerProps {
+  anchor?: "left" | "right"
+  flexContent?: number
+  flexOverlay?: number
```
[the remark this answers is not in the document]
araujobarret: any recommendation for this proportion case? Or we can hardcode the flex value proportion until we have a DS official size?

dzucconi (commented):
An important consideration here is accessibility. Let me see if I can dig up the aria pattern for this kind of thing. At a minimum we need to be managing focus: trapping it within this layer when it's open, and yielding it back to the trigger when it's closed. See the `ModalBase` for instance.

packages/palette/src/elements/Drawer/DrawerProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const DrawerProvider: FC = ({ children }) => {
+  const [isOpen, setIsOpen] = useState(false)
+
+  const toggle = useCallback(() => setIsOpen((prev) => !prev), [])
+
+  const close = useCallback(() => setIsOpen(false), [])
+
+  return (
+    <DrawerContext.Provider value={{ isOpen, toggle, close }}>
```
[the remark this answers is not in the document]
araujobarret: the idea here is more like the toaster, to enable the drawer to be called from multiple components without many declarations, but I can change it to be like the Modal style. I'll poke around to see how it works without it 👍

dzucconi:
Hm, [no 'official' guidance](https://github.com/w3c/aria-practices/issues/1590), though this is basically a modal with a different shape. [Chakra calls out the same things](https://chakra-ui.com/docs/components/drawer#accessibility) regarding focus management at least.

packages/palette/src/elements/Drawer/Drawer.tsx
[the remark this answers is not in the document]
dzucconi: I'd imagine the size is entirely dependent on the content? So that might change depending on use case. We're going to be using this for the filters for instance, which will be pretty specific. Just setting defaults that are over-writable with `BoxProps` is the way to go. I just read these props and have no idea what they mean really. We could have specific `contentProps` and `overlayProps` slots that each accept `BoxProps` is also fine here.

araujobarret:
I did the requested changes/ideas [here](https://github.com/artsy/palette/pull/1295/commits/07e49023e9ffb09a19c736001ecda7d9bdf49ab1), summary of what I did:
- User react-focus-on wrapping the drawer content
- Removed Context API(using similar approach of ModalBase now)
- simplified the animation using animation API instead of transitioning
- no out animation(we're not caring about it too much right now in Palette)

Result:

https://github.com/artsy/palette/assets/15792853/cef2fc1b-87fc-4097-b397-bc832380f0e2

If everything is fine here, should I drop a message in DS channel asking for a Designer to help craft the animation, maybe merging here, asking it to them, and having a follow-up PR with the more accurate animation?

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  animation-duration: 150ms;
+  animation-name: ${fadeIn};
+  animation-timing-function: ease-in-out;
+`
+
+const Focus = styled(FocusOn)`
+  height: 100%;
+  align-items: center;
+  justify-content: center;
+`
```
damassi, line 131: The animation CSS / props are spread over a pretty wide surface area in this file and i think it can be simplified considerably. Naively thinking this through, I could see the below semi-pseudocode being all that's needed:

```tsx
const Container = styled(Box)`
  opacity: 0; 
  right: -500px;
  
  .open {
    opacity: 1;
    right: 0;
    transition: ease-out; 
    transition-duration: 300ms; 
  }
`

Something like this; if state is to open, just add an open class. Or to avoid classNames, pass the open prop into the component and use the styled-components `css` helper. But yah, i'd try to get the LOC count in here down to the bare essentials.

File: packages/palette/src/elements/Drawer/Drawer.tsx [hunk 5]
```diff
@@ -0,0 +1,131 @@
...
+        display={["none", "flex"]}
+        onClick={onClose}
+        data-testid="drawer-overlay"
+        width="inherit"
+      />
+    </Container>
+  )
+}
+
+const DEFAULT_DRAWER_Z_INDEX = 9999
```
dzucconi, line 53: This should just be like 1

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,131 @@
...
+    <Container zIndex={zIndex} anchor={anchor}>
+      <Focus onClickOutside={onClose}>
+        <Content
+          backgroundColor="white100"
+          height="100%"
+          width={["100%", "auto"]}
+          overflowX="hidden"
+          overflowY="scroll"
+          anchor={anchor}
+          zIndex={zIndex}
```
dzucconi, line 35: Doesn't need z-index here

packages/palette/src/elements/Drawer/Drawer.tsx
damassi: Yah `transform: translateX(...)`, something on the gpu. `right` gets the idea across 😄

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  return (
+    <Container zIndex={zIndex} anchor={anchor}>
+      <Focus onClickOutside={onClose}>
+        <Content
+          backgroundColor="white100"
+          height="100%"
+          width={["100%", "auto"]}
+          overflowX="hidden"
+          overflowY="scroll"
```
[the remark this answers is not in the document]
araujobarret: done in https://github.com/artsy/palette/pull/1295/commits/657df35bf582513603b7d209058a22b6a3093a62

packages/palette/src/elements/Drawer/Drawer.tsx
araujobarret: I thought we'd follow a similar pattern as we have in the [BaseModal](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Modal/ModalBase.tsx#L53-L56), but I changed in https://github.com/artsy/palette/pull/1295/commits/657df35bf582513603b7d209058a22b6a3093a62

dzucconi: Good observation that that comment probably needs more context and less snark! At the time I was migrating legacy modals and that was what the value was at the time.

araujobarret: Ah, I should have guessed something came up during past migrations, but no worries from my side, default 1 seems to be reasonable 👍

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,105 @@
...
+  -webkit-overflow-scrolling: touch;
+
+  ${(props) => css`
+    ${props.anchor === "left" ? "left: 0;" : "right: 0;"}
+    transform: translateX(${props.anchor === "left" ? "-100%" : "100%"});
+  `};
+
+  &.open {
+    transform: translateX(0);
+  }
```
damassi, line 89: Why do we need to transform both container and content? Can we just animate one single thing?

araujobarret: That's a good catch! I was supposed to remove the `transform` from the Container 👍

araujobarret: changed in https://github.com/artsy/palette/pull/1295/commits/d34d6525758407f332ba51f43422e0ad35784ce0

araujobarret:
@dzucconi using `react-focus-on` I noticed we have by default `autoFocus={true}`, do we really want it always like this or do you think it's better to expose it throughout the Drawer API?

dzucconi:
Generally speaking we do want to focus the first item in the drawer (modal) once it's opened. Hard to imagine a case where we wouldn't want this?

araujobarret:
I share the same feeling, was more to check if I'm missing something in the API, thanks 👍

damassi (approved):

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <Focus onClickOutside={onClose} enabled={open} onEscapeKey={onClose}>
+        <Content
+          backgroundColor="white100"
+          height="100%"
+          width={["100%", "auto"]}
+          overflowX="hidden"
+          overflowY="scroll"
+          anchor={anchor}
+          zIndex={zIndex}
+          className={openClass}
```
dzucconi, line 34: Question: why target this using a `className`? You can use `${Content} {`

araujobarret: No reasoning, guilty of being away from pre/post-processed CSS too much time that forgot to use many features it has 👍 
changed on https://github.com/artsy/palette/pull/1295/commits/842e521adcd786313b04c02df8c367c06fe699e5 (I also had to change the order of declarations to comply with linter rules)

packages/palette/src/elements/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,103 @@
...
+  children,
+  anchor = "right",
+  zIndex = DEFAULT_DRAWER_Z_INDEX,
+  open,
+  onClose,
+}) => {
+  const openClass = open ? "open" : ""
+
+  return (
+    <Container zIndex={zIndex} anchor={anchor} className={openClass}>
```
dzucconi, line 24: No need to pass the classname any longer.

araujobarret: done in https://github.com/artsy/palette/pull/1295/commits/0a669fce9351066ec0b313301a2805dea7a9d8fb

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@32.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4032.2.0), [`@artsy/palette@33.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4033.2.0) :rocket:

artsyit:

PR #1296, opened 2023-05-16
feat: close dropdown when the footer is clicked

rajsam003:
This PR can resolve this issue https://artsyproduct.atlassian.net/browse/FX-4788

### DIFF
--- a/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
@@ -66,7 +66,6 @@ export interface AutocompleteInputProps<T extends AutocompleteInputOptionType>
   defaultValue?: string
   loading?: boolean
   header?: React.ReactNode
-  // footer?: React.ReactNode
   footer?:
     | React.ReactNode
     | ((dropdownActions: Pick<FooterActions, "onHide">) => void)
--- a/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
@@ -24,7 +24,7 @@ import { AutocompleteInputOption } from "./AutocompleteInputOption"
 import { AutocompleteInputOptionLabel } from "./AutocompleteInputOptionLabel"
 
 export interface AutocompleteFooterActions {
-  /** Call to hide dropdown */
+  /** Call to close dropdown */
   onClose(): void
 }
 
--- a/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
@@ -23,6 +23,11 @@ import { VisuallyHidden } from "../VisuallyHidden"
 import { AutocompleteInputOption } from "./AutocompleteInputOption"
 import { AutocompleteInputOptionLabel } from "./AutocompleteInputOptionLabel"
 
+export interface FooterActions {
+  /** Call to hide dropdown */
+  onHide(): void
+}
+
 /** Base option type — can be expanded */
 export interface AutocompleteInputOptionType {
   text: string
@@ -61,7 +66,10 @@ export interface AutocompleteInputProps<T extends AutocompleteInputOptionType>
   defaultValue?: string
   loading?: boolean
   header?: React.ReactNode
-  footer?: React.ReactNode
+  // footer?: React.ReactNode
+  footer?:
+    | React.ReactNode
+    | ((dropdownActions: Pick<FooterActions, "onHide">) => void)
   /** on <enter> when no option is selected */
   onSubmit?(query: string): void
   /** on <click> or <enter> when an option is selected */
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -23,6 +23,11 @@ import { VisuallyHidden } from "../VisuallyHidden"
 import { AutocompleteInputOption } from "./AutocompleteInputOption"
 import { AutocompleteInputOptionLabel } from "./AutocompleteInputOptionLabel"
 
+export interface FooterActions {
```
dzucconi, line 26: When exporting anything, types included, be sure to prefix: `AutocompleteFooterActions`

packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -219,6 +226,10 @@ export const AutocompleteInput = <T extends AutocompleteInputOptionType>({
     onChange: handleFocusChange,
   })
 
+  const handleHide = useCallback(() => {
+    onClose?.()
+  }, [onClose])
```
dzucconi, line 231: No need to wrap this in a `useCallback`

packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -226,9 +226,9 @@ export const AutocompleteInput = <T extends AutocompleteInputOptionType>({
     onChange: handleFocusChange,
   })
 
-  const handleHide = useCallback(() => {
+  const handleHide = () => {
     onClose?.()
-  }, [onClose])
+  }
```
dzucconi, line 231: This doesn't look like it actually closes the dropdown though? This is only conditionally calling the `onClose` callback; which is something that's passed in.

Isn't https://github.com/artsy/palette/blob/32df05b1e6d23dc9241bae775fd99527ebd869a3/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx#L208-L210 what you want?

rajsam003: I want the dropdown to be closed when the user clicks on the footer. 
Maybe we should call `resetUI()` instead of `onClose()`.

dzucconi:
```
   const handleClose = () => {
+    dispatch({ type: "CLOSE" })
+    reset()
     onClose?.()
   }
```

Consider the following: what do you actually want to happen? What action is in the footer? Do we want to close the drawer without resetting the value? Leave the value? I'd imagine the above is what you want.

Though I'm also a little confused by this PR, like the story didn't work to begin with?

I also think we should call this onClose not onHide, which is semantically closer to what's happening.

dzucconi (approved):

packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -219,6 +226,12 @@ export const AutocompleteInput = <T extends AutocompleteInputOptionType>({
     onChange: handleFocusChange,
   })
 
+  const handleClose = () => {
```
dimatretyak, line 229: **suggestion**: reuse it [here](https://github.com/artsy/palette/blob/ad57a6b10f6bdbfda1bd3087b161ddec04ad9b0f/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx#L215-L217)

packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -219,6 +226,12 @@ export const AutocompleteInput = <T extends AutocompleteInputOptionType>({
...
+    dispatch({ type: "CLOSE" })
+    reset()
+    onClose?.()
```
dimatretyak, line 232: **question**: do we need to call `onClose` here? As far as I remember, if the user selects one of the options, then `onClose` is not called 🤔

dimatretyak (approved):

anandaroop:
@rajsam003 would you mind adding a little context here — screenshots/links for the updated Palette component and/or the Force component that requires this change? 

(I'm having trouble repro-ing the unwanted behavior in staging, thus struggling to see what we're trying to solve here.)

packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
dzucconi: You do, because it's a callback, like if I pass some callback into this component to be called when it's closed, then I want that callback to be called *every* time it's closed, right?

anandaroop:
That's helpful, thanks.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@32.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4032.1.0), [`@artsy/palette@33.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4033.1.0) :rocket:

PR #1299, opened 2023-05-19
fix(autocomplete): improves focus management for pointer devices

dzucconi:
Closes [GRO-1720](https://artsyproduct.atlassian.net/browse/GRO-1720)

A bit unintuitive but, to summarize:

When we click something in the header that isn't a button or link, the click will propagate down to the body and so focus leaves this element. Previously we were using that as an indication that the drawer should be closed. So what we do now to work around this is that on mouse down: flip a flag, and mouse up: reset it. This prevents the element from closing but we _still_ want to close it on actual clicks outside, so we use `useClickOutside` to handle those. Otherwise if you move focus out of the element by tabbing, we still close.

[GRO-1720]: https://artsyproduct.atlassian.net/browse/GRO-1720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.story.tsx
+++ b/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.story.tsx
@@ -3,7 +3,11 @@ import React, { useState } from "react"
 import { States } from "storybook-states"
 import { Box } from "../Box"
 import { Text } from "../Text"
-import { AutocompleteInput, AutocompleteInputProps } from "./AutocompleteInput"
+import {
+  AutocompleteInput,
+  AutocompleteInputOptionType,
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@32.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4032.1.2), [`@artsy/palette@33.1.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4033.1.2) :rocket:

PR #1309, opened 2023-08-25
feat(autocompleteinput): exposes ref to the underlying input

dzucconi:
Exposes a ref to the input so that we can programmatically focus it (`/` keypress for global search)

### DIFF
--- a/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
+++ b/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
@@ -75,6 +75,8 @@ export interface AutocompleteInputProps<T extends AutocompleteInputOptionType>
   footer?:
     | React.ReactNode
     | ((dropdownActions: AutocompleteFooterActions) => void)
+  /** Ref to the input; workaround generics */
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -97,6 +99,7 @@ export const AutocompleteInput = <T extends AutocompleteInputOptionType>({
   dropdownMaxHeight = 308, // 308 = roughly 5.5 options
   flip = true,
   footer,
+  forwardRef: forwardedRef,
```
dzucconi, line 102: Kind of the easiest way to deal workaround limitations with `forwardRef` and generics. Just pass it as a prop! Easy!

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@34.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4034.1.0), [`@artsy/palette@35.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4035.1.0) :rocket:

PR #1313, opened 2023-09-15
fix(range): prevent min from exactly equaling max

anandaroop:
https://artsyproduct.atlassian.net/browse/ONYX-362

Addresses an issue where non-sensical artwork alerts are created, with a min value exactly equal to the max value. Apart from being nearly useless as alerts for collectors, this also confuses the data display for partners.

It seems to make sense to fix this at the library level since min=max is probably not terribly useful in any context?

**Bonus**: this brings the behavior in line with the corresponding component in Eigen.

## Before

`min` could exactly equal `max`.

before-fix gif 5 128

## After

`min` and  `max` are guaranteed to be separated by `step`

after-fix gif 5 128

### DIFF
--- a/packages/palette/src/elements/Range/Range.tsx
+++ b/packages/palette/src/elements/Range/Range.tsx
@@ -38,7 +38,7 @@ export const Range: React.FC<RangeProps> = ({
   const handleMinChange = ({
     target: { valueAsNumber: value },
   }: React.ChangeEvent<HTMLInputElement>) => {
-    if (value > values[1]) return
+    if (value > values[1] - step) return
     setValues([value, values[1]])
     onChange?.([value, values[1]])
   }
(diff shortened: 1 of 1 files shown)


anandaroop:
@dzucconi thoughts on whether any clients might actually rely on the prior behavior / whether this should be released as a breaking change?

dzucconi (approved):
Nice catch.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@35.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4035.1.0), [`@artsy/palette@36.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4036.1.0) :rocket:

PR #1314, opened 2023-09-18
feat(pills): adds profile and gray variant; removes artist variant

dzucconi:
We can alter these actual styles later if there turns out to be a Figma spec. For now this is fine. Breaking change because we're dropping the `artist` variant.

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.story.tsx
+++ b/packages/palette/src/elements/Pill/Pill.story.tsx
@@ -64,10 +64,10 @@ export const Variants = () => {
 export const LinkExample = () => {
   return (
     <Pill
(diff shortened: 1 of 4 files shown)


dzucconi:
I don't like the inconsistency with the disabled states but we actually had a design both of those

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@35.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4035.0.0), [`@artsy/palette@36.0.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4036.0.0) :rocket:

PR #1316, opened 2023-09-21
Be defensive when animating Collapse

starsirius:
I noticed a flaky error happens on Force CI ([example](https://app.circleci.com/pipelines/github/artsy/force/52795/workflows/8bb542a0-ed21-4864-bd46-053663728635/jobs/395284)) sometimes when running tests:

```
TypeError: Cannot read properties of null (reading 'style')

  at style (node_modules/@artsy/palette/src/elements/Collapse/Collapse.tsx:56:26)
```

When running tests individually, this error doesn't happen. My theory is that the callback in `setTimeout` is not finished in the same test and instead is leaked into the subsequent test and fails with the `null` error. Perhaps the element is already gone by then.

This PR makes the callback more defensive. Not really sure how to write a test for this. 😅

### DIFF
--- a/packages/palette/src/elements/Collapse/Collapse.tsx
+++ b/packages/palette/src/elements/Collapse/Collapse.tsx
@@ -44,7 +44,9 @@ export class Collapse extends React.Component<CollapseProps> {
       this.wrapperRef.style.height = prevHeight
       // wait for a tick before setting goal height to allow transition
       this.wrapperModifyTimeout = setTimeout(() => {
-        this.wrapperRef!.style.height = goalheight + "px"
+        if (this.wrapperRef) {
+          this.wrapperRef.style.height = goalheight + "px"
+        }
       }, 10)
     } else if (!this.props.open && this.wrapperRef.style.height !== "0px") {
       // animate closing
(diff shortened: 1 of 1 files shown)


dzucconi:
Geniunely surprised things use this component and that it's not deprecated — it's not in the design system and there are no stories for it. Usage is going to be isolated to the order app I imagine — possible to get rid of it?

starsirius:
@dzucconi Yeah you're right that [only the Orders app in Force](https://github.com/search?q=repo%3Aartsy%2Fforce%20import%20Collapse&type=code) uses it. So what I can do is to port this component to Force and apply this change, and then we can deprecate and remove it from palette. Does it sound good to you?

dzucconi:
@starsirius sure that sounds reasonable to me. Would be happy to get this out of Palette. But I was also wondering if there's something in the design system we could replace this with like the `Expandable` or something?

starsirius:
I can bring it to the team and designer and see if we can be more aligned with the design system! For now, I'm going to port this over to Force and address the test flakiness. Thank you for talking through!

And interesting that how we name elements in opposite way (e.g. Expandable vs. Collapse)! :D

starsirius:
Closing!

PR #1320, opened 2023-10-06
fix(pill): corrects x icon behavior

dzucconi:
The `selected` state can be stacked with the other states. As such the `filter` variant of `Pill` had an idle and selected state which were identical. This corrects that. We also add the x icon to the `gray` variant so that it matches it's larger version.

### DIFF
--- a/packages/palette/src/elements/Pill/Pill.tsx
+++ b/packages/palette/src/elements/Pill/Pill.tsx
@@ -111,7 +111,8 @@ export const Pill = forwardRef<
         )}
       </Text>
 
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@35.1.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4035.1.3), [`@artsy/palette@36.1.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4036.1.3) :rocket:

PR #1322, opened 2023-10-13
fix(EMI-1373): Fix Select issues with Firefox

MrSltun:
This PR resolves [EMI-1373]

> [!IMPORTANT]
> This PR will be merged to #1250 

### Description
This PR solves the issue with `:has` pseudo-class [not being supported by Firefox](https://caniuse.com/css-has) by applying the styles using JS instead of CSS selectors

### Videos
| Before | After |
|---|---|
| <video src="https://github.com/artsy/palette/assets/20655703/221a35af-56c4-4d3a-9974-c78d18e0ddc0" /> | <video src="https://github.com/artsy/palette/assets/20655703/5c8d3399-77fd-463c-9e9b-a2d938ebd66f" /> |

[EMI-1373]: https://artsyproduct.atlassian.net/browse/EMI-1373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -1,11 +1,17 @@
+import THEME from "@artsy/palette-tokens"
 import { themeGet } from "@styled-system/theme-get"
-import React, { forwardRef, ForwardRefExoticComponent, Ref } from "react"
+import React, {
+  forwardRef,
+  ForwardRefExoticComponent,
+  Ref,
+  useState,
+} from "react"
 import styled, { css } from "styled-components"
+import { RequiredField } from "../../shared/RequiredField"
 import { Box, BoxProps, splitBoxProps } from "../Box"
 import { Text } from "../Text"
-import { SELECT_STATES } from "./tokens"
 import { Tooltip } from "../Tooltip"
-import { RequiredField } from "../../shared/RequiredField"
+import { SELECT_STATES } from "./tokens"
 
 export interface Option {
   value: string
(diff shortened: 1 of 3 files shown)


MrSltun (commented):
Since we have the styles now in JS, should we remove the `:has` styles from [tokens.ts](https://github.com/artsy/palette/blob/ea2b6cafdadd9e1a775859434201ca543b5eb7dd/packages/palette/src/elements/Select/tokens.ts)?

iskounen:
This bug can be reproduced with the Price Range selector on the Account Settings page in Force, so we can test it there once this is merged.

packages/palette/src/elements/Select/Select.tsx
```diff
@@ -51,6 +57,50 @@ export const Select: ForwardRefExoticComponent<
...
+      }
+
+    // if the select is hovered and there is no selected option (or it is empty)
+    // and the select is not disabled
+    const labelStyleHovered = isHovered &&
+      !selectedOption &&
+      !disabled && {
+        textDecoration: "underline",
+        color: THEME.colors.blue100,
+      }
```
dzucconi, line 103: A few things:
* Why not leave these declarations in the tokens file and just use them here?
* You don't want to be accessing the `THEME` object — just spread the props on to the label, don't use `style`

MrSltun: > You don't want to be accessing the THEME object — just spread the props on to the label, don't use style

I don't think I can apply the styles by spreading the props on the components, I tested it and it wasn't working, especially with `<span>`, I'm definitely missing something but I'm not sure what, would love your input on this!

dzucconi: Sure you can: `<Box as="span">`

damassi (approved):
Amazed that firefox does not have `:has` yet!

artsyit:

PR #1329, opened 2023-10-27
fix(dia-240): optimize tooltip when inactive

dzucconi:
Closes [DIA-240](https://artsyproduct.atlassian.net/browse/DIA-240)

While although I cannot replicate a crash, no matter how many objects appear on page. I was seeing significant frame drops when resizing the window prior to this change, starting at around 3000 instances.

That said: obviously hundreds or thousands of components should never be in the DOM period and we should virtualize that list anyway. But, this should help.

[DIA-240]: https://artsyproduct.atlassian.net/browse/DIA-240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/Tooltip/Tooltip.story.tsx
+++ b/packages/palette/src/elements/Tooltip/Tooltip.story.tsx
@@ -10,6 +10,7 @@ import { Input } from "../Input"
 import { Spacer } from "../Spacer"
 import { Text } from "../Text"
 import { Tooltip, TooltipProps } from "./Tooltip"
+import { Stack } from "../Stack"
 
 const CONTENT = "Lorem ipsum dolor sit amet consectetur adipisicing elit?"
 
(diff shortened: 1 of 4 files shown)


dzucconi:
Going to merge. Give this a whirl to see if the bug can still be replicated.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@35.3.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4035.3.1), [`@artsy/palette@36.3.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4036.3.1) :rocket:

damassi (commented):
Thanks so much @dzucconi

PR #1330, opened 2023-10-30
fix(button): offsets icon for padding balance

dzucconi:
Surprised this never came up. With the `Pill`s we offset the icons so that the padding doesn't look visually off balance. We should do the same on `Button`

| Before | After |
| --- | --- |
|  |  |


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@35.3.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4035.3.2), [`@artsy/palette@36.3.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4036.3.2) :rocket:

PR #1334, opened 2023-11-07
fix(EMI-1560): Add margin-top to elements when the title is present

MrSltun:
This PR resolves [EMI-1560]

### Description
This PR adds a margin-top to the input elements when the `title` is present to prevent the title from overlapping on other elements

### Demo & Screenshots
| | Before | After |
|---|---|---|
| Input | <video src="https://github.com/artsy/palette/assets/20655703/4232fcf4-b6e6-4682-91fc-2ad601657e3b" /> | <video src="https://github.com/artsy/palette/assets/20655703/50168f9e-6932-46f5-b866-5df04a47138f" /> |
| TextArea | image | image |
| Select | image | image |	
| MultiSelect | image | image |

	

[EMI-1560]: https://artsyproduct.atlassian.net/browse/EMI-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/Input/Input.tsx
+++ b/packages/palette/src/elements/Input/Input.tsx
@@ -66,7 +66,7 @@ export const Input: React.ForwardRefExoticComponent<
           </Tooltip>
         )}
 
-        <Box position="relative">
+        <Box position="relative" mt={!!title && !description ? 1 : 0}>
           <StyledInput
             ref={ref as any}
             disabled={disabled}
--- a/packages/palette/src/elements/MultiSelect/MultiSelect.tsx
+++ b/packages/palette/src/elements/MultiSelect/MultiSelect.tsx
@@ -113,6 +113,7 @@ export const MultiSelect: React.FC<MultiSelectProps> = ({
       )}
 
       <Container
+        mt={!!title && !description ? 1 : 0}
         ref={anchorRef as any}
         onClick={onVisible}
         complete={complete || selection.length > 0}
--- a/packages/palette/src/elements/Select/Select.tsx
+++ b/packages/palette/src/elements/Select/Select.tsx
@@ -75,6 +75,7 @@ export const Select: ForwardRefExoticComponent<
         )}
 
         <Container
+          mt={!!title && !description ? 1 : 0}
           disabled={!!disabled}
           hover={!!hover || isHovered}
           error={error!}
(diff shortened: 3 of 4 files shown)


dzucconi:
I don't think we should be doing this and rather should be fixing it at the call sites.

dzucconi:
To be clear why: This change would make it very confusing to ever use these inputs with layouts that have flex or grid gaps.

damassi:
@dzucconi - Lets deal with this grid gap issue when it arises and get this shipped. I understand the no external margin rule, and really try to follow it, but in this case, where you have inputs on a page (typically many, back to back) this fix feels fine. The "external margin" is actually the fully-measured distance of the animated label. So in a sense its still internal. 

To fix all of this at the call-site areas feels a bit much. To refactor this later would be trivial. So lets ship this thing. We can debate this elsewhere, but for now going to merge this. Feel free to open a PR fixing things you feel need addressing and making it perfect after this goes live.

dzucconi:
IMO this is going to be way more annoying to work around than to just fix the problems where they are. I don't even understand how this was a problem since there wasn't external margins in the first place.

damassi:
I imagine it has something to do with the animated label being pulled outside of the boundary. So while it didn't have an external margin before, with the label, it does (kinda sorta). And now we need to account for things. 

We're so close to launching this. Lets work together here. Lets get the remaining issue in the QA board in and merged and we can all follow up where we see fit. Its the nature of iterating.

dzucconi:
It's fine but at the very least: how do I disable this so when I work with it I can have layouts that don't have illegal whitespace values?

dzucconi:
Also: nature of iterating this isn't. It's introducing a breaking visual change that's going to require constant workarounds.

dzucconi:
I also wonder if this is why in Material 3 they went with a bounding box around the whole thing.

damassi:
> It's fine but at the very least: how do I disable this so when I work with it I can have layouts that don't have illegal whitespace values?

Open a PR real quick with a prop. It won't be the end of the world, it will actually be totally fine. If you end up not needing to use the prop before you get around to refactoring every single input callsite across force, volt, forque, (what else?) first, just open a PR and delete it. 

For large last minute requests like this, feel free to take the torch and open some PRs. Its a lot to drop on someone not yourself. Feel free to ping me in slack and we can continue debating.

dzucconi:
How does this not immediately create a spacing issue in literally all grid layouts?

artsyit:

PR #1337, opened 2023-11-28
fix(range): correctly sets width when within drawer

dzucconi:
It's actually unclear to me how this differed from the ModalDialog which also uses a portal but: when we switched the `Drawer` to use a portal, the range was no longer initializing it's width correctly.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.0.2), [`@artsy/palette@37.0.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.0.2) :rocket:

PR #1338, opened 2023-12-07
POC: Spike on doing <Collapse/> with a <details> element

erikdstock:
In spired by a random question I asked in `#web-practice` yesterday... I thought this would be much easier, and by the end I'm not sure if this was any improvement over what Collapse already does. 

I basically ported the implementation from [this article](https://css-tricks.com/how-to-animate-the-details-element/) to a react FC. lots of type errors but it seems to work, both as a details element (click on summary) and with an external control. I don't know if this would work in React Native. I didn't bother with the article's little carat animation or substituting my own icon in, but that would obviously be possible.

Not assigning so I don't create extra noise.
#minor

<details><summary>Example: external controls via prop</summary>

</details>
<details><summary>Example: click-summary controls</summary>

</details>


erikdstock:
Closing, thanks all for the discussion in slacc

PR #1339, opened 2023-12-11
chore: Switch useHasEnteredViewport to use IntersectionObserver instead of unreliable logic

jo-rs:
Switches the `useHasEnteredViewport` hook to use Force's implementation of `IntersectionObserver` instead of the former logic, which was causing issues on Volt V2.

Open to other suggestions.


dzucconi:
We should probably do a pass on the charts sub-package at some point to modernize it.

You don't really want to interface with the `IntersectionObserver` outside of an effect though.

Feel free to port over our hook in Force to Palette and update callers, if you are motivated: https://github.com/artsy/force/blob/main/src/Utils/Hooks/useIntersectionObserver.ts

packages/palette-charts/src/elements/DataVis/utils/useHasEnteredViewPort.ts
```diff
@@ -1,24 +1,26 @@
...
 export const useHasEnteredViewport = (ref: React.RefObject<HTMLElement>) => {
+
   const [hasEntered, setHasEntered] = useState(false)
+
+  let observer: IntersectionObserver = null
+
+  if (typeof window !== "undefined"){
+    observer = useMemo(() => new IntersectionObserver(
+        ([entry]) => setHasEntered(entry.isIntersecting)
+      ), [ref])
```
araujobarret, line 16: non-blocker: probably prettier didn't pass here, the indentation is not aligned.

packages/palette-charts/src/elements/DataVis/utils/useHasEnteredViewPort.ts
```diff
@@ -1,24 +1,26 @@
...
  * @param ref reference to component
  */
 export const useHasEnteredViewport = (ref: React.RefObject<HTMLElement>) => {
+
   const [hasEntered, setHasEntered] = useState(false)
+
+  let observer: IntersectionObserver = null
+
+  if (typeof window !== "undefined"){
+    observer = useMemo(() => new IntersectionObserver(
```
araujobarret, line 14: that hurts one of the hooks' principles, a hook never should be after an if statement or inside of it.

packages/palette-charts/src/elements/DataVis/utils/useHasEnteredViewPort.ts
```diff
@@ -1,24 +1,26 @@
...
  * Returns true when component enters viewport
  * @param ref reference to component
  */
 export const useHasEnteredViewport = (ref: React.RefObject<HTMLElement>) => {
+
   const [hasEntered, setHasEntered] = useState(false)
+
+  let observer: IntersectionObserver = null
+
+  if (typeof window !== "undefined"){
```
dzucconi, line 13: This kind of thing creates problems for server-side rendering because the render will not match when it's hydrated.

packages/palette-charts/src/elements/DataVis/utils/useHasEnteredViewPort.ts
dzucconi, line 14: You can't call hooks conditionally

jo-rs:
Sorry, I was frustrated when I pushed the last commit, and it was a bit of a mess. As per @dzucconi's [comment](https://github.com/artsy/palette/pull/1339#issuecomment-1850063272), I ported over Force's hook.

packages/palette-charts/src/elements/BarChart/BarChart.test.tsx
```diff
@@ -35,6 +38,14 @@ const mockBars = [
...
 
 describe("BarChart", () => {
+  const mockUseIntersectionObserver = useIntersectionObserver as jest.Mock
+
+  beforeAll(() => {
+    mockUseIntersectionObserver.mockImplementation(() => ({
+      ref: createRef(),
+    }))
+  })
+
```
araujobarret, line 48: if you wanna walk the extra mile to have a proper test for it later, [example](https://github.com/artsy/volt-v2/blob/main/src/pages/conversations/%5BconversationId%5D/Components/__tests__/ConversationMessages.test.tsx#L181)

araujobarret (approved):

packages/palette-charts/src/elements/DataVis/useIntersectionObserver.ts
```diff
@@ -0,0 +1,72 @@
...
+    threshold: number | number[]
+    root?: Element
+    rootMargin?: string
+  }
+  onIntersection?: (entries: IntersectionObserverEntry[]) => void
+  onOffIntersection?: (entries: IntersectionObserverEntry[]) => void
+}
+
+// tslint:disable-next-line:completed-docs
+export const useIntersectionObserver = ({
```
damassi, line 16: 👍

damassi (approved):
As long as this still works in the artist page on force all good 👍

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.2.0), [`@artsy/palette@37.2.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.2.0) :rocket:

PR #1345, opened 2023-12-21
WIP => refactor(popover): New positioning engine 

damassi:
WIP

### DIFF
--- a/packages/palette/package.json
+++ b/packages/palette/package.json
@@ -119,6 +119,8 @@
   "dependencies": {
     "@artsy/icons": "^3.2.2",
     "@artsy/palette-tokens": "^5.0.0",
+    "@floating-ui/dom": "^1.5.3",
+    "@floating-ui/react": "^0.26.4",
     "@seznam/compose-react-refs": "^1.0.6",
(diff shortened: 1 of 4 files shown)


dzucconi:
Keep in mind we'd want to refactor everything that uses `usePosition`

damassi:
Right, yeah I just wanted to throw up something that (kinda) works to see what the crossover might be. I would imagine we'd simply replace usePosition internals with this to keep the same api, but this is more in your court. 

@dzucconi - Q: if this is a big project (and it seems like it is) would you be open to pulling down volt 2 and updating `Layout` to work as expected, where body scrolls instead of this wack inner div? I imagine that would be a 10m change for you, and that would push the necessity of this refactor out.

dzucconi:
Yeah that sounds reasonable. Will take a look tomorrow AM

damassi:
Thanks so much -- just be sure to boot volt 1 before starting volt 2, and everything should be good to dev in. (Can just push to my progressive onboarding branch, as the tooltips and so on are all ready to go) 👍 🙏

PR #1349, opened 2024-01-02
fix(useportal): makes safe for ssr

dzucconi:
Relying on state here instead of the check on window should work. Will publish canary and verify.


damassi:
Yep that's def it 👍

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.2.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.2.3), [`@artsy/palette@37.2.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.2.3) :rocket:

PR #1353, opened 2024-01-05
fix(checkbox): moves focus state onto focus-visible

dzucconi:
We were seeing some slightly confusing UX when checkboxes were the first element in dropdowns, as when focus moves to the dropdown, the first checkbox gets focused. This actually does feel much better to use with a pointer device now as the checkbox won't visibly retain focus.

### DIFF
--- a/packages/palette/src/elements/Checkbox/Checkbox.tsx
+++ b/packages/palette/src/elements/Checkbox/Checkbox.tsx
@@ -123,7 +123,7 @@ const Container = styled(Box)<{
         `}
       }
 
-      &:focus {
+      &:focus-visible {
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.3.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.3.2), [`@artsy/palette@37.3.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.3.2) :rocket:

damassi:
oh geez, i wonder why renovate wasn't updating. Will do

PR #1355, opened 2024-01-10
fix(button): avoids passing the loading prop to the underlying dom node

dzucconi:
Closes [DIA-387](https://artsyproduct.atlassian.net/browse/DIA-387)

This just renames the loading prop on the container to match the transient props API in SC5. It doesn't actually work as a transient prop in this case because we are on 4, but it suppresses the warning regardless because it's not passing anything to `loading`, which is a native attribute.

[DIA-387]: https://artsyproduct.atlassian.net/browse/DIA-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/Button/Button.tsx
+++ b/packages/palette/src/elements/Button/Button.tsx
@@ -79,7 +79,7 @@ export const Button: React.ForwardRefExoticComponent<
         ref={composeRefs(ref, forwardedRef) as any}
         onClick={handleClick}
         size={size}
-        loading={loading}
+        $loading={loading}
         success={success}
         tabIndex={loading ? -1 : 0}
         display="inline-flex"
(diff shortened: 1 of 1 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.3.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.3.3), [`@artsy/palette@37.3.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.3.3) :rocket:

damassi:
huge quality of life improvement. Thank you!

PR #1356, opened 2024-01-10
fix(tooltip): correctly handles positioning edge case by skipping when inactive

dzucconi:
We noticed that on some pages with `Tooltip` components in right sidebars, we would see the tooltip dramatically positioned offscreen. @tam-kis noticed that by changing the immediate positioning context to relative we could contain it within the parent. I don't think this change has gone unnoticed for 3 months, so it's unclear what changed to trigger this bug. However we just need to skip positioning when the `Tooltip` is inactive, regardless.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@36.3.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4036.3.4), [`@artsy/palette@37.3.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4037.3.4) :rocket:

PR #1367, opened 2024-02-29
fix(theme): returns the default theme if none is provided

dzucconi:
I'm not sure why I didn't think of this when I was upgrading Force.... 😞 

This just makes it much easier to test: tests may not have a component wrapped with a theme provider and if you're using a component that makes use of the theme, then these will fail. 99% of the time, this doesn't matter at all. So just return the default theme if there isn't one provided.

Re: https://github.com/artsy/volt-v2/pull/1244

### DIFF
--- a/packages/palette/src/Theme.tsx
+++ b/packages/palette/src/Theme.tsx
@@ -17,6 +17,8 @@ const THEMES = {
   dark: THEME_DARK,
 }
 
+const DEFAULT_THEME = THEME
+
 interface ThemeProps {
   children?: React.ReactNode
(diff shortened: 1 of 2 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@37.1.5`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4037.1.5), [`@artsy/palette@38.1.5`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4038.1.5) :rocket:

PR #1380, opened 2024-05-02
fix(avatar): hides broken image if avatar load fails

dzucconi:
When you upload an avatar, Metaphysics will return an invalid URL — which is crazy — but this is a reasonable thing to do with errored images, regardless.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@37.9.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4037.9.1), [`@artsy/palette@38.9.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4038.9.1) :rocket:

PR #1385, opened 2024-05-29
fix(phoneinput): fixes for dark mode

dzucconi:
* Fixes for dark mode
* Fixes border color to align with spec

### DIFF
--- a/packages/palette/src/elements/PhoneInput/PhoneInput.tsx
+++ b/packages/palette/src/elements/PhoneInput/PhoneInput.tsx
@@ -445,6 +445,7 @@ const SelectOption = styled(Box)<{ selected?: boolean }>`
 
 const StyledInput = styled.input`
   width: 100%;
+  background-color: ${themeGet("colors.white100")};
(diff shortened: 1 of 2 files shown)


packages/palette/src/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -445,6 +445,7 @@ const SelectOption = styled(Box)<{ selected?: boolean }>`
 
 const StyledInput = styled.input`
```
dzucconi, line 446: Wondering why we didn't just use a Palette input here so that these don't drift. Alternatively could have extracted a mixin. This will probably crop up again as a problem area.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@37.11.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4037.11.1), [`@artsy/palette@38.11.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4038.11.1) :rocket:

PR #1387, opened 2024-05-30
feat(DIA-663): add disclosure element

araujobarret:
[DIA-663]

Add the Disclosure element which is very similar to the `Expandable`.

This work is a bit ahead of time for the element being added to the DS so some things are still pending, such as disabled style.

https://github.com/artsy/palette/assets/15792853/e00a03ca-34ff-4751-83c8-1d00560381a0

[DIA-663]: https://artsyproduct.atlassian.net/browse/DIA-663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@37.13.0-canary.1387.30738.0
  npm install @artsy/palette@38.13.0-canary.1387.30738.0
  # or 
  yarn add @artsy/palette-charts@37.13.0-canary.1387.30738.0
  yarn add @artsy/palette@38.13.0-canary.1387.30738.0
  ```
</details>


packages/palette/src/elements/Disclosure/Disclosure.story.tsx
```diff
@@ -0,0 +1,55 @@
...
+              <Text variant="sm-display">Heading</Text>
+              <Clickable
+                textDecoration="underline"
+                onClick={(e) => {
+                  e.stopPropagation()
+                  action("Secondary action")
+                }}
+              >
+                <Text variant="sm-display">Reset</Text>
+              </Clickable>
```
dzucconi, line 32: This sort of stuff isn't relevant for this component. I wouldn't simply copy expandable's stories here.

dzucconi:
I'd just wait for the full spec and implementation before merging.

packages/palette/src/elements/Disclosure/Disclosure.story.tsx
araujobarret: I removed a few, this case got me if would make sense to be here or not, going to remove it then

araujobarret:
It makes sense, I will leave it here while DS is not fully ready, at least I can test with volt-v2 locally which is kinda done already, only missing tests.

dzucconi:
Here's the WAI-ARIA pattern for this thing: https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/

I'm pretty sure `Expandable` complies. Vaguely curious if there's a lower level abstraction that can be shared. Though also this is simple enough that it's probably not worth it.

damassi:
I'm kinda wondering why this would be in the DS -- its more of a one-off kinda thing. I can't imagine Disclosure being used widely?

dzucconi:
I had the same question but apparently it's being added to the broader design system. I do feel like it should be there _before_ we do the implementation. But 🤷

damassi:
Yah, i agree -- add to the spec, then add here.

araujobarret:
I'm closing it since it won't be added in the DS after some alignment.

PR #1395, opened 2024-08-19
fix: replaces lodash with es-toolkit

dzucconi:
Re: [DIA-790](https://artsyproduct.atlassian.net/browse/DIA-790)

lodash is currently 16% of palette's size.

[es-toolkit](https://es-toolkit.slash.page/) is a replacement for lodash that is [much smaller](https://es-toolkit.slash.page/bundle-size.html).

We shouldn't be importing all of lodash for a few utility functions.

We should attempt [to do this in Force](https://es-toolkit.slash.page/compatibility.html) as well. Thoughts?

[DIA-790]: https://artsyproduct.atlassian.net/browse/DIA-790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/.eslintrc.js
+++ b/packages/palette/.eslintrc.js
@@ -36,5 +36,12 @@ module.exports = {
     "react-hooks/exhaustive-deps": "warn",
     "react-hooks/rules-of-hooks": "error",
     "react/prop-types": 0,
+    "no-restricted-imports": [
+      "error",
+      {
+        name: "lodash",
+        message: "Please use `es-toolkit` instead.",
+      },
+    ],
   },
 }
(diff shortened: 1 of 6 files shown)


damassi:
I dig it! Moving this into force feels worksome, but i'd be in favor, depending on the output of `yarn why lodash` -- it might be a case where 3rd party libs bring it in as a transitive dep and it can't be avoided.

damassi:
@dzucconi - before merge, can you update the eslint restricted imports rule to guard against imports? (Might not be necessary tho due to throwing type error)

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@37.13.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4037.13.3), [`@artsy/palette@38.13.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4038.13.3) :rocket:

PR #1396, opened 2024-08-26
fix: moves focus-visible styling into global styles

dzucconi:
We're removing the [focus-visible polyfill from force](https://github.com/artsy/force/blob/main/src/Components/FocusVisible.tsx), where this styling was previously defined.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@37.13.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4037.13.4), [`@artsy/palette@38.13.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4038.13.4) :rocket:

PR #1410, opened 2024-11-15
fix: Removes delay when not transitioning

dzucconi:
We introduced a subtle delay to handle hover transitions, which can be removed when no transition is occurring for more responsive behavior.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@38.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4038.1.1), [`@artsy/palette@39.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4039.1.1) :rocket:

PR #1417, opened 2025-01-23
feat(FilterSelect): enable additional searchable text and quick select-all

anandaroop:
PoC approach for Hackathon project: 
https://www.notion.so/artsy/Allow-artwork-filtering-by-region-17bcab0764a080af9eb8e7e7a03b40cf?pvs=4

The initial search for `usa` is how the component currently works.

Subsequent searches for synonyms and regions, as well as quick-selecting all options is enabled by the functionality in this PR:


packages/palette-charts/package.json
```diff
@@ -13,7 +13,6 @@
     "clean": "rm -rf dist",
     "compile": "babel src --source-maps --extensions '.ts,.tsx' --ignore src/**/__tests__,src/**/__stories__ --out-dir dist",
     "lint": "tslint --project tsconfig.json 'src/*/**'",
-    "prepublishOnly": "yarn clean && yarn compile && yarn type-declarations",
```
damassi, line 16: When I was upgrading things to React 18 I had to hand publish this one -- couldn't figure out what was going on

anandaroop: Yeah, Not sure how to debug this. But I've removed that commit since I don't need the canary anymore 🤷🏽

packages/palette/src/elements/FilterSelect/FilterSelect.tsx
```diff
@@ -38,6 +44,12 @@ const _FilterSelect: React.FC<React.PropsWithChildren<unknown>> = () => {
     selectedItems,
   } = useFilterSelectContext()
 
+  if (!multiselect && enableSelectAll) {
+    throw new Error(
+      "FilterSelect: enableSelectAll is only available with multiselect mode."
+    )
```
damassi, line 50: Should we throw runtime errors like this on the client, or would a console.error be enough?

damassi (approved):
One tiny non-blocking q; otherwise LGTM 👍

packages/palette/src/elements/FilterSelect/FilterSelect.tsx
anandaroop: Yeah I was thinking about this — I figured one loud dev-time error to prevent you from using the component incorrectly was better than a console error that's (unfortunately) quite likely to be drowned out by other console noise, at least when in the development environment.

damassi: My personal rule of thumb is to never induce a runtime error because you never know if that might break the page in some subtle way. (ie, Incorrect dropdown or stalled out UX; the latter seems worse.)

packages/palette/src/elements/FilterSelect/FilterSelect.tsx
```diff
@@ -109,3 +125,50 @@ export const isBelowTheFoldSelected = (selectedItems, resultsSorted) => {
...
+          onClick={() => {
+            setSelectedItems(filteredItems)
+            onSelectAll?.({
+              items,
+              filteredItems,
+              selectedItems: filteredItems,
+              query,
+            })
+          }}
+          textDecoration={"underline"}
```
olerichter00, line 156:
```suggestion
          textDecoration="underline"
```

packages/palette/src/elements/FilterSelect/Components/FilterInput.tsx
```diff
@@ -32,6 +32,7 @@ export const FilterInput: React.FC<React.PropsWithChildren<InputProps>> = (
         query !== "" ? (
           // Active state
           <Clickable
+            className="clearSearchInput"
```
olerichter00, line 35: Is this class name used anywhere? 👀

olerichter00 (approved):

packages/palette/src/elements/FilterSelect/FilterSelect.tsx
anandaroop: Something about our vsc/prettier setup does this automatically I think, but I agree

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@39.3.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4039.3.0), [`@artsy/palette@40.4.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4040.4.0) :rocket:

PR #1422, opened 2025-02-20
fix(image): pass props even when not lazyloading

dzucconi:
Fixed an issue where props weren't being passed to the Image component when loading normally (not lazy), causing several features to break. Most notably, the placeholder URL was non-functional. This patch ensures props are passed uniformly across all component variations.


dzucconi:
All right, I don't think this is even worth fixing. This has sort of gotten away from me. I'm just gonna use a native image from now on.

PR #1427, opened 2025-04-04
feat(dia-1259): implements fitting for dropdowns

dzucconi:
Re: https://github.com/artsy/force/pull/15387 — here we implement a better implementation of dropdown fitting for all dropdowns.

Let's see how the canary looks in Force. Will probably want to also allow for opting out of this behavior as one can imagine dropdowns on scrollable elements that you may want to flip instead of fit.

### DIFF
--- a/packages/palette/src/elements/Dropdown/Dropdown.story.tsx
+++ b/packages/palette/src/elements/Dropdown/Dropdown.story.tsx
@@ -247,27 +247,47 @@ export const OverflowingContent = () => {
     </Text>
   )
 
+  const [placement, setPlacement] = useState<Position>("bottom")
+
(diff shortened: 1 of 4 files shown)


dzucconi:
Looks good in Force. Merging. Will circle back if we need to add an explicit `fit` prop.

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@39.7.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4039.7.0), [`@artsy/palette@40.8.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4040.8.0) :rocket:

PR #1432, opened 2025-04-11
fix(popover): prevents layout thrashing at specific zoom levels

dzucconi:
We discovered that displaying a popover at certain zoom levels causes the page to crash. Investigation revealed this is caused by layout thrashing when the browser repeatedly reads and writes fractional pixel values in the `MutationObserver` callback. `placeTooltip` performs a read of the DOM and then subsequently uses that measurement to write a transform.

By wrapping these state updates in `requestAnimationFrame` we can just batch them into a single pass, preventing the self-triggering loop.

### DIFF
--- a/packages/palette/src/elements/Popover/Popover.story.tsx
+++ b/packages/palette/src/elements/Popover/Popover.story.tsx
@@ -235,3 +235,67 @@ export const PopoverActions = () => {
     </States>
   )
 }
+
+export const CrashAtSpecificZoomLevels = () => {
+  return (
+    <>
+      <Box height={100}>Zoom to 90% in Chrome, click, then scroll.</Box>
+
+      <Box height={2000} bg="black5">
+        <Box height={200} />
+
+        <Popover
+          placement="top"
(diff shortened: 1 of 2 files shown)


dzucconi:
Weird that canary didn't deploy

dzucconi:
Looks good in Force

packages/palette/src/elements/Popover/Popover.story.tsx
```diff
@@ -235,3 +235,67 @@ export const PopoverActions = () => {
     </States>
   )
 }
+
+export const CrashAtSpecificZoomLevels = () => {
+  return (
+    <>
+      <Box height={100}>Zoom to 90% in Chrome, click, then scroll.</Box>
```
joeyAghion, line 242: Appreciate the repro 😍

damassi:
haha i wonder if we can go back to palette tooltips in Convos now 😓 😓 😅 - hardest bug ever to track down

dzucconi:
I think the key is really building out the repro in a story.

PR #1435, opened 2025-05-02
fix(dia-1305): fixes initial autocomplete render in safari

dzucconi:
It seems that Safari's rendering engine has slightly different timings. When rendering an autocomplete with autofocus it the positioning code ran before Safari has finished calculating the element's dimensions and position. This change ensures all layout calculations are complete.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@40.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4040.1.0), [`@artsy/palette@41.1.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4041.1.0) :rocket:

PR #1436, opened 2025-05-06
fix(avatar): prevents avatar from deforming

dzucconi:
If an avatar is in the same container as some element that doesn't shrink it will shrink/deform. It should always remain a circle and the same size. Text should be adjusted to accommodate/shrink (Will open a corresponding Force PR)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@40.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4040.1.1), [`@artsy/palette@41.1.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4041.1.1) :rocket:

PR #1444, opened 2025-05-22
fix: MultiSelect selected items check

olerichter00:
## Description

This PR fixes the issue of checking whether `MultiSelect` items are selected by comparing their values instead of object references.

When comparing object references to determine if an item is selected, any changes to the items in `options` (which are objects) will cause the check to fail, making all items appear unselected.


Mitch-Henson (approved):
So exciting, great find!

PR #1446, opened 2025-05-30
fix: Set Drawer `transform` to `none` when opened for fixed children

olerichter00:
## Description

Set Drawer `transform` to `none` when it's opened to make fixed elements work inside (e.g. AutocompleteInput).

[The uncanny relationship between position fixed and transform property](https://dev.to/salilnaik/the-uncanny-relationship-between-position-fixed-and-transform-property-32f6)

### DIFF
--- a/packages/palette/src/elements/Drawer/Drawer.tsx
+++ b/packages/palette/src/elements/Drawer/Drawer.tsx
@@ -1,11 +1,11 @@
+import { themeGet } from "@styled-system/theme-get"
 import React, { FC } from "react"
-import { Box } from "../Box"
-import { Flex } from "../Flex"
+import { FocusOn } from "react-focus-on"
 import styled, { css } from "styled-components"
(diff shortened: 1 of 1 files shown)


lidimayra (approved):
Thank you!!! 💪 💟

lidimayra:
For reference: This issue is currently discussed [on this thread](https://artsy.slack.com/archives/C05F8TNKGAV/p1748537455230889)

And a bug reproduction PR was submitted here just to illustrate the issue: https://github.com/artsy/palette/pull/1447

dzucconi:
Presumably this breaks the transition? (We can keep it by using the web animation API + callbacks)

olerichter00:
> Presumably this breaks the transition? (We can keep it by using the web animation API + callbacks)

I've tested with Safari and Chrome, and the animation looks fine.

dzucconi:
Nice then!

PR #1452, opened 2025-07-02
fix(stories): prevents layout thrashing

dzucconi:
Related: https://github.com/artsy/palette/pull/1432


rabidpaul (approved):
This fixes it for me! Thanks, @dzucconi 🙏

PR #1453, opened 2025-07-11
fix(dia-1404): dropdown fixes

dzucconi:
* sets the default placement to `bottom` — it is a drop*down* after all
* ensures the transition works in all cases
* prevents the `FocusOn` div from getting styled which was causing a small grey box in the top left corner of all dropdowns

cc @artsy/diamond-devs


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@41.0.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4041.0.1), [`@artsy/palette@42.0.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4042.0.1) :rocket:

PR #1458, opened 2025-08-08
fix: multiple PhoneInput instances now work together on the same page

chr-tatu:
This PR fixes: [EMI-2570](https://artsyproduct.atlassian.net/browse/EMI-2570).

### Description
- Refactors the PhoneInput component to fix multiple instances working together on the same page:
    - Removed `key` prop from usePosition that was causing issues with multiple instances.
    - Replaced useWidthOf hook with direct DOM measurement, which fixed incorrect dropdown sizing and positioning conflicts between multiple instances
- Removed the use-keyboard-list-navigation library usage. The arrow selection was not working and the library was not maintained in years. Implemented instead  custom keyboard navigation.

@artsy/emerald-devs 

[EMI-2570]: https://artsyproduct.atlassian.net/browse/EMI-2570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/palette/src/elements/PhoneInput/PhoneInput.tsx
+++ b/packages/palette/src/elements/PhoneInput/PhoneInput.tsx
@@ -44,7 +44,6 @@ export interface PhoneInputProps extends Omit<InputProps, "onSelect"> {
   required?: boolean
   dropdownValue?: string
   inputValue?: string
-  instanceId?: string
 }
 
 export const PhoneInput = React.forwardRef<HTMLInputElement, PhoneInputProps>(
@@ -60,7 +59,6 @@ export const PhoneInput = React.forwardRef<HTMLInputElement, PhoneInputProps>(
       onSelect,
       dropdownValue,
       inputValue,
-      instanceId,
       ...rest
     },
     forwardedRef
@@ -108,7 +106,7 @@ export const PhoneInput = React.forwardRef<HTMLInputElement, PhoneInputProps>(
     })
 
     const { anchorRef, tooltipRef } = usePosition({
-      key: `${instanceId || "default"}-${filteredOptions.length}`,
+      key: `default-${filteredOptions.length}`,
       position: "bottom",
       offset: 10,
       active: isDropdownVisible,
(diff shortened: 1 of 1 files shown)


packages/palette/src/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -93,28 +106,13 @@ export const PhoneInput = React.forwardRef<HTMLInputElement, PhoneInputProps>(
...
-      waitForInteractive: true,
-      onEnter: ({ element: option, event }) => {
-        event.preventDefault()
-        event.stopPropagation()
-        if (option) {
-          handleSelect(option)
-        }
-        resetUI()
-      },
-    })
```
dzucconi, line 115: Why would we want to remove keyboard nav?

dzucconi: Also I wrote this library :) if there's a problem with it, it's fixable. But it's very simple!

dzucconi: Yeah, whoever implemented this kind of just missed the point of the lib. Notice how the thing you use to actually highlight things isn't even destructured.

dzucconi:
I would suggest just making a story that replicates the issue you're seeing and I'll fix this component and implement the keyboard nav correctly.

packages/palette/src/elements/PhoneInput/PhoneInput.tsx
```diff
@@ -93,28 +106,13 @@ export const PhoneInput = React.forwardRef<HTMLInputElement, PhoneInputProps>(
     })
 
     const { anchorRef, tooltipRef } = usePosition({
-      key: filteredOptions.length,
```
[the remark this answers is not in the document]
chr-tatu: You're so right, @dzucconi, on adding a story with multiple Phone inputs. The display issue doesn't reproduce. Now it seems more related on how we use it. I marked this PR as draft, while I continue to investigate the force implementation. 

The keyboard still doesn't work, though.

chr-tatu:
> @chr-tatu ok to close this?

Yes!

PR #1461, opened 2025-08-13
feat(filter-select): Add moveSelectedToTop prop

damassi:
There's a use case in CMS where this auto move behavior is confusing (within long scrolling content where top is out of window), so adding a prop to disable.


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@42.3.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4042.3.0), [`@artsy/palette@43.3.0`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4043.3.0) :rocket:

PR #1464, opened 2025-08-19
feat: Generic phone input component

olerichter00:
## Description

Generic phone input component.


packages/palette/src/elements/SelectInput/SelectInput.tsx
```diff
@@ -0,0 +1,527 @@
...
+ * @property {string} `name` is the content that will be displayed in the dropdown list
+ * @property {string} `value` is the value that will be passed to onSelect
+ */
+export interface SelectOption {
+  text: string
+  name: string
+  value: string
+}
+
+export interface SelectInputProps extends Omit<InputProps, "onSelect"> {
```
olerichter00, line 28: I mostly copied the code from `PhoneInput` and added a few props for customization.

dzucconi (changes requested):
Have you rebased recently? All the stuff in Select got copied from the buggy version of the PhoneInput.

Additionally: You don't just copy the thing. Genericize it then use it as the underlying implementation instead of leaving the code to rot.

PR #1468, opened 2025-08-21
fix(autocomplete): better handles focusing when using a pointer device

dzucconi:
Here we make the distinction that pointer interactions should not automatically scroll the results, whereas keyboard innteractions do.

https://github.com/user-attachments/assets/0a44baa1-a16f-4feb-914b-389103f7a852


damassi:
Thanks so much @dzucconi 👍

PR #1515, opened 2026-04-07
fix(filterselect): handle nullish entries in selecteditems

dzucconi:
We can see this in production in some instances. I'm not sure how you can into this state naturally; as the filter aggregations should just not show up. But URLs accessed directly can reproduce it. (Currently: https://www.artsy.net/collect?location_cities%5B0%5D=Tel%20Aviv-Yafo%2C%20Israel&acquireable=true&offerable=true)

cc @artsy/diamond-devs

### DIFF
--- a/packages/palette/src/elements/FilterSelect/Components/FilterSelectContext.tsx
+++ b/packages/palette/src/elements/FilterSelect/Components/FilterSelectContext.tsx
@@ -157,6 +157,7 @@ export const FilterSelectContextProvider: React.FC<
 
   const contextValue = {
     ...state,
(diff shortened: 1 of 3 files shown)


artsyit:
:rocket: PR was released in [`@artsy/palette-charts@45.9.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4045.9.2), [`@artsy/palette@46.9.2`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4046.9.2) :rocket:

dzucconi:
OK, so this is the reason Palette hasn't been updated in ... nearly 5 months. Will fix on Monday!

PR #1518, opened 2026-06-24
fix(shelf): prevents content shift on ssr shelf instances

dzucconi:
I noticed on /artists that, on server-side render, the `Shelf` renders the cells flush with the left edge of the screen, then, on mount, shifts them over to the page margin where they belong.

https://github.com/user-attachments/assets/25f16ac7-29dd-4231-a9ee-e87951a19b85

This actually is happening on *all* instances of `Shelf`.

Previously, there was a piece of code in place to prevent this:

```tsx
// To prevent any page jank we initially partially disable this component
// so that content is left aligned with the parent container and then once
// we have the offset and page values; we reset it to actually full-bleed.
// The `offset` will push the content up to the parent margin.
{...(!mounted ? { left: null, right: null, marginLeft: null } : {})}
```

This just neutralized the behavior of `FullBleed` then re-enabled it once we mount and have the positions for everything.

[#1503](https://github.com/artsy/palette/issues/1503) migrated `FullBleed` from `defaultProps` to styled-components attrs with nullish coalescing. Unlike `defaultProps` (which only defaults on `undefined`), `??` coalesces null back to the full-bleed default.

So, this PR adds a way to disable the `FullBleed` component correctly, rather than having to just re-style it from the outside.

cc @artsy/diamond-devs 

<details>
  <br />

  ```bash
  npm install @artsy/palette-charts@45.9.3-canary.1518.33548.0
  npm install @artsy/palette@46.9.3-canary.1518.33548.0
  # or 
  yarn add @artsy/palette-charts@45.9.3-canary.1518.33548.0
  yarn add @artsy/palette@46.9.3-canary.1518.33548.0
  ```
</details>

### DIFF
--- a/packages/palette/src/elements/FullBleed/FullBleed.test.tsx
+++ b/packages/palette/src/elements/FullBleed/FullBleed.test.tsx
@@ -1,0 +1,30 @@
+import { mount } from "enzyme"
+import "jest-styled-components"
+import React from "react"
+import { FullBleed } from "./FullBleed"
+
+describe("FullBleed", () => {
+  it("breaks out of its parent container by default", () => {
+    const wrapper = mount(<FullBleed>content</FullBleed>)
+    const node = wrapper.find("div").last()
+
+    expect(node).toHaveStyleRule("width", "100vw")
+    expect(node).toHaveStyleRule("margin-left", "-50vw")
+    expect(node).toHaveStyleRule("left", "50%")
+  })
+
+  it("does not bleed when `enabled` is false", () => {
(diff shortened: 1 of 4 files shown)


dzucconi:
Confirmed in Force via canary

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@45.9.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4045.9.3), [`@artsy/palette@46.9.3`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4046.9.3) :rocket:

PR #1522, opened 2026-09-14
fix(filterselect): keeps selectedItems referentially stable

dzucconi:
### Description

Follow-up to #1515, which broke Force's build (every `ResultsFilter`-based test hangs with "Too long with no output" on palette 46.9.2 / 46.9.3).

**Cause.** #1515 built the FilterSelect context value with `selectedItems: state.selectedItems.filter(Boolean)`, which creates a new array on every render. `_FilterSelect` has a `useUpdateEffect` keyed on `selectedItems` that calls `onChange`. When the consumer's `onChange` updates state (as Force's `ResultsFilter` does via `ArtworkFilterContext`), the provider re-renders, produces a new array, the effect re-fires, and it loops forever. This would also loop in the browser on any artwork filter using `FilterSelect`.

**Fix.** Compact nullish entries once, in the initial state and in the `SET_SELECTED_ITEMS` reducer case, and pass `selectedItems` through from state untouched so its identity only changes when the selection changes. `compactItems` returns the same array when there is nothing to remove.

**Tests.** Adds a regression test that mounts a stateful consumer and asserts `onChange` fires exactly once per click. Against the 46.9.3 code it fails with React's "Maximum update depth exceeded". Also asserts nullish entries are stripped from the `onChange` payload. The existing null-data test from #1515 still passes.


dzucconi:
Working on getting https://github.com/artsy/force/pull/17379 merged (finally!?)

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@45.9.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4045.9.4), [`@artsy/palette@46.9.4`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4046.9.4) :rocket:

PR #1525, opened 2026-09-22
fix(input): associate labels with their form controls

rquartararo:
First of a series of small PRs working through the axe-core accessibility audit in `a11y-audit/` (680 failing nodes across 150 stories). This one clears roughly 65 of them and is the highest-leverage fix in the form-control family.

## The bug

`htmlFor` resolves against an element's **`id`**, not its `name`. `Input`, `TextArea` and `SelectInput` all pointed `htmlFor` at `name` while never setting an `id` on the control:

```tsx
<StyledInput name={inputProps.name} … />   // no id
<StyledLabel htmlFor={inputProps.name}>    // resolves against id — never matches
```

So the floating label rendered but was never associated with its input. With no real label, screen readers fall back to the `title` attribute, which axe flags as `label-title-only` (36 nodes).

`Select` already wired `htmlFor={id}` correctly, but `id` is optional and was never generated, so every story that didn't pass one rendered a `<select>` with no accessible name — `select-name` (29 nodes).

## The fix

Adds `src/utils/useStableId.ts` and uses it in all four components to generate a fallback id. A consumer-supplied `id` still wins everywhere.

The helper picks its implementation **at module load** rather than inside the hook, so the rules of hooks still hold:

```ts
const reactUseId = (React as any).useId
export const useStableId = reactUseId ? … : …
```

The counter fallback is needed because jest maps `react` → `react-17` (`moduleNameMapper` in `package.json`), where `useId` doesn't exist — calling it directly would crash every existing test of these components. The fallback is not SSR-safe and is commented as such; React 18 runtimes always take the `useId` path.

## Knock-on fixes

`LabeledInput`, `AutocompleteInput` and `FilterSelect`'s search field all render through `Input`, so they're fixed by the same change.

## Stories

Six `Select` stories (`Default`, `Focus`, `Hover`, `WithError`, `Disabled`, `WithSelectedValue`) render an intentionally untitled select. They get an `aria-label` rather than a `title` so the untitled visual state they exist to demonstrate is unchanged.

## Testing

- `yarn lint`, `yarn type-check`, `yarn test` all pass — 46 suites, 250 tests, no regressions.
- New coverage in the `Input`, `Select`, `TextArea` and `SelectInput` suites: label/control association, consumer id precedence, that the association is *not* keyed off `name`, and that two instances on a page get distinct ids.

No API changes and no visual changes, so no Chromatic diff is expected beyond the six Select stories being unchanged visually.


dzucconi (approved):
👍 🙏

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@45.10.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4045.10.1), [`@artsy/palette@46.10.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4046.10.1) :rocket:

PR #1526, opened 2026-09-22
fix(radio): use a valid role and always emit aria-checked

rquartararo:
Second in the series working through the axe-core audit in `a11y-audit/`. Two lines of source change, clearing roughly **110 of the 680 failing nodes** — the single biggest source-code win in the audit.

## Two independent bugs

**1. `role="radio"` on a `<label>` (76 nodes, `aria-allowed-role`)**

`Container` was rendered `as="label"`, and `radio` is not an allowed role on `<label>`. `Container` is a `styled(Flex)`, so simply dropping `as` yields a div — which is exactly what `Checkbox` already does with `role="checkbox"`.

Nothing was lost by removing the label: there is no `<input>` inside `Radio` for a label to be associated with. The whole control is a div plus a styled dot, and clicks are handled by `onClick` on the container either way.

**2. `aria-checked` omitted when unselected (34 nodes, `aria-required-attr`)**

`selected` was destructured with no default:

```tsx
selected,                      // undefined when not passed
…
aria-checked={selected}        // React drops the attribute entirely
```

So an unselected `Radio` exposed no checked state at all to assistive tech — a radio that can't report whether it's on. `Checkbox` and `Toggle` both already default this to `false`; `Radio` was the outlier.

## Blast radius

Both fixes are inherited for free by:
- `BorderedRadio`, which is a `styled(Radio)` (21 of the nodes)
- `FilterSelect`, whose single-select results render `Radio` (19 of the nodes)

No API changes — `role` and `as` are not part of the public `RadioProps` surface.

## ⚠️ Expect a Chromatic diff

The rendered element changes from `<label>` to `<div>`. I'd expect this to be visually identical (the component's own flex/spacing styles are unchanged, and neither element carries a meaningful UA default here), but Chromatic runs with `--exit-zero-on-changes`, so **please review the diffs deliberately rather than auto-accepting** — this is the one PR in the series where the element type changes.

Deliberately *not* added: `user-select: none`. `Checkbox` has it, but `<label>` didn't provide it either, so adding it here would be an unrelated behavior change.

## Testing

- `yarn lint`, `yarn type-check`, `yarn test` all pass — 47 suites, 245 tests.
- `RadioGroup`, `FilterSelect` and `MultiSelect` suites all still green, including `RadioGroup`'s `.simulate("click")` interaction tests.
- Adds `Radio/__tests__/Radio.test.tsx` — there was no `Radio` test file at all. Covers label rendering, `onSelect`, and specifically: that no `<label>` is emitted, that `aria-checked` is present and `false` when unselected, `true` when selected, and that `tabIndex` flips with `disabled`.

## Follow-up, not in this PR

A disabled `Radio` renders a `disabled` attribute on a div, where it does nothing — the `&:disabled` CSS branch can never match. `aria-disabled` would be the correct fix. It isn't an audit finding, so I've left it out to keep this reviewable; happy to file it separately.

The proper long-term fix is a real visually-hidden `<input type="radio">`, which requires threading `name` through `RadioGroup` and is a breaking change — that's queued as a later, opt-in PR.


anandaroop (approved):
Lgtm

artsyit:
:rocket: PR was released in [`@artsy/palette-charts@45.10.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette-charts%4045.10.1), [`@artsy/palette@46.10.1`](https://github.com/artsy/palette/releases/tag/%40artsy/palette%4046.10.1) :rocket:

